PR #4, opened 2021-09-15
[TEST] Add unit test cases and TCM config file

gichan-jang:
- Add unit tests for edge sensor using GTest.
 - Add TCM config file.


tests/CMakeLists.txt
```diff
@@ -4,3 +4,18 @@ TARGET_INCLUDE_DIRECTORIES(test_edge_sensor PRIVATE ${EDGE_REQUIRE_PKGS_INCLUDE_
 TARGET_LINK_LIBRARIES(test_edge_sensor -L${PROJECT_BINARY_DIR}/src edge-sensor)
 
 INSTALL (TARGETS test_edge_sensor DESTINATION ${BIN_INSTALL_DIR})
+
+# GoogleTest requires at least C++11
+SET(CMAKE_CXX_STANDARD 11)
```
jaeyun-jung, line 9: let's consider to set 14 as current nnstreamer requires it.
https://github.com/nnstreamer/nnstreamer/blob/50d82aa029c72053004cc35743ba54a79560bc0d/meson.build#L12

tests/CMakeLists.txt
```diff
@@ -4,3 +4,18 @@ TARGET_INCLUDE_DIRECTORIES(test_edge_sensor PRIVATE ${EDGE_REQUIRE_PKGS_INCLUDE_
...
+
+INCLUDE(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+)
+
+FetchContent_MakeAvailable(googletest)
```
again4you, line 17: Since gbs does not support the network, so it would be blocked for a while and failed.
Please remove this line.

tests/CMakeLists.txt
```diff
@@ -4,3 +4,18 @@ TARGET_INCLUDE_DIRECTORIES(test_edge_sensor PRIVATE ${EDGE_REQUIRE_PKGS_INCLUDE_
...
+
+ADD_EXECUTABLE(unittest_edge_sensor unittest_edge_sensor.cc)
+TARGET_INCLUDE_DIRECTORIES(unittest_edge_sensor PRIVATE ${INCLUDE_DIR})
+TARGET_LINK_LIBRARIES(unittest_edge_sensor gtest pthread edge-sensor)
```
again4you, line 21: Instead of using the linked library names directly, what about this?
```cmake
TARGET_LINK_LIBRARIES(unittest_edge_sensor ${TEST_REQUIRE_PKGS_LDFLAGS} edge-sensor)
```
To use this way, you need to update the `CMakefiles.txt` as below.
```cmake
  SET(TEST_REQUIRES_LIST
      gtest
  )
  INCLUDE(FindPkgConfig)
  PKG_CHECK_MODULES(TEST_REQUIRE_PKGS REQUIRED ${TEST_REQUIRES_LIST})
```

In case of Tizen, `BuildRequires:  gtest-devel` in spec is required.

again4you (changes requested):
Please submit the next version. :)

tests/CMakeLists.txt
gichan-jang: Good! Updated.

tests/CMakeLists.txt
gichan-jang: Thanks! Updated.

packaging/nnstreamer-edge.spec
```diff
@@ -12,6 +12,7 @@ Source1001: nnstreamer-edge.manifest
 
 BuildRequires:  cmake
 BuildRequires:  pkgconfig(paho-mqtt-c)
+BuildRequires:  gtest-devel
```
again4you, line 15: What about this?

```spec
%if 0%{?sensor_test}
BuildRequires:  gtest-devel
%endif
```

gichan-jang: It's better! Updated.

again4you:
LGTM!

again4you (approved):

PR #15, opened 2022-07-01
[Edge] size of data handle

jaeyun-jung:
Set arg - base type of edge-data handle.


gichan-jang (approved):
LGTM

PR #17, opened 2022-07-01
[Build] fix build error on tizen

jaeyun-jung:
Fix build error, temporal block the unittest.


again4you (approved):
AS workaround, LGTM!
@jaeyun-jung will fix it by updating the unit test.

PR #20, opened 2022-07-04
[Coverity] Fix coverity issue.

gichan-jang:
- Initialize command before use to avoid tainted data.
 - Check return value whether it is negative or not.


again4you (approved):
LGTM!

jaeyun-jung (approved):

PR #21, opened 2022-07-04
[Edge] validate command

jaeyun-jung:
Prevent receive failure, update source to validate edge command.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -34,6 +34,7 @@ typedef enum
  */
 typedef struct
 {
+  unsigned int magic;
   nns_edge_cmd_e cmd;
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -227,6 +229,27 @@ _nns_edge_cmd_clear (nns_edge_cmd_s * cmd)
   }
 }
 
+/**
+ * @brief Validate edge command.
+ */
```
gichan-jang, line 234: How about set `NNS_EDGE_MAGIC_DEAD` when command is cleared?
`_nns_edge_cmd_clear()`

jaeyun-jung: updated

gichan-jang (approved):
LGTM

PR #22, opened 2022-07-04
[Util] string util functions

jaeyun-jung:
Add util functions to copy memory. We should remove dependency to glib later.


gichan-jang (approved):
LGTM

myungjoo (approved):
LGTM.

Anyway.. do you really need to reimplement these? Isn't strdup available from string.h?
If there is a reason to reimplement these instead of using string.h, you may merge it.

PR #25, opened 2022-07-04
[Coverity] Check return value

gichan-jang:
Check return value of the socke lib function.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -191,9 +191,12 @@ _get_available_port (void)
   }
   sin.sin_port = port;
   if (bind (sock, (struct sockaddr *) &sin, sizeof (struct sockaddr)) == 0) {
-    getsockname (sock, (struct sockaddr *) &sin, &len);
-    port = ntohs (sin.sin_port);
-    nns_edge_logi ("Available port number: %d", port);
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -191,9 +191,10 @@ _get_available_port (void)
   }
   sin.sin_port = port;
   if (bind (sock, (struct sockaddr *) &sin, sizeof (struct sockaddr)) == 0) {
-    getsockname (sock, (struct sockaddr *) &sin, &len);
-    port = ntohs (sin.sin_port);
-    nns_edge_logi ("Available port number: %d", port);
+    if (getsockname (sock, (struct sockaddr *) &sin, &len) == 0) {
+      port = ntohs (sin.sin_port);
+      nns_edge_logi ("Available port number: %d", port);
+    }
```
myungjoo, line 197: Do you need to handle the case when getsockname returns a non-zero value?

gichan-jang: In that case, `_get_available_port ` returns 0 for the port number.
The caller should handle this case as error. ([L1003](https://github.com/nnstreamer/nnstreamer-edge/blob/78fe78e165f75051e86e222ba4b944edc8a1109e/src/libnnstreamer-edge/nnstreamer-edge-internal.c#L1003))

I added error message for this case.

PR #30, opened 2022-07-06
[Edge] unnecessary param and error case

jaeyun-jung:
Remove unnecessary param in edge APIs and fix error case.


again4you (approved):
More clear than before. LGTM!

PR #35, opened 2022-07-11
[Connection] cmd between server and client

jaeyun-jung:
Set flag to send/receive command and capability between server and client.


gichan-jang (approved):
LGTM

PR #36, opened 2022-07-11
[Socket] condition to check socket

jaeyun-jung:
Check socket state before sending/receiving the edge command.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -204,6 +204,28 @@ _get_available_port (void)
 }
 
 /**
+ * @brief Internal function to check connection.
+ */
+static bool
+_nns_edge_check_connection (nns_edge_conn_s * conn)
+{
+  GIOCondition condition;
+
+  if (!conn || !conn->socket || g_socket_is_closed (conn->socket))
+    return false;
+
+  condition = g_socket_condition_check (conn->socket,
+      G_IO_IN | G_IO_OUT | G_IO_PRI | G_IO_ERR | G_IO_HUP);
+
+  if (!condition || (condition & (G_IO_ERR | G_IO_HUP))) {
+    nns_edge_logw ("Socket is not available, possibly closed.");
+    return false;
+  }
+
+  return true;
+}
+
+/**
  * @brief initialize edge command.
  */
 static void
@@ -277,6 +299,11 @@ _nns_edge_cmd_send (nns_edge_conn_s * conn, nns_edge_cmd_s * cmd)
     return NNS_EDGE_ERROR_INVALID_PARAMETER;
   }
 
+  if (!_nns_edge_check_connection (conn)) {
+    nns_edge_loge ("Failed to send command, socket has error.");
+    return NNS_EDGE_ERROR_IO;
+  }
+
   if (!_send_raw_data (conn->socket, &cmd->info,
           sizeof (nns_edge_cmd_info_s), conn->cancellable)) {
     nns_edge_loge ("Failed to send command to socket.");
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -203,6 +203,28 @@ _get_available_port (void)
...
+/**
+ * @brief Internal function to check connection.
+ */
+static bool
+_nns_edge_check_connection (nns_edge_conn_s * conn)
+{
+  GIOCondition condition;
+
+  if (!conn)
+    return false;
```
again4you, line 215: If `GSocket` is NULL or a socket is already closed, then we don't need to check the socket's condition. 
So what about making a code as below?
```suggestion
  if (!conn || !conn->socket || g_socket_is_closed (conn->socket))
    return false;
```

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -277,6 +299,11 @@ _nns_edge_cmd_send (nns_edge_conn_s * conn, nns_edge_cmd_s * cmd)
     return NNS_EDGE_ERROR_INVALID_PARAMETER;
   }
 
+  if (!_nns_edge_check_connection (conn)) {
+    nns_edge_loge ("Failed to send command, socket has error.");
+    return NNS_EDGE_ERROR_IO;
+  }
```
again4you, line 305: Since `_nns_edge_check_connection()` function checks whether `conn` object is NULL or not,so NULL checking in 292 line is redundant. What about removing this NULL checking code?

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -306,6 +333,11 @@ _nns_edge_cmd_receive (nns_edge_conn_s * conn, nns_edge_cmd_s * cmd)
   if (!conn || !cmd)
     return NNS_EDGE_ERROR_INVALID_PARAMETER;
 
+  if (!_nns_edge_check_connection (conn)) {
+    nns_edge_loge ("Failed to receive command, socket has error.");
+    return NNS_EDGE_ERROR_IO;
+  }
```
again4you, line 339: ditto.

again4you (commented):
Checking the socket's state is necessary for edge communication. The basic operation of this PR looks good to me.

However, I want to recommend two things: One is GSocket status and the other is Redundant checking of `conn` object. Please check my opinion below and let me know your opinion.

src/libnnstreamer-edge/nnstreamer-edge-internal.c
jaeyun-jung: IMO returning invalid-param if connection data is null, and io-error if socket is not available.

src/libnnstreamer-edge/nnstreamer-edge-internal.c
jaeyun-jung: Thanks, I updated this condition.

again4you: Verified. I thanks it makes the code more robust. :100:

src/libnnstreamer-edge/nnstreamer-edge-internal.c
again4you: Ah! I agreed with your opinion. :+1:

again4you (approved):

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -203,6 +203,28 @@ _get_available_port (void)
...
+static bool
+_nns_edge_check_connection (nns_edge_conn_s * conn)
+{
+  GIOCondition condition;
+
+  if (!conn || !conn->socket || g_socket_is_closed (conn->socket))
+    return false;
+
+  condition = g_socket_condition_check (conn->socket,
+      G_IO_IN | G_IO_OUT | G_IO_PRI | G_IO_ERR | G_IO_HUP);
```
wooksong, line 218: FYI,
> It is meaningless to specify G_IO_ERR or G_IO_HUP in condition; these conditions will always be set in the output if they are true.

https://libsoup.org/gio/GSocket.html#g-socket-condition-check

jaeyun-jung: Agreed, but for future work - remove glib :)

PR #38, opened 2022-07-11
[Edge] remove client when received error

jaeyun-jung:
In message thread, remove connection data when received error from client.
Internally error command means closing socket.


gichan-jang (approved):
LGTM

PR #40, opened 2022-07-15
[Socket] condition to release thread

jaeyun-jung:
Update src to close socket and release msg thread.
GSocket will close socket automatically if its ref count is 0.


gichan-jang (approved):
LGTM

PR #41, opened 2022-07-15
[Svace] prevent warning

jaeyun-jung:
Prevent error case, handle memory limit when receiving data from socket.


wooksong (approved):
LGTM!

gichan-jang (approved):
LGTM

PR #42, opened 2022-07-15
[Util] function to get port number

jaeyun-jung:
Add util function to get available port.

TODO: update testcase (data transfer with fixed port) later.


gichan-jang (approved):
LGTM

PR #45, opened 2022-07-20
[Meta] list for metadata

jaeyun-jung:
Replace hashtable (glib) to list, setting/getting metadata of edge handle.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-common.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-common.c
@@ -629,8 +629,7 @@ nns_edge_data_create (nns_edge_data_h * data_h)
 
   memset (ed, 0, sizeof (nns_edge_data_s));
   ed->magic = NNS_EDGE_MAGIC;
-  ed->info_table = g_hash_table_new_full (g_str_hash, g_str_equal,
-      nns_edge_free, nns_edge_free);
+  nns_edge_metadata_init (&ed->metadata);
 
   *data_h = ed;
   return NNS_EDGE_ERROR_NONE;
@@ -659,7 +658,7 @@ nns_edge_data_destroy (nns_edge_data_h data_h)
       ed->data[i].destroy_cb (ed->data[i].data);
   }
 
-  g_hash_table_destroy (ed->info_table);
+  nns_edge_metadata_free (&ed->metadata);
 
   SAFE_FREE (ed);
   return NNS_EDGE_ERROR_NONE;
@@ -691,8 +690,6 @@ nns_edge_data_copy (nns_edge_data_h data_h, nns_edge_data_h * new_data_h)
 {
   nns_edge_data_s *ed;
   nns_edge_data_s *copied;
-  GHashTableIter iter;
-  gpointer key, value;
   unsigned int i;
   int ret;
 
@@ -724,13 +721,7 @@ nns_edge_data_copy (nns_edge_data_h data_h, nns_edge_data_h * new_data_h)
     copied->data[i].destroy_cb = nns_edge_free;
   }
 
-  g_hash_table_iter_init (&iter, ed->info_table);
-  while (g_hash_table_iter_next (&iter, &key, &value)) {
-    g_hash_table_insert (copied->info_table, nns_edge_strdup (key),
-        nns_edge_strdup (value));
-  }
-
-  return NNS_EDGE_ERROR_NONE;
+  return nns_edge_metadata_copy (&copied->metadata, &ed->metadata);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/libnnstreamer-edge/nnstreamer-edge-common.c
```diff
@@ -138,6 +138,276 @@ nns_edge_strdup_printf (const char *format, ...)
...
+    return NNS_EDGE_ERROR_INVALID_PARAMETER;
+
+  nns_edge_metadata_free (dest);
+
+  node = src->list;
+  while (node) {
+    ret = nns_edge_metadata_set (dest, node->key, node->value);
+    if (ret != NNS_EDGE_ERROR_NONE) {
+      nns_edge_metadata_free (dest);
+      return ret;
```
myungjoo, line 301: nns_edge_metadata_s tmp
tmp.list = dest->list;
loop: set(&tmp, ...)
if fail: return w/ error code (not touching dest)
if success: free dest, set dest->list = tmp.list

src/libnnstreamer-edge/nnstreamer-edge-common.h
```diff
@@ -45,6 +45,28 @@ extern "C" {
...
+  nns_edge_metadata_node_s *next;
+};
+
+/**
+ * @brief Internal data structure to handle metadata. This struct should be managed in the handle.
+ */
+typedef struct {
+  unsigned int list_len;
+  nns_edge_metadata_node_s *list;
+} nns_edge_metadata_s;
```
myungjoo, line 68: Need a lock to protect if the functions of nnstreamer-edge-common can be accessed from different threads.

again4you: I agree with myungjoo's opinion.  The result of `nns_edge_metadata_find()` would vary if `nns_edge_metadata_set()` function is running in another thread. I thinks that the read lock mechanism is good enough for this issue.

again4you (commented):
The code is good to read and easy to understand. All primitive functions based on Glib are changed to non-Glib functions so we can port `nnstreamer-edge` to other platforms more easily.

I just add some comments about multi-threading, memory function, and early exit. Please check my opinion.

src/libnnstreamer-edge/nnstreamer-edge-common.h
jaeyun-jung: Suggestion: adding lock in edge-data handle.
- metadata is internal struct.
- metadata is internally used in edge handle and edge-data handle, then called with edge handle lock. However edge-data handle does not have lock now.

src/libnnstreamer-edge/nnstreamer-edge-common.c
```diff
@@ -138,6 +138,276 @@ nns_edge_strdup_printf (const char *format, ...)
...
+    SAFE_FREE (node->value);
+    node->value = val;
+    return NNS_EDGE_ERROR_NONE;
+  }
+
+  node = malloc (sizeof (nns_edge_metadata_node_s));
+  if (!node)
+    return NNS_EDGE_ERROR_OUT_OF_MEMORY;
+
+  memset (node, 0, sizeof (nns_edge_metadata_node_s));
```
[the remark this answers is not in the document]
jaeyun-jung: updated

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -323,25 +338,68 @@ _nns_edge_cmd_receive (nns_edge_conn_s * conn, nns_edge_cmd_s * cmd)
...
+    if (!_receive_raw_data (conn->socket, cmd->meta, cmd->info.meta_size)) {
+      nns_edge_loge ("Failed to receive metadata from socket.");
+      ret = NNS_EDGE_ERROR_IO;
+      goto error;
+    }
+  }
+
+error:
   if (ret != NNS_EDGE_ERROR_NONE) {
     for (i = 0; i < n; i++) {
```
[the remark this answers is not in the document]
jaeyun-jung: Updated

src/libnnstreamer-edge/nnstreamer-edge-common.c
jaeyun-jung: updated

src/libnnstreamer-edge/nnstreamer-edge-common.h
myungjoo: That's be ok.

myungjoo (approved):
LGTM. PTAL: @again4you

myungjoo:
Anyway, please hide internal data structure from other .C files. The internal data structure of a handle should not be exposed to its users.

src/libnnstreamer-edge/nnstreamer-edge-common.h
again4you: I agreed with jaeyun-jung  's opinion. 👍

PR #46, opened 2022-07-21
[Debian] Use paho-mqtt w/o pkgconfing

gichan-jang:
- Use paho-mqtt-c library without pkgconfig.


myungjoo (approved):
LGTM

PR #48, opened 2022-07-26
[CodeClean] remove glib function

jaeyun-jung:
Update source to remove dependency to glib.


gichan-jang (approved):
LGTM

PR #52, opened 2022-08-03
[Edge] handle invalid port number

jaeyun-jung:
Add error case if given port number is invalid.


anyj0527 (approved):
LGTM ✔

PR #53, opened 2022-08-11
[MQTT] topic in broker handle

jaeyun-jung:
1. Do not replace topic string in edge handle (set topic in mqtt-handle).
2. Use util function to get/parse host string.
3. Fix res leak case when mqtt connection is failed.
4. Code clean, remove unnecessary conversion after allocating new memory for mqtt-handle.


gichan-jang (approved):
LGTM

PR #57, opened 2022-08-12
[Build] dependency to mqtt lib

jaeyun-jung:
Find pkgconfig or library of mqtt library, and set flags if found dependency to mqtt.
Fix build error (invalid args when calling mqtt-connect in test).
Handle invalid param - topic.


gichan-jang (approved):
LGTM

PR #58, opened 2022-08-12
[CodeClean] handle allocation failure case

jaeyun-jung:
Fix svace issue, handle error case if memory allocation is failed.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-mqtt.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-mqtt.c
@@ -186,7 +186,8 @@ mqtt_cb_message_arrived (void *context, char *topic, int topic_len,
       eh->id, eh->topic);
 
   msg = nns_edge_memdup (message->payload, message->payloadlen);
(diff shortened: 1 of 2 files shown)


src/libnnstreamer-edge/nnstreamer-edge-mqtt.c
```diff
@@ -186,7 +186,8 @@ mqtt_cb_message_arrived (void *context, char *topic, int topic_len,
       eh->id, eh->topic);
 
   msg = nns_edge_memdup (message->payload, message->payloadlen);
-  g_async_queue_push (bh->server_list, msg);
+  if (msg)
```
gichan-jang, line 189: Don't need to check `bh->server_list`?

jaeyun-jung: not here, if failed to create queue when calling mqtt-connect, it will raise an exception.

gichan-jang (approved):
LGTM

PR #59, opened 2022-08-12
[Test] info from edge handle

jaeyun-jung:
Add testcases to set/get the info from edge handle.


gichan-jang (approved):
LGTM

myungjoo (approved):

PR #63, opened 2022-08-16
[Test] testcase for serialization of edge-data

jaeyun-jung:
Add testcases to handle error case of serialization in edge-data.


gichan-jang (approved):
LGTM

PR #65, opened 2022-08-18
[CodeClean] param to send/recv data

jaeyun-jung:
Code clean, prepare socket impl without glib.


gichan-jang (approved):
LGTM

PR #67, opened 2022-08-18
[Test] set valid condition for MQTT test

jaeyun-jung:
Set valid condition and start MQTT to run invalid param TCs.


gichan-jang (approved):
LGTM

PR #74, opened 2022-08-25
[Interface] rename (publish to send)

jaeyun-jung:
Prepare interface udpate,
- rename publish() to send()
- remove testcase - no client ID in data (If data handle does not have client ID, it will send data to all connected nodes.)


gichan-jang (approved):
LGTM

myungjoo (approved):

PR #75, opened 2022-08-26
[MQTT] replace glib queue

jaeyun-jung:
Replace g-async-queue to internal queue in MQTT functions.


gichan-jang (approved):
LGTM

PR #79, opened 2022-08-30
Format to print client ID

jaeyun-jung:
Change data type to print client ID.


gichan-jang (approved):
LGTM

PR #80, opened 2022-08-31
[Util] generate client ID

jaeyun-jung:
Remove glib, add util function to generate client ID.


gichan-jang (approved):
LGTM

PR #81, opened 2022-08-31
[Test] alloc failure case

jaeyun-jung:
1. Add assertion when allocating memory for test-data is failed.
2. Replace g_free.


gichan-jang (approved):
LGTM

PR #83, opened 2022-08-31
[Build] remove glib dependency

jaeyun-jung:
We replaced all glib functions in nns-edge, remove glib dependency.


myungjoo (approved):
LGTM!!!

PR #88, opened 2022-09-07
[Build/Test] add mosquitto to run test

jaeyun-jung:
Clearly define build-require to mosquitto, to run unittest using mqtt broker.


anyj0527 (approved):
LGTM

PR #95, opened 2022-09-20
[AITT] add aitt option

jaeyun-jung:
Set AITT option to open AITT handle.


jaeyun-jung:
We need force update - new interface of AITT (aitt option handle) before group sr.

gichan-jang (commented):
AITT changed the c-api, to support option when create the aitt handle.
To sync with latest AITT package, need to update aitt related code.
LGTM.

PR #97, opened 2022-09-27
[CodeClean] prevent error case

jaeyun-jung:
Code clean, fix coverity issue and prevent error case.


again4you (approved):
LGTM!

PR #98, opened 2022-09-27
[Build] default option for extern libs

jaeyun-jung:
Set default option off for external libraries (AITT and Paho-MQTT).


myungjoo (approved):
LGTM.

Is this "for-next"? or "for-tizen-7.0"?
(You'll need branching out if it's for-next)

jaeyun-jung:
IMO default build option of aitt and mqtt should be off.

again4you (approved):
LGTM!

PR #106, opened 2022-10-05
[Sock] event in recv thread

jaeyun-jung:
Update code to handle sock event in recv thread.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -697,10 +697,8 @@ _nns_edge_message_handler (void *thread_data)
   nns_edge_thread_data_s *_tdata = (nns_edge_thread_data_s *) thread_data;
   nns_edge_handle_s *eh;
   nns_edge_conn_s *conn;
-  nns_edge_cmd_s cmd;
   bool remove_connection = false;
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -715,64 +713,75 @@ _nns_edge_message_handler (void *thread_data)
...
+    poll_fd.revents = 0;
 
-    if (cmd.info.cmd == _NNS_EDGE_CMD_ERROR) {
-      nns_edge_loge ("Received error, stop msg thread.");
-      _nns_edge_cmd_clear (&cmd);
-      remove_connection = true;
-      break;
-    }
+    /* 10 milliseconds */
+    if (poll (&poll_fd, 1, 10) > 0) {
```
gichan-jang, line 729: Don't need to handle error case except timeout?

jaeyun-jung: waiting for new data here, timeout is no data received.

jaeyun-jung: And, _nns_edge_cmd_receive() checks connection state internally.

gichan-jang (approved):
LGTM

PR #114, opened 2022-10-17
[edge] Remove unnecessary pthread call

anyj0527:
- Remove `pthread_cancel`, cause msg and listener thread do not handle thread cancellation.
- By the way, Android NDK does not support `pthread_cancel`.

Signed-off-by: Yongjoo Ahn <yongjoo1.ahn@samsung.com>

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -524,7 +524,6 @@ _nns_edge_close_connection (nns_edge_conn_s * conn)
   /* Stop and clear the message thread. */
   if (conn->msg_thread) {
     if (conn->running) {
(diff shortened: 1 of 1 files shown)


jaeyun-jung:
Discussion w/ @gichan-jang canceling the thread for tensor-query

gichan-jang:
The reason for introducing pthred_cancel() was that when the message thread was waiting for receiving the message and tried to close the connection, there was an infinite waiting in the pthread_join().
I don't know whether this is solved or not.
I didn't write a regression test before, so I'll test it by myself.

gichan-jang:
When I tested with this patch, there is no previous problems. There's been a lot of change of the edge lib, and honestly, I don't know how it's been resolved.

gichan-jang (approved):
LGTM

jaeyun-jung (approved):

PR #115, opened 2022-10-19
[Queue] leaky option

jaeyun-jung:
Add leaky option in nns-edge queue.


gichan-jang (approved):
LGTM! Thanks.

PR #117, opened 2022-10-20
[Log] set log level

jaeyun-jung:
Add new function to set log level.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-log.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-log.c
@@ -1,0 +1,125 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/**
+ * Copyright (C) 2022 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file   nnstreamer-edge-log.c
+ * @date   20 October 2022
+ * @brief  Logging functions
+ * @see    https://github.com/nnstreamer/nnstreamer-edge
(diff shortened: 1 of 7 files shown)


src/libnnstreamer-edge/nnstreamer-edge-log.c
```diff
@@ -0,0 +1,125 @@
...
+#define _print_logw(...) _ne_log_print (NNS_EDGE_LOG_WARNING, __VA_ARGS__)
+#define _print_loge(...) _ne_log_print (NNS_EDGE_LOG_ERROR, __VA_ARGS__)
+#define _print_logf(...) _ne_log_print (NNS_EDGE_LOG_FATAL, __VA_ARGS__)
+#endif
+
+/**
+ * @brief Internal logging level.
+ */
+static nns_edge_log_level_e g_ne_log_level =
+    (DEBUG) ? NNS_EDGE_LOG_DEBUG : NNS_EDGE_LOG_INFO;
```
gichan-jang, line 80: How about setting the default log level as a warning?

jaeyun-jung: Q: any reason to set 'warn' as default?

gichan-jang: I thought it would be enough to print only warning messages to normal users.
However, there are not many info logs in the edge lib, so I think it doesn't matter.

gichan-jang (approved):
LGTM

myungjoo (approved):

PR #122, opened 2022-11-01
[CodeClean] unnecessary cond

jaeyun-jung:
Code clean, remove unnecessary condition and fix var type.


gichan-jang (approved):
LGTM

PR #131, opened 2022-11-18
[CodeClean] invalid arg type

jaeyun-jung:
Fix svace issue, size_t in nns-edge was changed to unsigned int 64. Fix invalid arg type to print log.


anyj0527 (approved):
LGTM✔

PR #132, opened 2022-11-18
[Util] util function to allocate memory

jaeyun-jung:
Fix svace issue - size limit to allocate memory.

Add util function for mem allocation, checking max size before calling malloc.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-util.h
+++ b/src/libnnstreamer-edge/nnstreamer-edge-util.h
@@ -20,6 +20,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <limits.h>
 #include "nnstreamer-edge.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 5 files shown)


src/libnnstreamer-edge/nnstreamer-edge-util.h
```diff
@@ -41,6 +42,13 @@ extern "C" {
 #define FALSE (0)
 #endif
 
+/**
+ * @brief The max size to allocate memory. This would be defined in C99 limits.h.
+ */
+#ifndef SIZE_MAX
+#define SIZE_MAX ((size_t) -1)
```
anyj0527, line 49: `nns_size_t` ?

jaeyun-jung: No, arg of malloc() is size_t and its type may be uint32 but nns_size_t is uint64.

anyj0527: You're right, thanks

anyj0527 (approved):

PR #133, opened 2022-11-29
[CodeClean/MQTT] broker handle

jaeyun-jung:
Code clean, add broker handle and revise MQTT functions. This will remove dependency to edge-handle in MQTT impl.


gichan-jang (approved):
LGTM

PR #137, opened 2022-12-05
[CodeClean/Queue] handle data size

jaeyun-jung:
Code clean, prepare next PR. Handle data size of queue.


gichan-jang (approved):
LGTM

PR #138, opened 2022-12-05
[AITT] separate handle

jaeyun-jung:
Separate AITT handle and update related code, to remove dependency to nns-edge handle.


gichan-jang (approved):
LGTM

PR #139, opened 2022-12-05
[Handle] macro to check handle

jaeyun-jung:
Macro to check handle in nns-edge.
The magic key would be changed when a handle is destroyed. If handle is not valid, its mutex lock is also invalid.


gichan-jang (approved):
LGTM

PR #140, opened 2022-12-06
[CodeClean/Handle] internal struct for handle

jaeyun-jung:
Code clean, move definition for internal data struct and cleanup headers.


gichan-jang (approved):
LGTM

PR #144, opened 2023-01-03
[Common] fix ret code after creating a thread

jaeyun-jung:
Fix returned code after creating a thread.


gichan-jang (approved):
LGTM!

PR #145, opened 2023-01-03
Safe momory free

gichan-jang:
Set NULL after release the memory.


src/libnnstreamer-edge/nnstreamer-edge-util.c
```diff
@@ -135,6 +135,7 @@ nns_edge_free (void *data)
 {
   if (data)
     free (data);
+  data = NULL;
```
jaeyun-jung, line 138: well,, this does not work.

gichan-jang: Right, Updated.

jaeyun-jung (approved):

PR #146, opened 2023-01-03
[Connect] flag to check sending thread

jaeyun-jung:
1. Add flag to check sending thread.
2. Init data ptr and size in the queue.


gichan-jang (approved):
LGTM

PR #151, opened 2023-03-15
[MQTT] Support MQTT direct data transmission @open sesame 03/16 11:51

gichan-jang:
- Support MQTT direct data transmission.
 - Add unit test for mqtt


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -1464,6 +1481,71 @@ nns_edge_set_event_callback (nns_edge_h edge_h, nns_edge_event_cb cb,
...
+
+  do {
+    char *msg = NULL;
+    char *server_ip = NULL;
+    int server_port = 0;
+    nns_size_t msg_len = 0;
+
+    ret = nns_edge_mqtt_get_message (eh->broker_h, (void **) &msg, &msg_len);
+    if (ret != NNS_EDGE_ERROR_NONE || !msg || msg_len == 0)
+      break;
```
songgot, line 1500: How about 'return' instead of 'return' after 'break'?

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -1464,6 +1481,71 @@ nns_edge_set_event_callback (nns_edge_h edge_h, nns_edge_event_cb cb,
...
+    SAFE_FREE (msg);
+
+    nns_edge_logd ("[DEBUG] Parsed server info: Server [%s:%d] ", server_ip,
+        server_port);
+
+    ret = _nns_edge_connect_to (eh, eh->client_id, server_ip, server_port);
+    SAFE_FREE (server_ip);
+
+    if (NNS_EDGE_ERROR_NONE == ret) {
+      break;
```
songgot, line 1512: How about 'return' instead of 'return' after 'break'?

songgot (approved):
LGTM

src/libnnstreamer-edge/nnstreamer-edge-internal.c
gichan-jang: There is no other work after the break, so I changed it to return as you suggested. Thanks.

src/libnnstreamer-edge/nnstreamer-edge-internal.c
gichan-jang: ditto

tests/unittest_nnstreamer-edge.cc
```diff
@@ -3743,7 +3743,7 @@ _check_mqtt_broker ()
 /**
  * @brief Connect to the local host using the information received from mqtt.
  */
-TEST(edgeMqtt, connectLocal)
+TEST(edgeMqttHybrid, connectLocal)
```
jaeyun-jung, line 3746: How about removing all mqtt tests in this file?
IMO, negative cases are duplicated.

gichan-jang: I moved mqtt realted tests from `unittest_nnstreamer-edge.cc` to `unittest_nnstreamer-edge-mqtt.cc` and remove duplicated nagative tests. Thanks.

jaeyun-jung (approved):

PR #152, opened 2023-03-20
[Connection] remove conn-type udp

jaeyun-jung:
Remove enum - UDP connection, nnstreamer-edge handles TCP connection only.


gichan-jang (approved):
LGTM

again4you (approved):
Nice catch!

PR #160, opened 2023-04-04
[MQTT] Change timeout for getting message from mqtt broker

gichan-jang:
Change timeout for getting message from mqtt broker.
 default timeout: 0 (infinite timeout)

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -1037,11 +1037,6 @@ _nns_edge_accept_socket (nns_edge_handle_s * eh)
   /* Send capability and info to check compatibility. */
   if ((NNS_EDGE_NODE_TYPE_QUERY_SERVER == eh->node_type)
       || (NNS_EDGE_NODE_TYPE_PUB == eh->node_type)) {
-    if (!STR_IS_VALID (eh->caps_str)) {
-      nns_edge_loge ("Cannot accept socket, invalid server capability.");
-      goto error;
-    }
-
     _nns_edge_cmd_init (&cmd, _NNS_EDGE_CMD_CAPABILITY, client_id);
     cmd.info.num = 1;
     cmd.info.mem_size[0] = strlen (eh->caps_str) + 1;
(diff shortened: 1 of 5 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -1253,6 +1253,7 @@ nns_edge_create_handle (const char *id, nns_edge_connect_type_e connect_type,
   eh->listening = false;
   eh->sending = false;
   eh->listener_fd = -1;
+  eh->caps_str = nns_edge_strdup ("nnsEdgeCapsNotSet");
```
jaeyun-jung, line 1256: Q: init null, why do you set the meaningless string?

src/libnnstreamer-edge/nnstreamer-edge-mqtt-mosquitto.c
```diff
@@ -365,11 +365,21 @@ nns_edge_mqtt_subscribe (nns_edge_broker_h broker_h)
 }
 
 /**
- * @brief Get message from mqtt broker.
+ * @brief Get message from mqtt broker with infinite timeout.
```
jaeyun-jung, line 368: Suggestion: nns_edge_mqtt_get_message() is internal function for mqtt.
How about adding new param 'timeout' without new function?
```
int nns_edge_mqtt_get_message (nns_edge_broker_h broker_h, unsigned int timeout, void **msg, nns_size_t * msg_len);
```

src/libnnstreamer-edge/nnstreamer-edge-internal.c
gichan-jang: When caps is not set, an error occurs and the edge process is terminated.
This is to change the determination of the caps by the event recipient.

jaeyun-jung: Suggestion: nns_edge_strdup (""); plz check empty string case.

gichan-jang: Thanks. Updated.

src/libnnstreamer-edge/nnstreamer-edge-mqtt-mosquitto.c
gichan-jang: Thanks for the suggestion. Updated.

jaeyun-jung (approved):

PR #169, opened 2023-07-12
[CMAKE] Modify guidelines for Build and Run test @open sesame 07/12 16:43

songgot:
pkgconfig was being installed on the wrong path.
now, pkgconfig path is CMKAE_INSTALL_PREFIX/lib/pkgconfig


gichan-jang (approved):
LGTM

jaeyun-jung (approved):

PR #177, opened 2023-08-18
[Cmd] version info in buffer

jaeyun-jung:
Add version info while sending and receiving buffer between edge connection.

TODO:
Now max memories in edge-data is 16. Check the limit when updating this definition.


songgot (approved):
LGTM

myungjoo (approved):

gichan-jang (approved):
LGTM

PR #180, opened 2023-09-07
[MQTT] function to clear retained msg

jaeyun-jung:
Add function to clear retained message, and set wait-time interval.


gichan-jang (approved):
Looks more clear! Thanks.

PR #183, opened 2023-09-13
[AITT] set/get option for AITT @open sesame 09/14 10:43

gichan-jang:
- Add set/get option of AITT.
 - Add releated unittests.

*Waiting for AITT tizen release.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-aitt.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-aitt.c
@@ -38,17 +38,17 @@ typedef struct
   /* event callback for new message */
   nns_edge_event_cb event_cb;
   void *user_data;
+  aitt_option_h option;
 } nns_edge_aitt_handle_s;
 
 /**
  * @brief Create AITT handle and connect to AITT.
  */
 int
-nns_edge_aitt_connect (const char *id, const char *topic, const char *host,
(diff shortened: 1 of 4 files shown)


src/libnnstreamer-edge/nnstreamer-edge-aitt.c
```diff
@@ -74,20 +74,11 @@ nns_edge_aitt_connect (const char *id, const char *topic, const char *host,
...
-    nns_edge_loge ("Failed to allocate memory for AITT handle.");
-    return NNS_EDGE_ERROR_OUT_OF_MEMORY;
-  }
-
-  option = aitt_option_new ();
-  aitt_option_set (option, AITT_OPT_MY_IP, "localhost");
-
-  ah->aitt_handle = aitt_new (id, option);
-  aitt_option_destroy (option);
+  ah->aitt_handle = aitt_new (id, ah->option);
```
jaeyun-jung, line 81: suggestion if separating create() and connect() function,
code style - function proto
```
int nns_edge_aitt_connect (nns_edge_aitt_h handle, const char *id, const char *topic, const char *host, const int port);
```

gichan-jang: Agree, Updated.

jaeyun-jung (approved):

PR #186, opened 2023-09-22
[Edge] Add metadata creation exception handling

linuxias:
It is necessary to consider the case of failing to generate Metadata
 when creating the Edge handle. It is not good to deliver Edge handles
 that have not been created normally to the user.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -1214,6 +1214,7 @@ int
 nns_edge_create_handle (const char *id, nns_edge_connect_type_e connect_type,
     nns_edge_node_type_e node_type, nns_edge_h * edge_h)
 {
+  int ret = NNS_EDGE_ERROR_NONE;
   nns_edge_handle_s *eh;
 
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -1258,20 +1259,29 @@ nns_edge_create_handle (const char *id, nns_edge_connect_type_e connect_type,
...
+  ret = nns_edge_metadata_create (&eh->metadata);
+  if (ret != NNS_EDGE_ERROR_NONE) {
+    nns_edge_loge ("Failed to create edge metadata");
+    goto error;
+  }
+
   nns_edge_queue_create (&eh->send_queue);
 
   if (NNS_EDGE_CONNECT_TYPE_AITT == connect_type) {
     int ret = nns_edge_aitt_create (&eh->broker_h);
```
jaeyun-jung, line 1272: let's remove declaration here, (int ret = ...)
`ret = nns_edge_aitt_create (&eh->broker_h);`

linuxias: I removed 'int', Thanks for your review!

jaeyun-jung (approved):

PR #190, opened 2023-10-18
[bugfix] Change the range of valid port

songgot:
- set available port when user set port to '0'
    - ml service api set '0' as the default port if the port isn't set
      but '0' is an invalid port

### DIFF
--- a/include/nnstreamer-edge.h
+++ b/include/nnstreamer-edge.h
@@ -310,7 +310,7 @@ int nns_edge_is_connected (nns_edge_h edge_h);
  * ---------------------|--------------------------------------------------------------
  * CAPS or CAPABILITY   | capability strings.
  * IP or HOST           | IP address of the node to accept connection from other node.
- * PORT                 | Port of the node to accept connection from other node. The value should be 0 or higher.
+ * PORT                 | Port of the node to accept connection from other node. The value should be 0 or higher, if the port is set to 0 then the available port is allocated.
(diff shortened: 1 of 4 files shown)


src/libnnstreamer-edge/nnstreamer-edge-util.h
```diff
@@ -52,7 +52,7 @@ extern "C" {
 #endif
 
 #define STR_IS_VALID(s) ((s) && (s)[0] != '\0')
-#define PORT_IS_VALID(p) ((p) > 0 && (p) <= 65535)
+#define PORT_IS_VALID(p) ((p) >= 0 && (p) <= 65535)
```
gichan-jang, line 55: Instead of changing this macro, how about getting available port when the user try to set port number 0?
inside `nns_edge_parse_port_number()`

myungjoo:
After this is merged, please update nnstreamer's edgesrc and edgesink property descriptions for CONNECT_TYPE == NNS_EDGE_CONNECT_TYPE_AITT.

src/libnnstreamer-edge/nnstreamer-edge-util.h
songgot: Thank you.

gichan-jang (approved):
LGTM,
It's nice it if you update the description for available port.
https://github.com/nnstreamer/nnstreamer-edge/blob/6a5eb295fde3eb7158275418539f5d8b14c251db/include/nnstreamer-edge.h#L313C6-L313C6

PR #192, opened 2023-10-27
[Edge] Add cond for nns_edge_handle

anyj0527:
- Add a pthread_cond for nns_edge_handle. And let the start method wait for the cond to be signaled by send_thread.
- This prevents some race conditions when APIs are used while send_thread is not yet started.


jaeyun-jung (approved):

gichan-jang (approved):
LGTM
Let's destroy the cond when releasing edge handle.

PR #193, opened 2023-10-31
[Hybrid] Reconnect to available node

gichan-jang:
If current connection is lost, find available node and reconnect. 
Previously, This recovery work is done by user application. 
Let's make the users don't care about this work when using MQTT-HYBRID.

*edgesrc and tensor_query_* should be updated after this change.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -151,6 +151,11 @@ typedef struct
 } nns_edge_thread_data_s;
 
 /**
+ * @brief Parse the message received from the MQTT broker and connect to the server directly.
+ */
+static int _mqtt_hybrid_direct_connection (nns_edge_handle_s * eh);
+
+/**
  * @brief Set socket option. nnstreamer-edge handles TCP connection now.
  */
 static void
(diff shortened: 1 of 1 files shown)


src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -777,6 +782,10 @@ _nns_edge_message_handler (void *thread_data)
     _nns_edge_remove_connection (eh, client_id);
     nns_edge_event_invoke_callback (eh->event_cb, eh->user_data,
         NNS_EDGE_EVENT_CONNECTION_CLOSED, NULL, 0, NULL);
+    if (NNS_EDGE_CONNECT_TYPE_HYBRID == eh->connect_type) {
+      nns_edge_logd ("Connection lost! Reconnect to available node.");
+      ret = _mqtt_hybrid_direct_connection (eh);
```
songgot, line 787: Does _mqtt_hybrid_direct_connection() connect immediately without waiting(lock? ) to receive messages?

songgot (approved):
LGTM

src/libnnstreamer-edge/nnstreamer-edge-internal.c
gichan-jang: The lock was added to the connection inside the `_mqtt_hybrid_direct_connection`.
And unlock where waiting for the mqtt message.

jaeyun-jung: Suggestion: 
1. lock-unlock when calling _mqtt_hybrid_direct_connection (eh), not in _mqtt_hybrid_direct_connection().
   This function (_nns_edge_message_handler) is thread callback. 
2. call connection-lost event if failed to reconnect.

```
_nns_edge_remove_connection (eh, client_id);
ret = NNS_EDGE_ERROR_CONNECTION_FAILURE;

if (NNS_EDGE_CONNECT_TYPE_HYBRID == eh->connect_type) {
  nns_edge_logi ("Connection lost! Reconnect to available node.");

  nns_edge_lock (eh);
  ret = _mqtt_hybrid_direct_connection (eh);
  nns_edge_unlock (eh);
}

  nns_edge_event_invoke_callback (eh->event_cb, eh->user_data,
      NNS_EDGE_EVENT_CONNECTION_CLOSED, NULL, 0, NULL);
}
```

gichan-jang: I reverted lock because this is called inside the thread and call connection-lost event if failed to reconnect. Thanks.

jaeyun-jung (approved):

PR #194, opened 2023-11-06
[Common] max memories in nns-edge

jaeyun-jung:
nnstreamer now supports max 256 tensors in the buffer.
Increase max memories in nnstreamer-edge handle.

TODO:
We will fix the util functions in edge elements of nnstreamer, to handle memories in gst-buffer.


myungjoo (approved):
LGTM. Anyway, can you specify the reason in the commit message?

jaeyun-jung:
@myungjoo I updated commit message.
(nnstreamer now supports 256 memories in the buffer, and I will fix util functions in edge elements of nnstreamer.)

PR #200, opened 2024-02-27
Fix CMAKE_INSTALL_LIBDIR exampel in README.md

linuxias:
In README, CMAKE_INSTALL_LIBDIR is set up with the "lib" path.
 The lib path is recorded in the pc file referenced by the pkg-config tool.
 This path must be modified to the absolute path "/lib".

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ $ sudo systemctl start mosquitto
 ### Build and Run test
 ```
 # cd $NNST_EDGE_ROOT
-$ cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=lib -DENABLE_TEST=ON -DMQTT_SUPPORT=ON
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ $ sudo systemctl start mosquitto
 ### Build and Run test
 ```
 # cd $NNST_EDGE_ROOT
-$ cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=lib -DENABLE_TEST=ON -DMQTT_SUPPORT=ON
+$ cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/lib -DENABLE_TEST=ON -DMQTT_SUPPORT=ON
```
myungjoo, line 48: Nope. We don't install our .so files at /lib. I think the CMakeLists.txt is not handling install direcctories properly for creating .pc files. It should depend on PREFIX. (i.e., $PREFIX/$LIBDIR)

linuxias: @myungjoo Thanks for your comment.
I agree your comment.  It seems that the LIBDIR path is not used properly in the CMakeLists.txt file.
I will close this PR and fix the problem, and create a new PR.

thank you!! :+1:

PR #202, opened 2024-02-28
Fix LIB_INSTALL_DIR variable in CMakeLists.txt

linuxias:
CMakeLists.txt is not handling install direcctories properly for creating .pc files. It should depend on PREFIX.


linuxias:
I needs to check build error.
i close this PR for checking error.

PR #216, opened 2024-08-22
[Custom] separate custom impl

jaeyun-jung:
Separate custom connection implementation and update callbacks.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
+++ b/src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
@@ -1,0 +1,85 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/**
+ * Copyright (C) 2024 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file   nnstreamer-edge-custom-impl.h
+ * @date   21 Aug 2024
+ * @brief  Internal interface to support communication using custom library.
+ * @see    https://github.com/nnstreamer/nnstreamer
+ * @author Gichan Jang <gichan2.jang@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#ifndef __NNSTREAMER_EDGE_CUSTOM_IMPL_H__
+#define __NNSTREAMER_EDGE_CUSTOM_IMPL_H__
+
+#include "nnstreamer-edge-custom.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Data structure for edge custom connection.
+ */
+typedef struct
+{
+  void *dl_handle;
+  nns_edge_custom_s *instance;
+  void *priv;
+} custom_connection_s;
+
+/**
+ * @brief Internal function to load custom connection from library.
+ */
+int nns_edge_custom_load (custom_connection_s *custom, const char *lib_path);
+
+/**
+ * @brief Internal function to release custom connection.
+ */
+int nns_edge_custom_release (custom_connection_s *custom);
+
+/**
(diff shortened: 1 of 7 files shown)


tests/nnstreamer-edge-custom-test.c
```diff
@@ -146,37 +146,40 @@ nns_edge_custom_send_data (void *priv, nns_edge_data_h data_h)
...
-static char *
-nns_edge_custom_get_option (void *priv, const char *key)
+static int
+nns_edge_custom_get_info (void *priv, const char *key, char **value)
 {
-  if (!priv || !key) {
-    nns_edge_loge ("Invalid param, key(%s)", key);
-    return NULL;
+  if (!priv || !key || !value) {
+    nns_edge_loge ("Invalid param, key(%p), value(%p)", key, value);
```
gichan-jang, line 171: In the case of invalid param, it is meaningless to print the value.
```suggestion
    nns_edge_loge ("Invalid param,  custom handle, key or value shoud not be NULL.);
```

tests/nnstreamer-edge-custom-test.c
```diff
@@ -146,37 +146,40 @@ nns_edge_custom_send_data (void *priv, nns_edge_data_h data_h)
...
-  if (strcasecmp (key, "PEER_ADDRESS") == 0)
-    return nns_edge_strdup (custom_h->peer_address);
 
-  return NULL;
+  if (strcasecmp (key, "PEER_ADDRESS") == 0) {
+    *value = nns_edge_strdup (custom_h->peer_address);
+    return NNS_EDGE_ERROR_NONE;
+  }
+
+  nns_edge_loge ("key(%s) does not supported.", key);
```
gichan-jang, line 181:
```suggestion
  nns_edge_loge ("key(%s) is not supported.", key);
```

src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
```diff
@@ -0,0 +1,85 @@
...
+{
+  void *dl_handle;
+  void *instance;
+  void *priv;
+} custom_connection_s;
+
+/**
+ * @brief Internal function to create custom connection.
+ */
+int nns_edge_custom_create (custom_connection_s *custom, const char *lib_path);
```
gichan-jang, line 35: Isn't it confusing for users to use the same function name with nnstreamer-edge-custom.h?

src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
```diff
@@ -0,0 +1,85 @@
...
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Data structure for edge custom connection.
+ */
+typedef struct
+{
+  void *dl_handle;
+  void *instance;
```
gichan-jang, line 28: If `instance` is defined as `nns_edge_custom_s *` type instead of `void *` type, it seems that repetitive casting can be reduced.
```suggestion
  nns_edge_custom_s *instance;
```

gichan-jang (approved):
LGTM

src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
jaeyun-jung: I updated param type and function ptr to load library. plz review the changes again.

src/libnnstreamer-edge/nnstreamer-edge-custom-impl.h
jaeyun-jung: updated function name to _load()

gichan-jang (approved):
LGTM

PR #221, opened 2024-08-28
 [Event] util to get data

jaeyun-jung:
Add util function to get the edge event data, and remove unnecessary return code.


gichan-jang (approved):
There was no way to get event data of the custom event.
Thank you for adding new util function.
LGTM

PR #225, opened 2024-09-13
[Event] Add new event type

gichan-jang:
Add new event type for new connection.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped


songgot (approved):
LGTM :+1:

include/nnstreamer-edge-event.h
```diff
@@ -30,6 +30,7 @@ typedef enum {
   NNS_EDGE_EVENT_NEW_DATA_RECEIVED,
   NNS_EDGE_EVENT_CALLBACK_RELEASED,
   NNS_EDGE_EVENT_CONNECTION_CLOSED,
+  NNS_EDGE_EVENT_CONNECTION_COMPLETED,
```
songgot, line 33: Does this message indicate that the network connection was successfully (completed)established? Or does it mean that the network connection is active?

gichan-jang: It means  that the network connection was successfully established.
If network connection status is changed to inactive, `NNS_EDGE_EVENT_CONNECTION_CLOSED` event will be invoked.

songgot: Thank you for the detailed explanation. :pray:

jaeyun-jung (approved):

PR #230, opened 2025-01-02
[CodeClean] fix segfault case

jaeyun-jung:
Fix segfault case,
1. stop sending thread before closing connection handle.
2. handle error case while pushing edge data into sending queue.


myungjoo (approved):

niley7464 (approved):
LGTM :)

PR #235, opened 2025-01-10
[Build/Android] add mqtt source

jaeyun-jung:
Define mqtt source for android build.

TODO: now we uses paho-mqtt, consider other mqtt library later.


gichan-jang (approved):
LGTM

PR #250, opened 2026-08-26
[Fix] clear out-param when custom lib loading fails

### LINKED ISSUE #249: nns_edge_custom_create_handle() leaves *edge_h dangling when the custom library fails to load
**Summary**: When the custom library fails to load, `nns_edge_custom_create_handle()` releases the handle it created internally but leaves the caller's out-param pointing at the freed handle. Callers that clean up on the error return then operate on freed memory.

**Where** (`src/libnnstreamer-edge/nnstreamer-edge-internal.c`, current main `010be16`):

```c
int
nns_edge_custom_create_handle (const char *id, const char *lib_path,
    nns_edge_node_type_e node_type, nns_edge_h * edge_h)
{
  ...
  ret = _nns_edge_create_handle (id, node_type, edge_h);   /* L1347: *edge_h is populated */
  ...
  ret = nns_edge_custom_load (lib_path, &eh->custom_connection_h);   /* L1356 */
  if (ret != NNS_EDGE_ERROR_NONE)
    nns_edge_release_handle (eh);   /* L1358: frees eh (SAFE_FREE at L1630) ... */

  return ret;                       /* ... but *edge_h still points at the freed handle */
}
```

**Impact**: A caller that receives the error and tries to clean up its stored handle performs a read-after-free. Concretely, in nnstreamer's `gst/edge/edge_sink.c` / `edge_src.c`, `gst_edgesink_start()` stores the out-param in `self->edge_h` and, on failure, calls `nns_edge_release_handle (self->edge_h)` on the already-freed handle. In practice `nns_edge_handle_is_valid()` reads the dead magic out of the freed block and returns `NNS_EDGE_ERROR_INVALID_PARAMETER` — no double free, no crash — but it is genuine undefined behavior that ASan/Valgrind will report, and it silently relies on the freed block not being reused in between.

This was found while reviewing the regression tests in nnstreamer/nnstreamer#4808 (the `edgeCustom.sinkInvalidProp2_n` / `srcInvalidProp2_n` cases with `custom-lib=libINVALID.so` exercise exactly this path).

**Suggested fix**: clear the out-param on the failure path so every caller is fixed at once:

```c
  ret = nns_edge_custom_load (lib_path, &eh->custom_connection_h);
    nns_edge_release_handle (eh);
    *edge_h = NULL;
  }
```

The sibling paths are already clean: `_nns_edge_create_handle()` only assigns `*edge_h` on success, and `nns_edge_create_handle()` has no failure path after populating the out-param — `nns_edge_custom_create_handle()` is the only spot with this pattern.

myungjoo:
## What

`nns_edge_custom_create_handle()` populates the caller's out-param via `_nns_edge_create_handle()` and then releases the handle internally when `nns_edge_custom_load()` fails, leaving `*edge_h` pointing at the freed handle. A caller that cleans up its stored handle on the error return (e.g., `gst_edgesink_start()` / `gst_edgesrc_start()` in nnstreamer) then reads freed memory — undefined behavior that ASan/Valgrind reports as use-after-free.

This PR clears the out-param on the failure path so every caller is fixed at once.

## How

- `src/libnnstreamer-edge/nnstreamer-edge-internal.c`: set `*edge_h = NULL` after `nns_edge_release_handle()` on the `nns_edge_custom_load()` failure path.
- `tests/unittest_nnstreamer-edge-custom.cc`: add `edgeCustom.createHandleLoadFail_n`, which passes an invalid library path (`libINVALID.so`) to reach the load-failure path and asserts the out-param is cleared. Verified the test fails against the unfixed library and passes with the fix.

## Verification

- `unittest_nnstreamer-edge-custom`: 26/26 passed (including the new test).
- `unittest_nnstreamer-edge`: 171/171 passed — no regression in existing behavior.

Fixes #249


myungjoo:
> **Note**: This comment was written by an AI review agent (not the human reviewer) and is being pasted in as-is per the requester's instructions.

## Review summary — Approve

I reviewed this PR by (1) reading the diff, (2) tracing `nns_edge_custom_create_handle()` / `nns_edge_release_handle()` / `_nns_edge_create_handle()` in `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, (3) checking the actual external caller of this API — `gst_edgesink_start()` / `gst_edgesrc_start()` in `nnstreamer/gst/edge/edge_sink.c` and `edge_src.c` — and (4) checking whether the new unit test is actually exercised by this repo's CI gates.

**Issue #249 fix — correct and minimally scoped**
- Root cause confirmed: in `nns_edge_custom_create_handle()`, `_nns_edge_create_handle()` already wrote `*edge_h = eh` before `nns_edge_custom_load()` is attempted. On load failure, the old code called `nns_edge_release_handle(eh)` (which frees `eh`) but never cleared `*edge_h`, leaving the caller's out-param pointing at freed memory.
- The fix (`*edge_h = NULL;` right after the release, only on that failure branch) is the correct, minimal fix. It doesn't touch the success path, and it doesn't change any other create-handle path — I checked `nns_edge_create_handle()` and the generic `_nns_edge_create_handle()`, neither has this bug (they never write `*edge_h` before failing), so widening the fix elsewhere isn't needed.

**No regression, and it actually prevents a real bug in nnstreamer callers**
- I pulled `gst_edgesink_start()` from `nnstreamer/gst/edge/edge_sink.c` (main branch) to check the real-world caller path described in the PR body. It's exactly as claimed: on `nns_edge_custom_create_handle()` failure it calls `gst_edgesink_release_handle(self)`, which does `if (self->edge_h) { nns_edge_release_handle(self->edge_h); self->edge_h = NULL; }`. Before this fix, `self->edge_h` would still hold the pointer just freed inside `nns_edge_custom_create_handle()`, so `gst_edgesink_release_handle()` would call `nns_edge_release_handle()` a second time on already-freed memory — a double-free/UAF. With this fix, `self->edge_h` becomes NULL and the `if` guard skips the second release. `gst_edgesrc_start()` in `edge_src.c` has the identical pattern. So this change is a regression *fix* for those callers, not a regression risk.
- The success path is untouched, so no behavioral change for the common case.

**Code size — appropriately scoped**
- 2 lines of production change + one new negative test case (~14 lines). No unrelated files touched, no other modules modified. Good.

**Test coverage**
- The new `edgeCustom.createHandleLoadFail_n` test directly asserts the fixed contract: passing an invalid `lib_path` returns an error and leaves `edge_h == NULL`. This directly guards against regression of the exact bug in #249. Test style/naming matches the existing `_n` (negative) tests in the same file.

**CI enforcement — verified this test is actually built and run, and will fail the PR gate on regression**
- `.github/workflows/gbs_build.yml` runs on `pull_request: [main]` with the `x86_64` matrix building with `--define "unit_test 1"`.
- `packaging/nnstreamer-edge.spec` under `%if 0%{?unit_test}` (with `custom_connection_support` on by default) runs `LD_LIBRARY_PATH=./src:./tests bash %{test_script} ./tests/unittest_nnstreamer-edge-custom` during `%install`.
- `packaging/run_unittests.sh` runs the gtest binary and `exit $ret` — a failing test fails the GBS build, which fails the CI job, which blocks merge. So a future change that reintroduces the dangling-pointer bug (or breaks this contract) would be caught by CI, not just by local runs.
- Minor/non-blocking side note (pre-existing, unrelated to this PR): `.github/workflows/ubuntu_clean_cmake_build.yml` runs `cmake` without `-DENABLE_TEST=ON`, so the `tests/` subdirectory is never added there and `ctest` has nothing to run — that workflow currently doesn't exercise any unit tests, including this new one. This isn't something this PR needs to fix, just flagging it as an existing gap in case it's useful context.

**Docs / API surface**
- No public API signature change, no ABI change, no architecture change. The existing Doxygen remark on `nns_edge_custom_create_handle()` ("If the function succeeds, @a edge_h should be released...") is still accurate; no doc update is required for this fix. Optional nice-to-have: an explicit `@remarks` noting `*edge_h` is set to NULL on failure would make the now-fixed contract self-documenting, but this is a suggestion, not a blocker.

**Note**: the PR is currently in Draft state with a `DONOTMERGE` label — presumably intentional/in-progress, just flagging in case that's stale.

**Verdict: No blocking issues found. Approve / mergeable as-is** (once the draft/DONOTMERGE state is cleared by the author).

myungjoo:
> **Note**: This comment was written by an AI review agent (not the human reviewer) and is being pasted in as-is per the requester's instructions. This is a re-review after the two follow-up commits (`3866862` unchanged, `c4bbb23` new).

## Re-review of `c4bbb23` — both follow-up items resolved correctly

**1. Doxygen remark (`include/nnstreamer-edge.h`)**

New wording:
> "If the function fails, @a edge_h does not hold a valid handle and should not be released."

I checked this against every failure branch of `nns_edge_custom_create_handle()`, not just the one this PR fixes:
- invalid `node_type` / invalid `lib_path` / `edge_h == NULL` → function returns before `*edge_h` is ever written, so the out-param is left exactly as the caller set it (untouched).
- `_nns_edge_create_handle()` fails → its own error path releases the internal handle but never assigns `*edge_h` (only the success path does `*edge_h = eh`), so again the out-param is left untouched.
- `nns_edge_custom_load()` fails (the bug this PR fixes) → `*edge_h` is now explicitly set to `NULL`.

So across all four failure branches, the out-param is either untouched or explicitly NULL — never a stale/freed pointer. "Does not hold a valid handle" is accurate for all of them, whereas a blanket "set to NULL" claim would have been wrong for the first two branches (which don't touch it at all). This is a more precise and honest contract statement than my original suggestion, and it's still exactly as actionable for callers ("don't release it on failure"). No further change needed here.

**2. CMake `ENABLE_TEST` gap split out**

Agreed this is the right call — it's a pre-existing, unrelated CI infra gap (`ubuntu_clean_cmake_build.yml` never passes `-DENABLE_TEST=ON`), not something introduced by or in scope for this fix. Keeping #250 focused on the issue #249 fix is correct; happy to see it tracked separately.

## Verdict: **Approve / mergeable as-is**

No remaining issues. Diff is still minimally scoped (2-line fix + 1 new negative test + 1 doc-comment line), correctly resolves the double-free/UAF in `nns_edge_custom_create_handle()` (and, as traced in the previous review pass, the corresponding double-free in `nnstreamer`'s `gst_edgesink_start()`/`gst_edgesrc_start()` callers), is covered by a CI-enforced regression test (`gbs_build.yml` → `unittest_nnstreamer-edge-custom` via `packaging/run_unittests.sh`, which fails the build on test failure), and now has an accurate doc contract for the failure path. Only remaining item is clearing the Draft/`DONOTMERGE` state, which is an author/process decision, not a code issue.

PR #256, opened 2026-09-02
[Build] Compile and package on Ubuntu 26.04 (gcc 15)

myungjoo:
## Why

`ppa:nnstreamer` publishes nothing for Ubuntu 26.04 (resolute), and nnstreamer-edge is the first package that has to get there: `nnstreamer-core` depends on it. Two things stop the Launchpad recipe build of nnstreamer-edge on that series today.

1. **The source stage never starts.** `Build-Depends` names only `gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5`, none of which exist in Ubuntu 25.04 or later, so the recipe build parks in `Dependency wait: gcc-9` (that is what happened on plucky, recipe build 3939544). Same fix as nnstreamer/nnstreamer#4895: append the unversioned `gcc`, which the default compiler always provides.

2. **gcc 15 rejects two lines under `-Werror`.** Ubuntu 26.04 ships gcc 15, which compiles C as C23 by default; there glibc's `strchr()` and `strstr()` keep the constness of their argument, so storing the result in a `char *` is `-Werror=discarded-qualifiers`:

   ```
   nnstreamer-edge-util.c:131:13: error: initialization discards 'const' qualifier from pointer target type
   nnstreamer-edge-internal.c:2038:7: error: assignment discards 'const' qualifier from pointer target type
   ```

   Both pointers are only read (offset arithmetic and `strcasecmp`), so `const char *` is the correct type and compiles identically on every older compiler.

The unit tests need one more change to build there: the archive's googletest 1.17 refuses anything below C++17. Only the tests are C++, and nnstreamer itself already builds as C++17, so `CMAKE_CXX_STANDARD` goes from 14 to 17.

## Regression coverage

The existing CMake workflow runs on the `ubuntu-latest` image, whose gcc does not default to C23 and whose gtest accepts C++14, so it could not catch either problem. This adds a second job in that workflow that builds and runs the tests inside a `ubuntu:26.04` container. Without the source fixes that job fails exactly as the log above; with them it passes.

## Verified

On an Ubuntu 26.04 container (gcc 15.2, gtest 1.17):

- `cmake -DENABLE_TEST=ON` + build + `ctest`: 2/2 tests pass.
- The same with `-DMQTT_SUPPORT=ON`, which is what `debian/rules` builds.
- `dpkg-buildpackage -us -uc -b` with the patched `debian/control`: `nnstreamer-edge_0.2.8.0_amd64.deb` and `nnstreamer-edge-dev_0.2.8.0_amd64.deb` build.

## After merging

`nnstreamer-edge-daily-2` on Launchpad needs Resolute added to its series and a build requested; adding a series alone does not trigger one. Part of nnstreamer/nnstreamer#4902.


myungjoo:
[Automated review by a separate Claude agent, relayed to this PR]

No defects found. Findings, most important first:

- **Suggestion (coverage):** the new `build-ubuntu-26-04` job configures with `MQTT_SUPPORT` off, but the Launchpad build this PR exists for runs `debian/rules` with `-DMQTT_SUPPORT=ON` (mosquitto). Adding `libmosquitto-dev` to the apt line and `-DMQTT_SUPPORT=ON` to the configure step would make CI compile `nnstreamer-edge-mqtt-mosquitto.c` under gcc 15 as well; the MQTT test is not registered with `ADD_TEST`, so `ctest` is unaffected. Not blocking: that file has no `strchr`/`strstr` call, and the PR body says MQTT=ON was verified locally.
- **Nit:** `CMakeLists.txt:20` sets `CMAKE_CXX_STANDARD 17` without `CMAKE_CXX_STANDARD_REQUIRED ON`, so a compiler without C++17 would silently decay instead of failing at configure. No current runner is that old, so optional.

What I checked:

- `nnstreamer-edge-internal.c:2034-2052`: `s` is used only in `s - value`, `strcasecmp (s + 1, ...)` and a `%s` log; `nnstreamer-edge-util.c:131-136`: `p` is used only in `p - host_str` and `strtoll (p + 1, ...)`. All sinks take `const char *`, so `const` is correct and the behaviour is unchanged on older compilers.
- Grepped the whole head tree: these are the only two `strchr`/`strstr`/`strrchr`/`strpbrk` callers, so the MQTT, Tizen and custom-connection sources hide no third one.
- C++17 on existing runners: `ubuntu-latest` (gtest 1.14), pdebuild on jammy (gcc 11), and the GBS spec, which sets no `-std=`, all accept it. No docs still mention C++14.
- Workflow YAML: valid; the job already shows up as the check "Build and test on Ubuntu 26.04 (gcc 15)". Root container, so `apt-get` without `sudo` is right; `pkg-config` is installed for `PKG_CHECK_MODULES(gtest)`; `env.BUILD_TYPE` is workflow-level and visible to the job.
- `debian/control`: `gcc-5 (>=5.4) | gcc` keeps the version relation bound to gcc-5 only, as before.
- Both commits carry a single `Signed-off-by`, matching `signed-off-by.sh`; the DCO check passes; subjects follow `[Component] Summary`.

myungjoo:
[Relayed on behalf of the author by a Claude agent]

Both suggestions applied in the follow-up commit: the 26.04 job now installs libmosquitto-dev and configures with -DMQTT_SUPPORT=ON, matching debian/rules, and CMakeLists.txt sets CMAKE_CXX_STANDARD_REQUIRED ON. The MQTT configuration was also built on a 26.04 container before this PR was opened.

myungjoo:
[Relayed on behalf of the author by a Claude agent]

Rebased onto current main (f09f0d6). The only conflict was in .github/workflows/ubuntu_clean_cmake_build.yml, where main added the AddressSanitizer steps to the build job; both are kept: the ASan steps stay in that job and the Ubuntu 26.04 container job follows it. No other file changed in the rebase.

PR #257, opened 2026-09-03
[CI] Restore the per-function Doxygen check

myungjoo:
Port of nnstreamer/nnstreamer#4914 (issue nnstreamer/nnstreamer#4908).

## What was wrong

`.github/workflows/static.check.scripts/doxygen-tag.sh` here is byte-identical to the nnstreamer copy: `local function_check_flag="f+p"` at script top level, which bash rejects, and no `set -e`, so `ctags -x --c-kinds=` ran with an empty kind list and the per-function `@brief` check has passed vacuously since the Actions import. `$report_path` was never set (an "ambiguous redirect" per file on stderr) and `$brief` was not reset between files.

## What this PR does

**Checker.** Takes the corrected checker and its self-test from nnstreamer/nnstreamer#4914 verbatim, and adds the self-test step to `static.check.yml` with the same merge-ref guard:

- definitions are checked everywhere, prototypes only in headers;
- the trailing `/**< ... */` form is recognised (declaration line, a line opening with `/**<` right after it, or after an inline body);
- block comments are tracked so a `*` at the start of a wrapped parameter line is not taken for a comment continuation;
- `report_path` defaults to `/dev/null`, `brief` resets per file;
- `test_doxygen_tag.sh` generates 18 fixtures into a temp dir; all 18 fail against the checker on `main` and pass here.

**Backlog.** Under the corrected scope the tree (24 C/C++ files, Exuberant Ctags 5.9 as in CI) has exactly one file with findings: the sixteen functions of `tests/nnstreamer-edge-custom-test.c`. They are documented in the first commit; the checker lands in the second, so each commit is green on its own.


myungjoo:
[Automated review by a separate Claude agent, relayed to this PR]

Scope: read #4908 (nnstreamer/nnstreamer#4908) and the three review rounds on the source PR nnstreamer/nnstreamer#4914 (`gh pr view 4914 --repo nnstreamer/nnstreamer --comments`), read this PR's full diff at head `fc22080` (`doxygen-tag.sh`, `test_doxygen_tag.sh`, `static.check.yml`, `tests/nnstreamer-edge-custom-test.c`), hand-traced the checker's bash logic, verified all 16 new `@brief` comments against the functions they document, and — since local `ctags` wasn't available in the primary sandbox — used a WSL environment with Exuberant Ctags 5.9 (matching this repo's CI package) to actually **run** both the old and new checker against the entire 24-file tracked C/C++ tree, rather than trust the PR description's backlog count. Also watched the live CI run to completion.

## Does it solve the stated problem?

Yes. `doxygen-tag.sh` in this repo was byte-identical to the pre-fix nnstreamer copy: `local function_check_flag="f+p"` at script top level (illegal outside a function, and with no `set -e` the script silently continued with an empty ctags kind list), `report_path` unset (ambiguous-redirect noise), and `brief`/`function_positions`/`structure_positions` never reset per file. All three are fixed here exactly as in the final, 3-times-reviewed and approved nnstreamer#4914: `case`-based flag selection (`f` — definitions, checked everywhere; `f+p` — definitions+prototypes, headers only), `report_path=${report_path:-/dev/null}`, and `brief=0`/`in_comment=0` reset inside the per-file loop (`doxygen-tag.sh:104-105`).

I diffed this PR's `has_trailing_doc()` and the `$brief`/`in_comment` state machine against the *final* state of nnstreamer's script (after all three review rounds fixed the `/**<`-adjacency bug, the pointer-parameter false-reset, and the wrapped-pointer-line false-reset) — this port matches that final version verbatim, not an earlier, still-buggy intermediate commit. That's the right thing to port.

## Independent backlog verification (not just trusting the PR body)

The PR claims "exactly one file with findings: the sixteen functions of `tests/nnstreamer-edge-custom-test.c`" under the corrected scope. I verified this directly rather than by inspection:

- Extracted `git archive upstream/main` and `git archive HEAD` into two trees, built a file list of all 24 tracked `*.c/*.h/*.cc/*.hh/*.cpp/*.hpp` files (`git ls-files`), and ran the **new** `doxygen-tag.sh` (advanced mode) from this PR against both trees, using Exuberant Ctags 5.9 in WSL.
- Against the pre-PR tree content: exactly 16 `[ERROR] ... needs @brief tag` lines, all in `tests/nnstreamer-edge-custom-test.c`, at lines 27/42/48/67/80/93/106/119/125/141/152/167/182/193/212/248 — a 1:1 match with the 16 functions this PR's first commit documents.
- Against the PR-head tree content: zero findings across the whole 24-file tree.

This confirms the backlog claim is accurate and not a false comfort from only checking the changed-file list.

## Content spot-check: all 16 `@brief` comments (commit `bfc577f`)

Checked each against its function body in `tests/nnstreamer-edge-custom-test.c`:

| Function | `@brief` | Verified against body |
|---|---|---|
| `nns_edge_custom_close` | Release the private handle | frees `peer_address` + `custom_h` — match |
| `nns_edge_custom_get_description` | Return the description string | returns `"custom"` — match |
| `nns_edge_custom_create` | Allocate the private handle | `calloc`s and sets `*priv` — match |
| `nns_edge_custom_start` | Start … resetting to unconnected | `is_connected = 0` — match |
| `nns_edge_custom_stop` | Stop … clear connected flag | `is_connected = 0` — match |
| `nns_edge_custom_connect` | Mark … as connected | `is_connected = 1` — match |
| `nns_edge_custom_disconnect` | Clear the connected flag | `is_connected = 0` — match |
| `nns_edge_custom_subscribe` | Subscription is not supported | returns `NNS_EDGE_ERROR_NOT_SUPPORTED` — match |
| `nns_edge_custom_start_discovery` | Report a found device once via callback | invokes `event_cb` with `NNS_EDGE_EVENT_DEVICE_FOUND` — match |
| `nns_edge_custom_stop_discovery` | Nothing to do beyond validating the handle | only the null check, then `NNS_EDGE_ERROR_NONE` — match |
| `nns_edge_custom_is_connected` | Report whether connected | checks `is_connected == 1` — match |
| `nns_edge_custom_set_event_cb` | Store the event callback and user data | sets both fields — match |
| `nns_edge_custom_send_data` | Discard the given data after validating | validates only, no-op otherwise — match |
| `nns_edge_custom_set_info` | Store PEER_ADDRESS; every other key rejected | `strcasecmp` gate, else `INVALID_PARAMETER` — match |
| `nns_edge_custom_get_info` | Return a newly allocated copy of PEER_ADDRESS | `nns_edge_strdup` — match |
| `nns_edge_custom_get_instance` | Return the callback table | returns `&edge_custom_h` — match |

No inaccurate description found.

## Size / scope

Proportionate and minimal: two commits, first documents the backlog (test file only), second ports the checker/self-test/workflow step verbatim. No production code path is touched — `tests/nnstreamer-edge-custom-test.c` gets comment-only additions. No unrelated module is touched. `pr_doxygen_check_skip_function_definition` (the old flag this replaces) has no remaining references anywhere in the tree after the change (checked via grep).

## Test coverage (`test_doxygen_tag.sh`, 18 fixtures)

All 18 `run_checker` calls present, matching the final nnstreamer fixture set 1:1 (negative control, `f`-vs-`f+p` scope split, all three `/**<` placements plus the adjacency-doesn't-cover-the-previous-declaration case, struct `@brief`, mid-line and wrapped-line pointer-parameter reset plus its converse (`;` inside a real comment block must not reset), per-file state reset in both directions, and missing file-level `@author`). Fixtures are generated into a temp dir, so they can't leak into a future changed-file list and self-trigger.

## Future-regression detection and CI enforcement

- The self-test step in `static.check.yml` runs unconditionally ahead of the actual `doxygen-tag.sh` invocation on `$changed_file_list`, and fails the job (`exit 1`) on any fixture mismatch — so a future edit to `doxygen-tag.sh` that reintroduces any of the fixed defects (or regresses the `/**<`/pointer-line handling) is caught here, independent of what any given PR's own changed files happen to be.
- The merge-ref guard (`static.check.yml:86-98`) is self-contained even though this is the *only* self-test step in this repo (unlike nnstreamer's `static.check.yml`, which has several to pattern-match against): the comment directly explains the mechanism — `actions/checkout` here uses `ref: ${{ github.event.pull_request.head.sha }}` (an explicit PR-head checkout, not the default merge ref), while the *workflow step list itself* is always resolved by GitHub Actions from the merge ref for `pull_request`-triggered runs. That mismatch is exactly why a tree that lacks `test_doxygen_tag.sh` needs a graceful skip rather than a hard failure, and the `elif` branch (using the same `git show --diff-filter=D … | grep -qxF` shape already established by this same file's "Identify changed files" step, just filtering deletions instead of screening them out) turns a genuine removal-without-updating-the-step into a hard error instead of a silent skip. I traced both branches by hand and both are correct.
- Verified live in the CI log (run `33715040502`, `/Checker/ doxygen-tag self-test` step): all 18 `PASS:` lines present, `test_doxygen_tag.sh: all checks passed.`, and the preceding `/Checker/ Doxygen tag check` step (the real check, over the 4 actually-changed files including `tests/nnstreamer-edge-custom-test.c`) produced no `[ERROR]` lines.

## Architecture / API docs

None needed. This repo has no `CLAUDE.md`, `CONTRIBUTING.md`, or `Documentation/` directory to update (checked: none exist), and the change is confined to a CI script, its self-test, one workflow step, and comment-only additions to a test file — no architecture or public API surface is touched.

## CI status

`gh pr checks 257`: `Static checks` — pass (35s). `DCO` — pass. `Spell Check with Typos` — pass. `Tizen GBS build on Ubuntu` (x86_64/aarch64/armv7l) — all pass. `build` — pass. The four `build (-A …)` cross-compile jobs were still pending at review time; they build C/C++ code this PR does not touch beyond adding comments to one test file, so I don't expect them to be affected, but they weren't observed to completion.

## Findings

No blocking or should-fix issues found in this port. This mirrors the outcome of the three review rounds already completed on the source PR (nnstreamer/nnstreamer#4914), which this PR ports verbatim after that PR's fixes had already landed — and independent re-verification here (live tree-wide ctags run, full self-test log, all 16 brief-vs-body checks) did not surface anything the upstream review missed or that the port introduced.

**Nit (informational, not blocking, carried over from upstream and unchanged by this port):** the structure/`@brief` check (`structure_positions` branch, just below the function check in `doxygen-tag.sh`) was not given the same `has_trailing_doc` treatment as the function check — a struct documented only via a trailing `/**<` would still be flagged. This mirrors nnstreamer's own final, approved version exactly (not a regression introduced here), so it's out of scope for this PR; noting it only for completeness in case a future PR wants to extend `has_trailing_doc` to structs in both repos together.

## Verdict

**Mergeable.** The diagnosis is correct and independently reproduced (not just re-stated), the fix restores the check to the same scope and correctness already validated across three review rounds on the source PR, the entire backlog under the corrected scope is documented in comment-only changes verified accurate against every function body, the self-test is wired as a required, blocking step with a self-contained guard, and CI is green on every check this PR's content can plausibly affect.

Issue #259, opened 2026-09-07
Memory-safety bugs in libnnstreamer-edge: OOB reads in deserializers, UAF/double free in connection lifecycle, thread/handle leaks


myungjoo:
## Status (checked against `main` @ 8ea7fe1, 2026-09-10)

Ticked items are merged into `main` and were re-verified in the source, not only by PR title. Items 11 and 12 and the receive-timeout note come from the comments below.

**Memory-safety items**

- [x] 1. Metadata deserializer OOB read — #260
- [x] 2. Serialized data header trusted (short buffer, size overflow, unchecked `memdup`, leak into non-empty handle) — #261
- [x] 3. `nns_edge_parse_host_string()` on peer data (NULL, no NUL, no `':'`, uninitialised port; TCP server + hybrid client) — #262
- [x] 4. Capability string not NUL-terminated — #262
- [x] 5. Peer-controlled allocation sizes / unbounded MQTT queue — #267
- [x] 6. Uninitialised header bytes serialised — #261
- [x] 7. Dangling `src_conn` / `sink_conn` after a rejected accept → double free — #265
- [x] 8. Message thread removes its own connection: self-join, thread leak, UAF after release — #264
- [x] 9. Connection list and `caps_str` touched from three threads without a lock — #266
- [x] 10. Smaller leaks (all five bullets: `broker_h` overwrite on reconnect, double `nns_edge_start()`, MQTT `msg` leak on failed push, `clear_info()` dangling metadata, `strlen()` on a failed `nns_edge_get_host_string()`) — #263
- [x] 11. Edge data without metadata never deserializes (from the comments) — #261
- [ ] 12. `_nns_edge_message_handler()` and both MQTT callbacks ignore the metadata / data deserialize result (from the comments) — #270 (open)

**Related non-memory issues** (`A1`-`A8`)

These are tagged with an `A` prefix so that they cannot be read as the memory-safety numbers above: `A`*n* is simply the *n*-th entry of this list, in the order it has always had. Earlier comments that point at one of them by position ("the third of the related non-memory issues") predate the tags and were written against a different order; the tags below are what to go by.

- [x] A1. `conn->running` raised by the thread itself → `pthread_join()` never returns — #263 (b06ded6)
- [x] A2. Lost wake-up between `nns_edge_queue_clear()` and the send thread — #263 (d33aae0)
[...]

Issue #258, opened 2026-09-04
nns_edge_custom_load() dereferences NULL callbacks of a partially implemented custom library


myungjoo:
`_load_custom_library()` accepts any instance a custom library hands back without checking that the callbacks in it are actually set, and the wrappers then call through them unconditionally. A custom library that implements only part of `nns_edge_custom_s` crashes the calling process with a NULL function-pointer dereference.

Found while working on #253 / #254; it is a different failure mode from those and needs a change under `src/`, so it is filed separately.

## The problem

`nns_edge_custom_s` is a struct of 15 function pointers, filled in by the library author. The in-tree example uses designated initializers, so any member an author leaves out is NULL:

```c
nns_edge_custom_s edge_custom_h = {
  .nns_edge_custom_get_description = nns_edge_custom_get_description,
  .nns_edge_custom_create = nns_edge_custom_create,
  ...
};
```

`_load_custom_library()` validates exactly three things: `dlopen()` succeeded, the `nns_edge_custom_get_instance` symbol exists, and `get_instance()` returned non-NULL. It never looks inside the instance. `nns_edge_custom_load()` then does:

```c
  custom_h = custom->instance;

  ret = custom_h->nns_edge_custom_create (&custom->priv);   /* nnstreamer-edge-custom-impl.c:107 */
```

A library that does not set `nns_edge_custom_create` crashes here, immediately, inside `nns_edge_custom_create_handle()`. Eleven call sites are unguarded in the same way:

| callback | call site | guarded |
| --- | --- | --- |
| `create` | `nnstreamer-edge-custom-impl.c:107` | no |
| `close` | `:134` | no |
| `start` | `:166` | no |
| `stop` | `:189` | no |
| `set_event_cb` | `:213` | no |
| `start_discovery` | `:236` | no |
| `stop_discovery` | `:259` | no |
| `connect` | `:282` | no |
| `disconnect` | `:305` | no |
| `is_connected` | `:327` | no |
| `send_data` | `:350` | no |
| `set_info` | `:377` | **yes** (`if (custom_h->nns_edge_custom_set_info)`) |
| `get_info` | `:406` | **yes** |

For `create` the crash is at load time. For the others it is deferred to whichever API the application calls first — `nns_edge_start_discovery()` on a point-to-point transport that had no reason to implement discovery, for instance.

This is a robustness bug, not a security one: reaching it requires loading a library of the caller's choosing, which already implies running its code.

[...]

PR #260, opened 2026-09-07
[Fix] bound the metadata deserializer by the buffer length

myungjoo:
Fixes item 1 of #259.

## The bug

`nns_edge_metadata_deserialize()` trusted the buffer it was handed, and that buffer comes straight off the wire. `_nns_edge_message_handler()` passes the peer-supplied `cmd.meta` / `meta_size` of a TCP command into it via `nns_edge_data_deserialize_meta()`, and both MQTT message callbacks reach it via `nns_edge_data_deserialize()`. Neither path authenticates first, so all three reads below were remotely reachable:

- the four-byte pair count was read after checking only `data_len != 0`, so a 1-3 byte payload already read out of bounds;
- the loop ran while `cur < data_len || meta->list_len < total` — with the disjunction, a peer announcing `0xFFFFFFFF` pairs kept the loop alive after `cur` passed `data_len`, walking `strlen()` across the heap until it hit an unmapped page;
- `strlen()` on the key and the value was never bounded by `data_len`, so a payload whose last pair is not NUL-terminated ran off the end even while `cur` was still inside the buffer.

## The fix

The count is read only after `data_len` is known to hold it, and with `memcpy` — in `nns_edge_data_deserialize()` the metadata blob follows raw memories of arbitrary length, so the pointer is not necessarily aligned. Pairs are walked with `memchr` limited to the bytes that remain, which cannot leave the allocation and always advances, so the loop terminates on any input. A buffer is accepted only when it yields exactly the announced number of pairs and is consumed to its last byte; anything else is rejected and leaves the handle empty rather than half-parsed.

Well-formed input is unaffected: `nns_edge_metadata_serialize()` writes the count and consumes the buffer exactly, which is what the new checks require. No API or ABI change.

## Tests

Every buffer the new tests hand to the deserializer is allocated at exactly the length passed as `data_len`, so a read that leaves the buffer is a heap-buffer-overflow AddressSanitizer reports rather than a silent pass — and the ubuntu cmake job already runs the suite under ASan, so a regression fails CI.

- `edgeMeta`: short header; the `0xFFFFFFFF` count from #259; unterminated key and value; count larger than the pairs present; count of zero with bytes left over; empty key; duplicate keys. Two positive tests keep the parser from becoming merely strict — an empty blob still deserializes, and a handle that rejected a malformed buffer is left empty and still usable.
- `edgeData`: the two public entry points that carry peer data in. `deserializeInvalidParam05_n` is the TCP path; `06_n` is the MQTT path, corrupting the metadata pair count in place so the size checks in `nns_edge_data_is_serialized()` still pass and the malformed blob reaches the deserializer. `deserializeMetaBoundary` is the positive counterpart, with a three-byte third memory so the metadata section starts at an odd offset — the unaligned read the fix handles with `memcpy`.

Against the unfixed deserializer, ten of these fail, nine with an ASan heap-buffer-overflow in `strlen()` or in the header read. `plain` and `asan` builds both pass 100% locally.

## Note on the first commit

`tests/unittest_nnstreamer-edge.cc` predates the GitHub-Action static checkers and has never been run through `clang-format`, so it does not match the repository's `.clang-format` (761 lines of diff, mostly `TEST(` -> `TEST (` and include sorting). The checker reformats every changed `.cc` file and fails on a non-empty diff, so any PR touching this file fails static.check before its own changes are looked at. The first commit formats it once, with no behavioural change, kept separate so the two commits that follow stay readable. clang-format 14 and 16 produce identical output here.


myungjoo:
**Note:** This review was produced by a separate automated review agent (not the PR author's own agent), asked to independently audit this PR against issue #259 item 1. It is being transcribed here verbatim as its comment.

## Scope checked

Reviewed the 3 commits (`12b4d50` format-only, `647c2b9` fix, `601aa59` tests) against `src/libnnstreamer-edge/nnstreamer-edge-metadata.c`, its callers in `nnstreamer-edge-data.c`/`nnstreamer-edge-internal.c`, the repo's CI config (`.github/workflows/static.check.yml`, `ubuntu_clean_cmake_build.yml`), and built + ran the test suite locally (plain and `-fsanitize=address`, both 100% pass).

## Does it solve the stated problem? ✅

All three OOB-read mechanisms from #259 item 1 are fixed:
- Header count is now read only after `data_len >= sizeof(uint32_t)` is confirmed, and via `memcpy` (avoids an unaligned read — this path is reached with an arbitrary byte offset since the metadata blob follows raw memories of arbitrary length in `nns_edge_data_deserialize()`).
- The `||` loop condition is gone; `nns_edge_metadata_read_string()` bounds every read with `memchr(str, '\0', len - *cur)`, so it can never read past `data_len`, and `cur` strictly increases on every successful read, so the loop is guaranteed to terminate.
- `strlen()` on key/value is replaced by the same bounded `memchr`, so an unterminated trailing string is rejected instead of read past the buffer.
- A buffer is accepted only if it produces exactly `num` pairs *and* consumes exactly `data_len` bytes — this also catches the `0xFFFFFFFF`-pair-count PoC from the issue, and duplicate keys (which can't reach the declared count since `nns_edge_metadata_set` dedupes in place).

I confirmed this against both the vulnerable and fixed shapes by running the added tests under ASan locally (`run259.sh asan`) — 100% pass on the fixed tree.

## Regression risk — none found (Blocker/Major: none)

- `nns_edge_metadata_serialize()` (unchanged) always writes the exact count and consumes the buffer exactly on the way out, which is exactly what the new deserializer requires — well-formed round-trips are unaffected. Verified directly with `deserializeRecovery` and `deserializeMetaBoundary` (odd-offset/unaligned case) locally.
- No API/ABI change — `nns_edge_metadata_deserialize()`'s signature is untouched, the new `nns_edge_metadata_read_string()` is `static`.
- Diff is confined to `nnstreamer-edge-metadata.c` + the test file; no other module is touched.

## Test coverage — thorough (Blocker/Major: none)

11 negative + 2 positive cases at the `edgeMeta` level (short header, the `0xFFFFFFFF` PoC, unterminated key/value, count too large, count zero with leftover bytes, empty key, duplicate keys, empty blob, recovery-after-failure), plus `edgeData` level coverage of both ingress paths that actually carry peer data (`deserializeInvalidParam05_n` = TCP via `nns_edge_data_deserialize_meta()`, `06_n` = MQTT via `nns_edge_data_deserialize()` with the pair count corrupted in place so `nns_edge_data_is_serialized()`'s size check still passes it through), plus a positive `deserializeMetaBoundary` case that exercises the unaligned-pointer path. Every buffer is allocated at exactly `data_len`, so a future regression that reads even one byte past the end fails as an ASan heap-buffer-overflow rather than passing silently.

## Will a future regression be caught by CI? ✅

Confirmed independently (not just taking the PR body's word for it):
- `.github/workflows/ubuntu_clean_cmake_build.yml` does run the suite under `-fsanitize=address` as part of the `build` job (not a separate job — matches this repo's practice of folding ASan into the required `build` check rather than an optional job, since only `DCO`/`build` are required branch-protection checks here).
- `.github/workflows/static.check.yml` does `git diff -- *.cc *.hh *.hpp *.cpp` after running `clang-format -i` and fails the job (exit 1) on any diff — so commit 1's rationale for reformatting `unittest_nnstreamer-edge.cc` once, separately, is correct: any PR touching that file would otherwise fail static.check on unrelated pre-existing formatting drift. I verified locally with clang-format 16.0.6 that the file as committed produces a zero-line diff, matching the PR body's claim that clang-format 14 and 16 agree here.

## Minor / Nit findings

**Minor** — Inconsistent state on failure: in `nns_edge_metadata_deserialize()`, when `data_len < sizeof(uint32_t)` the function returns early *without* calling `nns_edge_metadata_free(meta)`, so a handle that already held metadata keeps its old contents after this "failed" call. Every other failure path (`nns_edge_metadata_free(meta)` is called unconditionally right after the size check, and again on the final mismatch check) leaves the handle empty. The function's contract on failure is therefore not consistent (sometimes preserves old state, sometimes wipes it), depending on exactly how the input is malformed. This isn't exploitable and doesn't appear to matter for the current callers (they hand in a freshly-created handle), but it's a sharp edge for future callers who might rely on "deserialize failed ⇒ handle unchanged" or "⇒ handle empty" as a guarantee either way. Worth a one-line fix (e.g. also skip the free on the size-check path being made consistent with the rest, or vice versa) but not blocking this PR.

**Info, out of scope for this diff** — While tracing the callers I reproduced (locally, both before and after this fix — it is unaffected by this PR) that `nns_edge_data_deserialize()` unconditionally calls `nns_edge_metadata_deserialize(ed->metadata, ptr, header->meta_len)` even when `header->meta_len == 0`, i.e. when the sender never set any metadata at all (the common case). Since `nns_edge_metadata_serialize()` emits 0 bytes (not even the 4-byte count) when there is no metadata, and the deserializer (both before and after this PR) rejects any buffer shorter than `sizeof(uint32_t)`, a round trip of plain data with zero metadata entries fails with `NNS_EDGE_ERROR_INVALID_PARAMETER`. This predates this PR (the old check was `data_len <= 0`, which already excluded 0) so it is **not a regression introduced here** and I'm not asking for it to be fixed in this PR — but it's a real, currently-reproducible bug reachable from ordinary (non-malicious) use, and none of the existing or newly-added tests cover the meta-less round trip (every serialize/deserialize test in the suite calls `nns_edge_data_set_info()` at least once first). Recommend filing it as a separate, dedicated issue (or a new item under #259) so it doesn't get lost, and adding a positive test for it wherever it gets fixed.

**Nit** — `nns_edge_metadata_read_string()` does `(size_t)(len - *cur)` where both are `nns_size_t` (`uint64_t`); on a 32-bit target where `size_t` is 4 bytes this narrows silently. Not exploitable in practice (buffer sizes here can't realistically approach 4 GiB), just noting it for completeness.

**Nit** — The new tests use `unsigned int`/`sizeof (unsigned int)` for the 4-byte pair count rather than `uint32_t`, unlike the production code they're testing. Harmless on every realistic build target, purely stylistic.

## Documentation

No architecture or public API/ABI change, so no accompanying doc update is needed here, and none is missing.

## Verdict

**Approve.** No Blocker or Major issues. The fix correctly and completely closes item 1 of #259, is tightly scoped, well tested (including the exact PoC from the issue), verified to build and pass under both plain and ASan configurations locally, and its CI-related claims (ASan coverage, clang-format gate) check out against the actual workflow files. The two Minor/Info points above are suggestions for follow-up, not blockers for this PR.

myungjoo:
Thanks — all four points addressed below. `523d210` is the only code change.

**Minor, inconsistent state on failure — fixed in `523d210`.** The handle is now cleared before the size check, so every failure path leaves it empty. That direction rather than "unchanged on failure" because the input arrives from a peer: metadata from an earlier message should not survive a buffer that was rejected. The contract is now stated in the `nnstreamer-edge-metadata.h` declaration, and `deserializeRecovery` asserts it for both shapes — the short-header path and a malformed pair — by setting a key first and checking it is gone after each rejected call.

**Info, meta-less round trip — not fixed here, agreed it needs tracking.** Reproduced independently while preparing this PR: `nns_edge_data_serialize()` on a handle with no info yields `meta_len = 0`, and `nns_edge_data_deserialize()` then returns `NNS_EDGE_ERROR_INVALID_PARAMETER` even though the memories parsed fine (`count = 1`). Every MQTT publish from a handle that never called `nns_edge_data_set_info()` is dropped by the subscriber. It predates this PR and the fix belongs with the header validation of #259 item 2 — `nns_edge_data_deserialize()` has to learn that `meta_len == 0` means "no metadata", not "malformed" — so folding it in here would mean changing a second function's semantics inside a bounds-checking fix. It will be raised on #259 rather than left to this PR.

**Nit, `(size_t)(len - *cur)` on a 32-bit target — no change.** The narrowing can only lose bits when `len` exceeds `SIZE_MAX`, and any truncation shortens the `memchr` range, so the failure mode is a rejected buffer, never a read past the end. On a 32-bit target a `data_len` above 4 GiB cannot describe a real allocation, so the only inputs that could narrow are ones that must be rejected anyway.

**Nit, `unsigned int` vs `uint32_t` in the tests — no change.** The surrounding tests in this file already write the pair count as `((unsigned int *) data)[0]`; matching them keeps the new cases consistent with the block they sit in.

`plain` and `asan` both still pass 100% locally, and `clang-format` and GNU `indent` both report a zero-line diff on the changed files.

myungjoo:
**Note:** This is a second review pass, produced by a separate automated review agent (not the PR author's own agent, and not the same agent that wrote the first review comment above). It focuses on commit `523d210` (the response to the first review's Minor finding) and re-checks the author's three "not fixing here" decisions. Built and ran the full suite locally (plain and `-fsanitize=address`) on the worktree at `fix/issue-259-metadata-bounds`.

## Verification performed

- `cmake --build` (Debug) + `ctest` on both a plain config and an ASan config (`-fsanitize=address -fno-omit-frame-pointer` for C/C++ compile+link): **100% tests passed (2/2 suites)** in both configs.
- Read the full current `nns_edge_metadata_deserialize()` and `nns_edge_metadata_free()`/`nns_edge_metadata_read_string()` and traced every caller (`nns_edge_data_deserialize_meta`, `nns_edge_data_deserialize` in `nnstreamer-edge-data.c`, and `_nns_edge_message_handler` in `nnstreamer-edge-internal.c`).
- Ran `clang-format` and (informationally) GNU `indent` against the changed files to sanity-check the formatting claims.

## `523d210` — moving `nns_edge_metadata_free()` before the size check

**Verdict: correct, no new issues (no Blocker/Major/Minor).**

- Order is still safe: the `!meta` (NULL handle) check remains *before* the free, so there's no new NULL-deref path. `nns_edge_metadata_free()` is idempotent (resets `list`/`list_len` and walks a possibly-NULL list), so calling it here and then again later on other failure branches is not a double-free.
- Checked both real callers for a regression from the changed failure contract:
  - `nns_edge_data_deserialize_meta()` and `nns_edge_data_deserialize()` (`nnstreamer-edge-data.c:512,632`) both just propagate `nns_edge_metadata_deserialize()`'s return code to their own caller with no branch that depended on "metadata unchanged on failure" — nothing to break.
  - `_nns_edge_message_handler()` (`nnstreamer-edge-internal.c:774`) calls `nns_edge_data_deserialize_meta()` on a `data_h` that was created fresh two lines above (`nns_edge_data_create(&data_h)`) inside the same loop iteration and destroyed at the end of it — there is no scenario in the current codebase where a handle carrying real prior metadata is re-deserialized in place, so the "always empty on failure" contract change is currently a no-op in practice and a genuine robustness improvement for future callers, exactly as intended.
- New `deserializeRecovery` additions are sound: they seed `stale-key`, feed a short-header buffer (2 bytes, below `sizeof(uint32_t)`) and separately a malformed-pair buffer, and assert `nns_edge_metadata_get()` fails (key gone) after *each*. `nns_edge_metadata_get()` doesn't touch `*value` on failure and the test correctly never frees `value` on those paths, so no leak/UB in the test itself.
- Doc update in `nnstreamer-edge-metadata.h` (`@note` on the contract) is appropriate and sufficient — this is a private/internal header (not installed, not part of the public API surface), so no public docs need touching.

## The three "won't fix" responses

**1. `meta_len == 0` round-trip bug → deferred to a new #259 item.** Reasonable. Confirmed independently: `nns_edge_metadata_serialize()` emits 0 bytes when `list_len == 0`, and `nns_edge_metadata_deserialize()` (both before and after this PR) rejects any `data_len < sizeof(uint32_t)`, so a plain data message from a handle that never called `nns_edge_data_set_info()` fails deserialization. This is pre-existing (unaffected by this bounds-check PR either way) and, as the author notes, a real fix means teaching `nns_edge_data_deserialize()` that `meta_len == 0` is a valid "no metadata" case rather than a malformed buffer — that's a semantic change to a different function, out of scope for a bounds-fix PR. Agree it should stay tracked separately rather than be folded in here.

**2. `(size_t)(len - *cur)` narrowing nit → no change.** Verified this holds up: `nns_size_t` is a fixed `uint64_t` (`include/nnstreamer-edge-data.h:23`), and both call sites that ever reach `nns_edge_metadata_deserialize()` bind `data_len` to the size of a buffer that was actually allocated and fully populated (`nns_edge_data_is_serialized()`'s `total != data_len` check for the MQTT/generic path, and the `nns_edge_malloc(cmd->info.meta_size)` + `_receive_raw_data()` pair for the TCP path, which bails via `NNS_EDGE_ERROR_OUT_OF_MEMORY`/`IO` before ever calling deserialize if the size is unreasonable). So `len - *cur` can never exceed the size of memory that was actually allocated in the current process, which on a 32-bit target is always representable in `size_t`. The narrowing is unreachable for any input that isn't already going to be rejected. Agree with leaving it as-is.

**3. `unsigned int` vs `uint32_t` in tests → no change.** Confirmed the surrounding pre-existing tests in the same file already write the pair count via `((unsigned int *) data)[0]`; matching that convention within the same test block is the right call over introducing a one-off inconsistency. Purely stylistic either way — agree.

## Scope / regression / CI (re-confirmed)

- `523d210`'s diff is still confined to `nnstreamer-edge-metadata.c`, `nnstreamer-edge-metadata.h`, and the test file — no other module touched, consistent with the rest of the PR.
- No API/ABI change (internal-only header, signature untouched).
- Both plain and ASan builds pass 100% locally on this exact commit, matching the author's claim.

## Verdict

**Approve.** No Blocker, Major, or Minor issues found in `523d210` or in the three items the author chose not to change. The fix for the first review's Minor finding is correct, minimal, well-documented, and covered by a test that actually exercises both failure shapes. The three deferred/declined items are all well-reasoned and appropriately scoped out of this PR.

myungjoo:
**Note:** This review was produced by a second, independent automated review agent (not the PR author's own agent, and not the agent that wrote the first review above). It is being transcribed here verbatim as its comment, per the reviewer's request. It focuses on commit `523d210` (the response to the first review's Minor finding) and on evaluating the author's "no change" replies to the remaining three points.

## Scope checked

Re-read all 4 commits against `src/libnnstreamer-edge/nnstreamer-edge-metadata.c`/`.h`, the two real ingress paths (`nns_edge_data_deserialize_meta()` in `nnstreamer-edge-data.c` and `nnstreamer-edge-internal.c:774`), and the CI workflows. Built and ran the suite locally on this exact commit (`523d210` included):

- `plain`: both `unittest_nnstreamer-edge` and `unittest_nnstreamer-edge-custom` pass, 2/2.
- `asan` (`-fsanitize=address -fno-omit-frame-pointer`, `ASAN_OPTIONS=detect_leaks=1`): `unittest_nnstreamer-edge` — the binary containing every `edgeMeta`/`edgeData` test this PR added or touches — reports `[==========] 184 tests from 9 test suites ran.` / `[  PASSED  ] 184 tests.` / `Test Passed.`, with zero AddressSanitizer reports anywhere in the log. This directly confirms the PR body's and the first review's claim for the code this PR actually changes.
- Independently re-verified `clang-format 16.0.6` produces a zero-line diff on `tests/unittest_nnstreamer-edge.cc` as committed, matching the PR body's rationale for the separate format-only commit.
- Aside, not a finding against this PR: `unittest_nnstreamer-edge-custom` hung under this same ASan build at `edgeCustom.expectedReturn` (had to be killed after >20 min). That binary and test file are untouched by this PR's diff — the hang sits in the custom-connection code path, which lines up with the threading/lifecycle bugs #259 already catalogs separately (items 7-9) and the custom-connection issues explicitly deferred to #258. Mentioned only for repo-health awareness; it does not bear on this diff.

## 1. Does `523d210` correctly fix the Minor finding? Yes.

Before: `nns_edge_metadata_free(meta)` ran only *after* the `data_len < sizeof(uint32_t)` check, so a too-short buffer returned `NNS_EDGE_ERROR_INVALID_PARAMETER` while leaving any previously-held metadata on the handle untouched — inconsistent with every other failure path in the function, which cleared the handle.

After: the unconditional `nns_edge_metadata_free(meta)` moved to the top, immediately after the `!meta` check and before the size check. Every failure path — short buffer, malformed pair, count/length mismatch — now clears the handle first. Confirmed in the current source (`nnstreamer-edge-metadata.c:364-372`):

```c
if (!meta)
  return NNS_EDGE_ERROR_INVALID_PARAMETER;

nns_edge_metadata_free (meta);

if (!data || data_len < sizeof (uint32_t)) {
  ...
  return NNS_EDGE_ERROR_INVALID_PARAMETER;
}
```

No double-free/UAF risk from calling `nns_edge_metadata_free()` more than once in a single invocation (e.g. once here, again on a later failure path at `:390`/`:397`): the function sets `meta->list = NULL` and `meta->list_len = 0` right after unlinking the list (`:90-92`), so a second call in the same invocation walks an already-NULL list and is a no-op. Confirmed by reading `nns_edge_metadata_free()` directly.

The chosen contract ("always empty on failure" rather than "unchanged on failure") is the right one for peer-supplied data — a message rejected as malformed should not leave a previous message's metadata sitting on the handle — and it's now documented at the point of change (`nnstreamer-edge-metadata.h:56-58`) with an explicit `@note`. Since this header is marked `DO NOT export this file`, there is no public API/ABI surface affected — a doc comment at the declaration is the right (and sufficient) amount of documentation here.

The new `deserializeRecovery` case exercises both failure shapes: a 2-byte truncated buffer and a well-formed-looking-but-malformed 1-pair buffer, each preceded by `nns_edge_metadata_set(meta, "stale-key", ...)` and followed by an assertion that the key is gone. This is exactly the right regression test for the new contract — it would fail without `523d210`, since the truncated-buffer branch used to leave `stale-key` in place.

Checked both real callers for anything that depended on the old, inconsistent contract: `nns_edge_data_deserialize_meta()` and `nns_edge_data_deserialize()` (`nnstreamer-edge-data.c:512,632`) both simply propagate the return code with no branch on "metadata unchanged." `_nns_edge_message_handler()` (`nnstreamer-edge-internal.c:774`) calls into this on a `data_h` created fresh a couple of lines above and destroyed at the end of the same iteration — there is no code path today where a handle carrying real prior metadata gets re-deserialized in place, so the contract change is a no-op in current behavior and a genuine robustness improvement for future callers, exactly as intended.

## 2. Does moving the free earlier change behavior for the `meta_len == 0` case? No regression.

Traced `nns_edge_data_deserialize()` (`nnstreamer-edge-data.c:632`), which calls `nns_edge_metadata_deserialize(ed->metadata, ptr, header->meta_len)` unconditionally, including when `header->meta_len == 0` (a handle that never called `set_info()`). That call fails both before and after this PR (`data_len < sizeof(uint32_t)` rejects 0), so `523d210` doesn't change whether this case fails — only that `ed->metadata` is now guaranteed empty afterward instead of possibly retaining a previous successful deserialize's metadata. For the normal call pattern (a freshly created `dest_h` per message) this has no observable effect.

## 3. Are the three "not fixing" decisions reasonable?

- **`meta_len == 0` round-trip bug → deferred to a new #259 item: reasonable.** Confirmed independently: `nns_edge_metadata_serialize()` emits 0 bytes when there's no metadata, and the deserializer (both before and after this PR) requires `data_len >= sizeof(uint32_t)`, so a plain send with no metadata is rejected on the receive side. This is a real, pre-existing, non-security bug — but it lives in `nns_edge_data_deserialize()`'s handling of `meta_len == 0`, not in `nns_edge_metadata_deserialize()` itself, whose "reject anything under 4 bytes" behavior is correct in isolation. Fixing it here would mean `nns_edge_data_deserialize()` special-casing `meta_len == 0` before calling into the metadata deserializer — a second function's semantics, which the author correctly ties to #259 item 2 (header validation) instead of this bounds-checking PR. Worth confirming it actually gets filed so it isn't lost, but correctly out of scope here.

- **`(size_t)(len - *cur)` narrowing nit → not fixing: reasonable.** `nns_size_t` is `uint64_t` (`include/nnstreamer-edge-data.h:23`). On a 32-bit target where `size_t` is 32 bits, the cast can only lose information when the true remaining length exceeds `SIZE_MAX` (~4 GiB); a `memchr` given a *smaller* count than the true remaining length can only search fewer bytes, never more, so a truncated value can at worst cause a spurious rejection, never an out-of-bounds read. A 32-bit process cannot legitimately hold a >4 GiB contiguous allocation in the first place, so any `data_len` that large is already bogus input that must be rejected regardless. Correct to leave as-is.

- **`unsigned int` vs `uint32_t` in the new tests → not fixing: reasonable.** Consistent with the pre-existing style in the surrounding test cases in this file (`((unsigned int *) data)[0]` appears throughout the tests adjacent to the new ones). Pure style, zero behavioral risk on any target this project builds for.

## Regression / scope check (Blocker/Major: none)

- `523d210`'s diff is confined to `nnstreamer-edge-metadata.c`, `nnstreamer-edge-metadata.h`, and the test file — no other module touched, consistent with the rest of the PR.
- `nns_edge_metadata_serialize()` is untouched; well-formed round-trips are unaffected (`deserializeRecovery`'s tail and `deserializeMetaBoundary` both still pass).
- No signature change to `nns_edge_metadata_deserialize()`; the new `nns_edge_metadata_read_string()` helper is `static` and unexported.

## CI enforcement (re-confirmed independently)

`.github/workflows/ubuntu_clean_cmake_build.yml`'s `build` job — the one covered by branch protection alongside `DCO` — runs a `-fsanitize=address` configure/build/test step in addition to the plain one, and the local ASan run above confirms it actually catches what it's meant to (100% pass, no ASan reports, on this exact commit). `static.check.yml`'s clang-format gate does `exit 1` on any diff in changed `.cc`/`.hh`/`.hpp`/`.cpp` files, which is why the separate format-only commit is necessary and correctly kept apart from the behavioral change.

## Verdict

**Approve.** `523d210` correctly and minimally fixes the Minor inconsistency raised in the first review, with no double-free/UAF risk and a regression test that would fail without it. The three "won't fix" responses are all technically sound — none represents a live correctness or security gap, and the two left as pure nits are genuinely inert given this project's target platforms. Locally verified: plain and ASan builds both pass 100% on the current commit for the test binary this PR actually touches. No Blocker/Major/Minor issues remain open against this PR.

myungjoo:
One correction to the aside in the second review above: I could not reproduce the `edgeCustom.expectedReturn` hang.

On the same ASan build (`-fsanitize=address -fno-omit-frame-pointer`) of this branch, `unittest_nnstreamer-edge-custom` runs 32 tests from 1 test suite in 2 ms, all passing, and `--gtest_filter=*expectedReturn*` on its own passes in 0 ms. CI agrees — the `build` and `Ubuntu 26.04` jobs both run the suite and finished in 45-47 s. So the hang looks like an artifact of that reviewer's local run rather than a repo-health signal, and nothing here points at #258 or at items 7-9 of #259. Those items are real, but this is not evidence of them.

Everything else in that review stands. All checks are green on `e1c10d7`: `build` and `DCO` (the two required checks), `Static checks`, `Build and test on Ubuntu 26.04 (gcc 15)`, all three Tizen GBS builds, the three integration-test builds, and `Spell Check`.

The only change since the second review is `e1c10d7`, which turns a two-line `/* ... */` into a single-line comment because the doxygen-tag checker requires multi-line comments to open with `/**`. No code change.

myungjoo:
Thanks. All three follow-ups are non-blocking as noted, and none of them changes this PR.

1 and 2 are now filed on the issue as items 11 and 12: https://github.com/nnstreamer/nnstreamer-edge/issues/259#issuecomment-5568145277 — with the measurement for the meta-less round trip and a note on why the fix belongs in `nns_edge_data_deserialize()` rather than in the deserializer this PR touches. Accepting `data_len == 0` here would contradict `edgeMeta.deserializeInvalidParam03_n` and `edgeData.deserializeInvalidParam04_n`, which both pass `0` and expect an error, and it would make the deserializer accept a buffer the serializer cannot emit. Both items land naturally with item 2 of #259, since they are the same "trusts the header" problem on the same path. The round-trip test with no info set will come with that fix — it cannot be added here, because it would have to assert the broken behaviour.

3 is a fair thing to state explicitly; it is recorded on the issue as a note so the new bounds checking is not mistaken for covering it.

PR #261, opened 2026-09-07
[Fix] validate the serialized edge data header against the buffer

myungjoo:
Addresses items **2** and **6** of #259.

Both MQTT message callbacks (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`) call `nns_edge_data_deserialize()` on any payload with `payloadlen > 0`, so the serialized header is fully peer controlled.

### Item 2 - the deserializer trusted the header

* `nns_edge_data_is_serialized()` read `key`, `version`, `num_mem`, `data_len[]` and `meta_len` before checking the buffer length, so a 1 byte publish made the subscriber read ~2 KB past a 1 byte heap allocation. The length is now checked first.
* The size check summed `meta_len` and every `data_len[n]` into a plain `uint64_t`. Values chosen to wrap the sum back to the real buffer length passed the check, and `nns_edge_data_deserialize()` then advanced its pointer by those values and called `nns_edge_memdup()` on a wild pointer. The total is now accumulated within `data_len`, so it can never exceed the buffer, and empty memories are rejected (`nns_edge_data_add()` rejects them as well, so the serializer never emits one).
* `nns_edge_data_deserialize()` ignored the `nns_edge_memdup()` result, leaving a `NULL` data pointer with a non-zero length on allocation failure, and overwrote `num` / `data[]` without releasing the previous content of the handle. It now clears the handle first, keeps `num` in step with the entries that are filled in, and returns `NNS_EDGE_ERROR_OUT_OF_MEMORY` when a copy fails.
* A buffer without metadata has `meta_len == 0`, which `nns_edge_metadata_deserialize()` rejects as an invalid parameter, so a round trip of a handle carrying no metadata failed. The metadata is reset instead of deserializing an empty buffer.

### Item 6 - uninitialised stack bytes were serialised

`nns_edge_data_serialize()` filled `data_len[0..num)` of a stack allocated header and copied the whole struct out, sending the remaining entries and the struct padding - about 2 KB of uninitialised stack - to the broker on every publish. The header is cleared before it is filled.

### Tests

Every new negative test fails on the code before the fix, verified with an ASan build that keeps the new test file and reverts only `nnstreamer-edge-data.c` to `main`:

| test | failure on `main` |
| --- | --- |
| `edgeDataIsSerialized.invalidParam04_n`, `invalidParam05_n`, `edgeDataDeserialize.invalidParam05_n` | `AddressSanitizer: heap-buffer-overflow` |
| `edgeDataIsSerialized.invalidParam06_n`, `edgeDataDeserialize.invalidParam06_n` | overflowed sizes accepted, 16 EiB allocation requested |
| `edgeDataIsSerialized.invalidParam07_n` | empty memory accepted |
| `edgeDataSerialize.headerIsCleared` | uninitialised stack bytes in the header |
| `edgeDataDeserialize.replaceHandleContent` | previous content leaked, stale metadata |
| `edgeDataSerialize.noMetadata`, `edgeDataDeserialize.copyDeserialized` | round trip without metadata rejected |
| `edgeDataDeserialize.invalidParam07_n` | raw memories left in the handle after the metadata was rejected |

`invalidParam08_n`, `invalidParam09_n`, `invalidParam10_n` and `invalidParam11_n` cover the remaining branches of the size, version and memory count checks rather than a past bug.

`nnstreamer-edge-data.c` coverage is 94.2% of lines and 83.9% of branches; what is left uncovered are `malloc()` failure paths that need fault injection.

### Notes

* The first commit is a mechanical `clang-format` pass over `tests/unittest_nnstreamer-edge.cc`. The file predates the clang-format checker, so any PR touching it fails `Static checks` with an unrelated 197 line diff. #260 carries the same commit; whichever lands first makes the other a no-op.
* The MQTT callbacks still ignore the `nns_edge_data_deserialize()` return value and raise `NNS_EDGE_EVENT_NEW_DATA_RECEIVED` with an empty handle when a payload is rejected. That is pre-existing behaviour and not a memory safety issue, so it is left out of this change.


myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Scope of review

Checked whether this PR resolves items **2** and **6** of #259, whether it introduces any regression to other modules, whether its size/footprint matches the topic, and whether the accompanying tests are sufficient to both prove the fix and catch future regressions (including whether CI actually gates on them).

## Item coverage (#259)

Verified by reading `src/libnnstreamer-edge/nnstreamer-edge-data.c` against each sub-bug listed under item 2 and item 6 of #259:

- **Item 2a** (header fields read before `data_len` is checked against `sizeof(header)`): fixed — `nns_edge_data_is_serialized()` now checks `data_len < sizeof (nns_edge_data_header_s)` before touching `header->key`/`version`/etc.
- **Item 2b** (`total += header->data_len[n]` plain `uint64_t` sum wraps around and bypasses the size check): fixed — the running `total` is now bounded by accumulating within `data_len` (`data_len[n] > data_len - total` rejects before the add can overflow), so `total` can never exceed the real buffer length.
- **Item 2c** (`nns_edge_memdup()` return value ignored, leaving `data == NULL` with non-zero `data_len` on OOM): fixed — the return value is checked, the partially-filled handle is cleared, and `NNS_EDGE_ERROR_OUT_OF_MEMORY` is returned.
- **Item 2d** (`ed->num`/`ed->data[]` overwritten without releasing previous content → leak): fixed — `_nns_edge_data_clear_raw_data()` is called before repopulating the handle, and is reused correctly from `nns_edge_data_destroy()` too.
- Necessary side fix: rejecting `meta_len == 0` in `nns_edge_metadata_deserialize()` used to make a legitimate round-trip of a handle with no metadata fail; `nns_edge_data_deserialize()` now resets the metadata instead of calling deserialize on an empty buffer. This is required for the stricter size check to work without breaking the "no metadata" case, so it belongs in this PR.
- **Item 6** (uninitialised stack bytes serialised, ~2 KB info leak per publish): fixed — `edata_header` is `memset` to 0 before being filled in `nns_edge_data_serialize()`.

I confirmed the accumulation invariant (`total <= data_len` at every step, so the subtraction `data_len - total` never underflows) holds by hand-tracing `nns_edge_data_is_serialized()`, and that `num_mem > NNS_EDGE_DATA_LIMIT` is rejected *before* the `data_len[]` loop, so a hostile `num_mem` can't index past the fixed 256-entry array.

Both sub-items are fully addressed. Item 1 (`nns_edge_metadata_deserialize` OOB read), items 3–5, 7–10 from #259 are explicitly out of scope and left for follow-up PRs, which matches the PR description.

## Regression / blast radius check

- `nns_edge_data_deserialize()` is only called from the two MQTT message callbacks (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`), always on a freshly-created, empty `data_h`. So the new "clear existing content first" behavior is purely defensive here and changes nothing on the happy path.
- `nns_edge_data_is_serialized()` is a public API (declared in `include/nnstreamer-edge-data.h`) but the change only makes it *stricter* (rejects previously-exploitable malformed inputs); no legitimate caller could have depended on the old permissive/overflowing behavior.
- Diff touches exactly 3 files: `include/nnstreamer-edge-data.h` (+1 doc line), `src/libnnstreamer-edge/nnstreamer-edge-data.c` (+69/-16), and the test file. No other module is touched. The test file's large diff (662/-197) is dominated by a separate, mechanical `clang-format` commit — justified by `.github/workflows/static.check.yml`, which reformats and diffs the *whole* changed file, so any PR touching this pre-clang-format file would otherwise fail "Static checks" with an unrelated diff. Good call keeping it as its own commit.
- Confirmed on GitHub that PR CI (`build`, which runs `ctest` twice — plain and with `-fsanitize=address` — and is one of the two required checks per branch protection) is green, along with "Build and test on Ubuntu 26.04 (gcc 15)" and "Static checks". The ASan run in particular is the check that would catch a heap-buffer-overflow like the one item 2 describes, so this isn't just a claim in the PR description — it's independently verified passing.

## Size vs. topic

The functional fix itself is small and focused (69 added / 16 removed lines in `nnstreamer-edge-data.c`). Appropriate for the topic; no scope creep into unrelated modules.

## Test coverage

Comprehensive and well-targeted:
- Negative tests for every sub-bug: short buffer (`invalidParam04/05_n` on `edgeDataIsSerialized`/`edgeDataDeserialize`), overflowing `data_len[]` sum (`invalidParam06_n` on both), empty-memory rejection (`invalidParam07_n`), metadata-overrun (`invalidParam08_n`), under-declared size (`invalidParam09_n`).
- `edgeDataSerialize.headerIsCleared` dirties the stack before serializing and asserts every byte past `num_mem` is zero — directly proves item 6.
- `edgeDataDeserialize.replaceHandleContent` fills a handle with existing memories + metadata before deserializing into it, so LeakSanitizer (via the ASan CI job) would catch the leak this PR fixes (item 2d) if it regressed.
- `edgeDataSerialize.noMetadata` / `edgeDataDeserialize.copyDeserialized` cover the "no metadata" round-trip.
- A nice touch for **future regression detection**: `ne_test_data_header_s` mirrors the internal header layout, and `_get_serialized_data()` asserts `sizeof(ne_test_data_header_s)` matches the actual serialized size. If someone changes `nns_edge_data_header_s` (e.g. bumps `NNS_EDGE_DATA_LIMIT`) without updating the mirror, this assertion fails loudly instead of the tests silently testing the wrong offsets.
- These tests run under the required `build` CI check (both plain and ASan builds via `ctest`), so a future regression in this code path would fail CI and block merge — verified this is actually the case for this PR run, not just asserted in the description.

## Documentation / API

No signature or architecture change. The one new possible return value (`NNS_EDGE_ERROR_OUT_OF_MEMORY` from `nns_edge_data_deserialize()`) is documented with an added `@retval` line in `include/nnstreamer-edge-data.h` in the same PR. No further doc updates needed.

## Minor finding (not a merge blocker)

**Minor** — In `nns_edge_data_deserialize()`, the new "no metadata" branch:
```c
nns_edge_metadata_destroy (ed->metadata);
ed->metadata = NULL;
ret = nns_edge_metadata_create (&ed->metadata);
```
does not special-case a failure of `nns_edge_metadata_create()` (OOM only). On that rare failure, `ed->metadata` is left `NULL` and the error propagates, but the raw-data entries deserialized just before this point are *not* rolled back. This doesn't crash — every other `nns_edge_metadata_*` call already null-checks its handle and degrades to `NNS_EDGE_ERROR_INVALID_PARAMETER` — but it does leave the handle in a partially-populated, metadata-less state after an error return. Given this is an OOM-only path and the PR itself notes that `malloc()` failure paths are the remaining uncovered branches (94.1%/84.0% coverage), I'd call this a follow-up nit rather than something blocking this PR.

**Nit** — Already flagged by the PR author itself, restating for visibility: both MQTT callbacks still ignore `nns_edge_data_deserialize()`'s return value and raise `NNS_EDGE_EVENT_NEW_DATA_RECEIVED` with a possibly-empty handle when a payload is rejected. Correctly left out of this PR's scope (not a memory-safety issue), but worth a follow-up issue so a rejected/malformed payload doesn't silently surface as a "valid" empty data event to the application.

## Verdict

**No Critical or Major issues found.** Item 2 and item 6 of #259 are both fully and correctly addressed, the fix is tightly scoped to the affected file, the test suite thoroughly exercises every sub-bug (including a mechanism to catch future struct-layout drift), and CI (including the ASan build, a required check) is green. Recommend **merge**, optionally after opening a small follow-up for the two Minor/Nit points above.

myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Round 2 review

This is a **second review round**. It re-checks the whole PR from scratch (not just the delta) and specifically verifies that commit b2459d8, which was added in response to the "Minor" finding in the previous round's review comment on this PR, is correct.

Reviewed at `pr261` tip = `b2459d8` (`f167a8b`, `d377fe1`, `b2459d8` on top of the `5df56cd` clang-format commit; merge-base with `main` = `7980f8f`).

## Verification of the round-1 follow-up fix (b2459d8)

Round 1 flagged: in the "no metadata" branch of `nns_edge_data_deserialize()`, the old code did

```c
nns_edge_metadata_destroy (ed->metadata);
ed->metadata = NULL;
ret = nns_edge_metadata_create (&ed->metadata);
```

so an OOM in `nns_edge_metadata_create()` left `ed->metadata == NULL` -- a handle that is no longer safely usable by later `nns_edge_metadata_*` calls expecting a valid handle to still be around after a "failed" deserialize.

`b2459d8` changes this to create the replacement into a local `new_meta` first and only swap it into `ed->metadata` (destroying the old one) once creation succeeded:

```c
nns_edge_metadata_h new_meta = NULL;
ret = nns_edge_metadata_create (&new_meta);
if (NNS_EDGE_ERROR_NONE == ret) {
  nns_edge_metadata_destroy (ed->metadata);
  ed->metadata = new_meta;
}
```

This is correct: verified `nns_edge_metadata_create()` (`nnstreamer-edge-metadata.c:110`) leaves `*metadata_h` untouched (never partially assigned) on its only failure path (`calloc` returning NULL), so on failure here `ed->metadata` still points at the same, previously-valid handle, and the function still returns `NNS_EDGE_ERROR_OUT_OF_MEMORY` as before. **The round-1 finding is properly fixed.**

**Nit (residual, not introduced by this PR, not a blocker):** even after this fix, if `nns_edge_metadata_create()` fails here, the raw-data entries already deserialized earlier in the same call are *not* rolled back -- the handle ends up with new raw data but old/stale metadata, alongside the `OUT_OF_MEMORY` return. This is a pre-existing, OOM-only corner case (no fault-injection harness exists in this repo to test it, consistent with the PR's stated 94.1%/84.0% coverage and the untestable malloc-failure branches). Given the caller already gets a non-`NNS_EDGE_ERROR_NONE` return and both current callers (MQTT message callbacks) discard the handle on error paths in practice, this remains a nit, same severity judgement as round 1.

**Nit:** no test exercises the b2459d8 swap logic itself (would need malloc-failure injection for `nns_edge_metadata_create()`), so a future accidental revert of the create-then-swap ordering would not be caught by CI. Correctness was verified here by hand-tracing instead. Not a blocker given the same limitation applies to the rest of this file's OOM paths.

## Re-verification of the rest of the PR (Issue #259 items 2 and 6)

Re-derived independently rather than trusting round 1's summary:

- **Item 2a** (header fields read before length check): `nns_edge_data_is_serialized()` checks `data_len < sizeof (nns_edge_data_header_s)` before dereferencing `header->key`/`version`/`num_mem`. Confirmed correct.
- **Item 2b** (integer-overflow bypass of the size check): the loop `if (header->data_len[n] == 0U || header->data_len[n] > data_len - total) goto invalid_size; total += header->data_len[n];` maintains the invariant `total <= data_len` at every step (base case `total = sizeof(header) <= data_len`, and each addition is pre-bounded by `data_len - total`), so `total` can never wrap. The same invariant is applied to the `meta_len` check right after. Hand-traced and confirmed sound; also confirmed `num_mem > NNS_EDGE_DATA_LIMIT` (256) is rejected before the loop, so a hostile `num_mem` cannot index `data_len[]` out of the fixed-size array.
- **Item 2c** (`nns_edge_memdup()` return ignored, leaving `NULL` data with non-zero length on OOM): `nns_edge_data_deserialize()` now checks the return, clears what was already deserialized via `_nns_edge_data_clear_raw_data()`, and returns `NNS_EDGE_ERROR_OUT_OF_MEMORY`. `ed->num` is advanced one entry at a time (`ed->num = n + 1` after each successful `memdup`), so the cleanup on failure only frees what was actually allocated -- no double free, no leak.
- **Item 2d** (handle content overwritten without releasing previous content): `_nns_edge_data_clear_raw_data()` is called up front in `nns_edge_data_deserialize()`, and is correctly reused (not duplicated) in `nns_edge_data_destroy()`.
- **meta_len == 0 handling**: confirmed `nns_edge_data_add()` (`nnstreamer-edge-data.c:241`) rejects `data_len <= 0`, so the serializer can never emit a `data_len[n] == 0` entry -- the new `data_len[n] == 0U` rejection in `is_serialized()` cannot reject any legitimately-serialized buffer, only hostile ones. Also confirmed `nns_edge_metadata_deserialize()` (`nnstreamer-edge-metadata.c:347`) already frees the target handle's existing list (`nns_edge_metadata_free (meta)`) before repopulating it, so the `meta_len > 0` branch does not need the same create-then-swap treatment as the `meta_len == 0` branch -- no inconsistency between the two branches.
- **Item 6** (uninitialized stack bytes serialized): `edata_header` is `memset` to 0 before being filled in `nns_edge_data_serialize()`; the `headerIsCleared` test asserts every byte past `num_mem` is zero after dirtying the stack first -- a direct proof, not just a plausibility check.

## Regression / blast radius

- Confirmed (not just asserted) both call sites of `nns_edge_data_deserialize()` -- `nnstreamer-edge-mqtt-paho.c:85` and `nnstreamer-edge-mqtt-mosquitto.c:85` -- always call it on a `data_h` freshly created by `nns_edge_data_create()` a few lines above. So the added "clear existing content first" defensive behavior changes nothing on today's only call path; it only matters for future/direct API callers, or if a handle is ever reused across messages later.
- Diff footprint: `include/nnstreamer-edge-data.h` (+1 doc line), `src/libnnstreamer-edge/nnstreamer-edge-data.c` (round 1: +69/-16, round-2 follow-up: +7/-3), and the test file. No other module (MQTT transport, TCP/internal, metadata parsing, event handling) is touched. This matches items 2 and 6 of #259 exactly; item 1 (`nns_edge_metadata_deserialize()`'s own OOB-read bug, `||` loop condition) and items 3-10 remain correctly out of scope for this PR.
- `nns_edge_data_is_serialized()` is public API (`include/nnstreamer-edge-data.h`) but the change is strictly more restrictive (rejects previously-exploitable malformed input); no legitimate caller could depend on the old permissive/overflow-prone behavior.

## Size vs. topic

Still small and focused: the functional change across both rounds is under 80 net lines in a single source file plus its header doc comment. No scope creep.

## Tests / CI

- The new negative tests (`edgeDataIsSerialized.invalidParam04_n` through `invalidParam11_n`, `edgeDataDeserialize.invalidParam05_n`/`06_n`, `edgeDataSerialize.headerIsCleared`/`noMetadata`, `edgeDataDeserialize.replaceHandleContent`/`copyDeserialized`) were read in full; each targets exactly one sub-bug and would fail without the corresponding fix (e.g. `invalidParam06_n` constructs a `data_len[0] = 0xFFFFFFFFFFFFFF00` overflow attempt matching the old wraparound bug).
- `ne_test_data_header_s` mirrors the internal header layout and `_get_serialized_data()` asserts its `sizeof()` matches the real serialized size -- a real (if partial) guard against future struct-layout drift (e.g. bumping `NNS_EDGE_DATA_LIMIT`) silently breaking these tests' offset assumptions.
- Verified against the live PR (not just the description): for commit `b2459d8`, the two **required** status checks per branch protection (`DCO`, `build`, which runs `ctest` including the ASan pass) are both `completed`/`success`. `Static checks`/Tizen GBS builds/architecture cross-builds are still queued at review time but are not required checks on this repo (confirmed separately), so they do not gate merge either way.

## Documentation / API

No architecture change. The one new possible return value (`NNS_EDGE_ERROR_OUT_OF_MEMORY` from `nns_edge_data_deserialize()`) has its `@retval` line added in the same PR. No other doc changes needed; the repo has no separate architecture doc to update.

## Verdict

**No Critical or Major issues found in either round.** The round-1 Minor finding (metadata handle nulled out on OOM) is correctly and completely fixed by `b2459d8`. Items 2 and 6 of #259 are both fully addressed, the fix is tightly scoped, regression risk to other modules is effectively nil (single call path, both on fresh handles), and the required CI checks (DCO, build incl. ASan) are green on the latest commit. The two residual Nits above (no rollback of raw data on the metadata-OOM corner case; that corner case itself untested due to lack of fault injection) are pre-existing-severity, non-blocking, and consistent with round 1's judgement.

**Recommend: merge.**

myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Round 3 review

This is the **third review round**, re-checking the whole PR from scratch after commit `404ef4e`, which was added in response to round 2's Nit ("deserialize failure can leave raw memories in the handle while metadata is broken"). Reviewed at PR tip `404ef4e` (`5df56cd`, `f167a8b`, `d377fe1`, `b2459d8`, `404ef4e` on top of merge-base `7980f8f` with `main`).

## Verification that the round-2 Nit is fixed (404ef4e)

Round 2 flagged: when `nns_edge_metadata_deserialize()` (or the metadata-recreate path for a buffer with no metadata) fails, the just-deserialized raw memories were left in `ed->data[]`/`ed->num`, while the `nns_edge_memdup()` failure path already cleared them — the two error exits disagreed on what the caller was left holding.

`404ef4e` adds, right after the metadata step in `nns_edge_data_deserialize()`:

```c
if (NNS_EDGE_ERROR_NONE != ret)
  _nns_edge_data_clear_raw_data (ed);
```

Traced both failure branches:
- `header->meta_len > 0` and `nns_edge_metadata_deserialize()` fails: raw data is now cleared. Confirmed `nns_edge_metadata_deserialize()` itself already calls `nns_edge_metadata_free(meta)` before returning on error (`nnstreamer-edge-metadata.c:347` area), so `ed->metadata` is left as a valid-but-empty handle — no double free, no leak.
- `header->meta_len == 0` and the replacement `nns_edge_metadata_create()` fails (OOM, the b2459d8 case): raw data is now also cleared, consistent with the other branch.

Also added the matching unit test `edgeDataDeserialize.invalidParam07_n`: it appends a metadata blob with one key/value pair whose key is empty (`nns_edge_metadata_set()` rejects empty keys), sets `header->meta_len` to match, and asserts `nns_edge_data_get_count() == 0` after the failed deserialize. I hand-traced `nns_edge_metadata_deserialize()`'s internal loop against this exact byte layout (`total=1`, `cur` reaches exactly `data_len=7` after one iteration before returning) and confirmed the test does not itself trigger the known-and-intentionally-out-of-scope item-1 OOB bug (the `cur < data_len || ...` condition) — it fails cleanly on the empty-key check before the loop could re-enter. **The round-2 finding is properly and correctly fixed, with adequate test coverage.**

**Bonus, not claimed by the commit message**: this same one-line change also happens to close the *other* residual Nit that round 2 left open — "if `nns_edge_metadata_create()` fails, the raw-data entries already deserialized are not rolled back." That is no longer true; the unconditional clear at the end of the function now covers that path too. Worth calling out since neither the PR description nor the commit message mentions it, but it's a real (small) improvement beyond what the round-2 comment asked for.

## New finding: `@remarks` doc line overstates the guarantee in one corner case (Minor/Nit)

`404ef4e` adds to `include/nnstreamer-edge-data.h`:

```c
/**
 * @brief Deserialize entire edge data (meta data + raw data).
 * @remarks The previous content of @a data_h is released. The handle is left empty when the function fails.
```

This is not quite accurate for the `meta_len == 0` + `nns_edge_metadata_create()`-OOM path (the b2459d8 case): by design, `ed->metadata` is **not** released there — the whole point of b2459d8 was to keep the *previous* metadata handle intact rather than lose it on OOM. So on that one failure path, the handle ends up with raw data cleared (empty) but metadata **not** released and **not** empty — it still holds whatever key/value pairs were set on the handle before this `deserialize()` call. That contradicts both sentences of the new `@remarks` line for this one branch.

- Severity: **Minor/Nit** — this is an OOM-only, currently-untestable corner case (consistent with the PR's own note that `malloc()` failure paths are the remaining uncovered branches). It has no memory-safety impact (no leak/UAF; the old metadata handle is still owned and will be freed normally by `nns_edge_data_destroy()`), and both current callers (the MQTT message callbacks) always pass a *freshly created* handle whose metadata is already empty, so it doesn't manifest as an observable bug today. It only matters for a future/direct caller that reuses a handle with real metadata already on it and hits this exact OOM branch — which the added `@remarks` line seems to be documenting a contract for.
- Suggested fix (does not need to block this PR): soften the wording, e.g. "The previous raw data of @a data_h is released; on failure the handle's metadata may retain its previous content if replacing it also failed," or clear `ed->metadata` too on that specific failure path (at the cost of losing the previous metadata, which is exactly what b2459d8 was written to avoid — so wording the doc more precisely is the cheaper fix).

## Re-confirmation of items 2 and 6 of #259 (independently re-derived)

- **Item 2a** (header fields read before length check): `nns_edge_data_is_serialized()` checks `data_len < sizeof (nns_edge_data_header_s)` before touching `header->key`/`version`/`num_mem`. Confirmed, and confirmed `nns_edge_data_deserialize()` calls `nns_edge_data_is_serialized()` before touching `header` itself, so the deserializer is covered by the same gate.
- **Item 2b** (integer-overflow bypass of the size check): `total` is accumulated within `data_len` (`data_len[n] > data_len - total → invalid`), so `total` provably never exceeds `data_len`; `num_mem > NNS_EDGE_DATA_LIMIT` is rejected before the loop, so a hostile `num_mem` cannot index the fixed 256-entry array out of bounds. Hand-traced and sound.
- **Item 2c** (`nns_edge_memdup()` return ignored): now checked; on failure the partial handle is cleared and `NNS_EDGE_ERROR_OUT_OF_MEMORY` is returned. `ed->num` is advanced one entry at a time, so cleanup on failure frees exactly what was allocated.
- **Item 2d** (handle overwritten without releasing previous content): `_nns_edge_data_clear_raw_data()` is called up front, and correctly reused (not duplicated) from `nns_edge_data_destroy()`.
- **`meta_len == 0` handling**: confirmed `nns_edge_data_add()` rejects `data_len <= 0`, so the serializer can never emit `data_len[n] == 0`; the new `data_len[n] == 0U` rejection in `is_serialized()` therefore cannot reject any legitimately-serialized buffer.
- **Item 6** (uninitialised stack bytes serialised): `edata_header` is `memset` to 0 before being filled; `edgeDataSerialize.headerIsCleared` dirties the stack first and asserts every byte past `num_mem` is zero afterwards — a direct proof. Also noticed the accompanying `if (edata_header.meta_len > 0) memcpy (...)` guard added in the same commit avoids a `memcpy(ptr, NULL, 0)` call (technically UB, since `nns_edge_metadata_serialize()` returns `*data = NULL` when there is no metadata) — a small correctness improvement riding along with the item-6 fix.

Both sub-items remain fully addressed; item 1 (`nns_edge_metadata_deserialize()`'s own `||`-loop OOB read) and items 3–10 of #259 remain correctly out of scope.

## Regression / blast radius

- Diff footprint across all commits: `include/nnstreamer-edge-data.h` (+2 doc lines), `src/libnnstreamer-edge/nnstreamer-edge-data.c` (net +76/-16 across the three functional commits), and the test file (dominated by the standalone, justified clang-format commit). No other module (MQTT transport, TCP/internal, metadata parsing, event handling) is touched.
- Confirmed both call sites of `nns_edge_data_deserialize()` (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`) always pass a `data_h` created immediately above by `nns_edge_data_create()`. The new "clear existing content first" behavior is therefore purely defensive on today's only call path — no behavior change for existing users.
- `nns_edge_data_is_serialized()` is public API but the change only makes it stricter (rejects previously-exploitable malformed input); no legitimate caller could depend on the old permissive/overflow-prone behavior.

## Size vs. topic

Still small and tightly scoped: under 80 net functional lines in one source file plus its header doc comment, across all rounds combined. No scope creep.

## Tests / CI

- Test coverage is thorough and each new negative test targets exactly one sub-bug (verified by reading the full test diff, not just the PR's own table).
- `ne_test_data_header_s` mirrors the internal header layout and `_get_serialized_data()` asserts its `sizeof()` matches the real serialized size — a real guard against future struct-layout drift (e.g. bumping `NNS_EDGE_DATA_LIMIT`) silently breaking these tests' offset assumptions.
- **At the time of this review, CI on the latest commit (`404ef4e`) is still `pending`** for every check except `DCO` (`pass`), including the required `build` check (which runs `ctest` twice — plain and ASan — and is what would actually catch a regression here). Rounds 1 and 2 both explicitly verified `build` was green before recommending merge; this round could not do the same because the checks had not finished running yet. **This should be re-verified as green before merging** — it is very likely to pass given rounds 1–2 already verified the equivalent diff up to `b2459d8`, and `404ef4e` is a small, well-tested increment on top, but it has not been independently confirmed for this exact commit as of this review.

## Documentation / API

No architecture change. The one new possible return value (`NNS_EDGE_ERROR_OUT_OF_MEMORY` from `nns_edge_data_deserialize()`) is documented. See the Minor/Nit finding above regarding the precision of the new `@remarks` line for one OOM corner case.

## Verdict

**No Critical or Major issues found in any of the three rounds.** The round-2 Nit is correctly and completely fixed by `404ef4e`, with good test coverage, and incidentally also resolves the other residual Nit from round 2. One new **Minor/Nit** is raised in this round: the `@remarks` doc line added in `404ef4e` slightly overstates the "handle is left empty on failure" guarantee for the metadata-recreate-OOM corner case — no functional or memory-safety impact, non-blocking, fixable by wording alone.

**Recommend: merge**, once the required `build` (and `DCO`) checks report green on `404ef4e` (still pending at review time). Optionally tighten the `@remarks` wording in this PR or a fast follow-up.

myungjoo:
*Note: This is a follow-up from the same AI agent (Claude) round-3 review above, posted on behalf of @myungjoo.*

**CI status update:** all checks on `404ef4e` have now completed and are green, including the required `build` (ctest, plain + ASan) and `DCO`, plus `Static checks`, `Spell Check with Typos`, `Build and test on Ubuntu 26.04 (gcc 15)`, and all Tizen GBS / cross-arch (`aarch64`, `armv7l`, `x86_64`) builds. This resolves the one open caveat in the round-3 review above (CI was still pending at review time).

**Updated verdict: no remaining blockers. Recommend merge.** The single outstanding item is the Minor/Nit on the `@remarks` doc wording noted above, which is non-blocking.

myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Round 4 review

This is the **fourth review round**, re-checking the whole PR after commit `526c453`, which was added in response to round 3's Minor/Nit finding (the `@remarks` line on `nns_edge_data_deserialize()` overstated the guarantee for the `meta_len == 0` + OOM corner case). Round 3 reviewed at `404ef4e` and did not see this commit. Reviewed at PR tip `526c453` (`5df56cd`, `f167a8b`, `d377fe1`, `b2459d8`, `404ef4e`, `526c453` on top of merge-base `7980f8f` with `main`).

## Verification that the round-3 finding is fixed (526c453)

Round 3 flagged that this line

```c
* @remarks The previous content of @a data_h is released. The handle is left empty when the function fails.
```

was inaccurate for the `meta_len == 0` + `nns_edge_metadata_create()`-OOM path, where `ed->metadata` is deliberately **not** replaced/released (that's the whole point of `b2459d8`).

`526c453` narrows the claim to raw data only:

```c
* @remarks The previous raw data of @a data_h is released, and the handle holds no raw data when the function fails.
```

This correctly removes the false claim about metadata always being released/empty on failure — confirmed by re-reading `nns_edge_data_deserialize()`: on the `meta_len == 0` + OOM branch, `ed->metadata` is left pointing at whatever it held before the call, exactly as the new wording implies (it only talks about raw data now). **The round-3 finding, as stated, is fixed.**

## New finding: the narrowed `@remarks` still overstates the guarantee for one class of failure (Minor)

The new wording claims raw data is *always* released "when the function fails," for every non-`NNS_EDGE_ERROR_NONE` return. Tracing every return path of `nns_edge_data_deserialize()` (`src/libnnstreamer-edge/nnstreamer-edge-data.c:604-676`):

```c
ed = (nns_edge_data_s *) data_h;
if (!ed || !data) { ... return NNS_EDGE_ERROR_INVALID_PARAMETER; }        /* (a) */
if (!nns_edge_handle_is_valid (ed)) { ... return NNS_EDGE_ERROR_INVALID_PARAMETER; }  /* (b) */
ret = nns_edge_data_is_serialized (data, data_len);
if (ret != NNS_EDGE_ERROR_NONE) return ret;                               /* (c) */

nns_edge_lock (ed);
_nns_edge_data_clear_raw_data (ed);        /* <-- only reached past (a)-(c) */
...
```

`_nns_edge_data_clear_raw_data(ed)` is only reached once `nns_edge_data_is_serialized()` has already validated the buffer. Failure path (c) — the buffer is too short, has a bad key/version, `num_mem` too large, or a size-check mismatch (i.e. a malformed/hostile input, exactly the threat model item 2 addresses) — returns *before* any clearing happens, so a handle that already held real raw data (added via `nns_edge_data_add()`) keeps that raw data untouched. The function fails, but the handle does **not** hold "no raw data" as the `@remarks` now claims; it holds whatever it had before the call.

This is easy to demonstrate: create a handle, `nns_edge_data_add()` a real memory into it, then call `nns_edge_data_deserialize()` with a 1-byte buffer (fails at the `data_len < sizeof(header)` check in `nns_edge_data_is_serialized()`). The call returns `NNS_EDGE_ERROR_INVALID_PARAMETER`, and `nns_edge_data_get_count()` on the handle still reports the pre-existing memory — not empty.

- **Severity: Minor** — documentation-only, no memory-safety or functional impact. `_nns_edge_data_clear_raw_data()` itself is correct and doesn't double-free/leak on any path; this is purely about the `@remarks` line asserting a stronger guarantee than the code provides.
- **Why it still matters enough to note**: both current callers (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`) always pass a handle fresh off `nns_edge_data_create()`, so there is no observable bug today — this was already established in rounds 1-3 and still holds. But the doc line is a general API contract, and a future/direct caller that reuses a handle across calls (the scenario the whole "clear previous content" fix in this PR was written to make safe) could reasonably rely on "handle holds no raw data when the function fails" to skip an explicit cleanup after a rejected/malformed payload — which would be wrong specifically for the most common failure mode (a malformed input buffer), while being right for the OOM-during-deserialize failure modes.
- No test in this PR exercises this distinction either way (`invalidParam0X_n` tests for `nns_edge_data_deserialize` all pass handles with no pre-existing raw memory, so none of them would have caught the doc/behavior mismatch).
- **Suggested fix (non-blocking):** either (1) reword once more to scope the guarantee correctly, e.g. "Once the input buffer passes basic format/size validation, the previous raw data of @a data_h is released before new data is read, and no raw data is left if a later step then fails; if the input buffer itself is rejected as invalid, the handle's existing raw data is left untouched." or (2) make the code match the simpler doc by moving `_nns_edge_data_clear_raw_data(ed)` earlier (e.g., right after the `is_serialized()` check succeeds is already "before touching header", but to cover path (c) too it would have to run even when the buffer is rejected — probably not worth doing since it would mean a syntactically invalid buffer wipes valid existing data, which seems like the wrong tradeoff). Wording is the cheaper and safer fix, consistent with how round 3's finding was resolved.

## Re-confirmation of items 2 and 6 of #259

No functional code changed since round 3 (`404ef4e`); `526c453` is a one-line doc diff only (`git diff 404ef4e..526c453` touches only `include/nnstreamer-edge-data.h`). Rounds 2 and 3 already independently re-derived and hand-traced the full fix for both items from scratch; re-confirmed the diff is unchanged:

- Item 2a (header read before length check), 2b (overflow-safe `total` accumulation bounded by `data_len - total`), 2c (`nns_edge_memdup()` failure now checked and cleaned up), 2d (previous raw data released via `_nns_edge_data_clear_raw_data()` before repopulating) — all still correctly in place, byte-for-byte identical to what rounds 2-3 verified.
- Item 6 (`edata_header` now `memset` to 0 before being filled in `nns_edge_data_serialize()`, proven by `edgeDataSerialize.headerIsCleared`) — unchanged, still correct.

Both sub-items remain fully addressed. Item 1 and items 3-10 of #259 remain correctly out of scope.

## Regression / blast radius

- Confirmed globally (not just in the two known files) that `nns_edge_data_deserialize()` has exactly two callers in the whole tree (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`), both always passing a handle freshly created a few lines above. `nns_edge_data_is_serialized()` has no callers outside this file and the test suite. `nns_edge_data_deserialize_meta()` (item 1, a different function) is untouched by this PR. No other module reads or depends on the changed code paths.
- Diff footprint across all six commits: `include/nnstreamer-edge-data.h` (+4 doc/retval lines total), `src/libnnstreamer-edge/nnstreamer-edge-data.c` (net +76/-16 across the functional commits), and the test file (dominated by the standalone, justified clang-format commit). Still no other module touched.

## Size vs. topic

Unchanged from round 3's assessment: tightly scoped, under 80 net functional lines in one source file plus its header doc comment. `526c453` itself is a 2-line diff (1 removed, 1 added). No scope creep.

## Tests / CI

- All **11** CI checks are green on `526c453` (confirmed live against the PR, not just the description): `DCO`, `build` (required, runs `ctest` plain + ASan), `Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, `Spell Check with Typos`, and all six Tizen GBS / cross-arch builds (`aarch64`, `armv7l`, `x86_64` × unit-test and no-unit-test variants). Round 3's one open caveat (CI still pending) is resolved.
- `526c453` is doc-only and doesn't need a new test; no regression risk from it. The test suite added in earlier commits (`invalidParamNN_n` series, `headerIsCleared`, `replaceHandleContent`, `noMetadata`/`copyDeserialized`, the metadata-failure `invalidParam07_n`) remains unchanged and, as verified in rounds 1-3, is well targeted at each sub-bug and runs under the required, ASan-enabled `build` check — a future regression here would fail CI and block merge.

## Documentation / API

No architecture change. `NNS_EDGE_ERROR_OUT_OF_MEMORY` is documented as a new `@retval`. The `@remarks` line is the subject of this round's one new finding above — accurate for the metadata question round 3 raised, still slightly overbroad for one class of failure (malformed input rejected before any clearing happens).

## Verdict

**No Critical or Major issues found in any of the four rounds.** The round-3 finding is correctly fixed by `526c453` for the case it was raised about (metadata handling on the OOM path). This round raises one new **Minor** finding: the revised `@remarks` line still overstates the "no raw data on failure" guarantee for inputs that `nns_edge_data_is_serialized()` rejects before any clearing occurs — no functional or memory-safety impact given both current call sites always use freshly created handles, fixable by wording alone, and not a merge blocker.

Items 2 and 6 of #259 remain fully and correctly addressed, the change stays tightly scoped to `nnstreamer-edge-data.c`/`.h` with no regression risk to other modules (single pair of call sites, both on fresh handles), and all 11 CI checks (including the required ASan `build`) are green on the current head `526c453`.

**Recommend: merge.** Optionally fold one more wording pass on the `@remarks` line into this PR or a fast follow-up; it is not worth another review round by itself.

myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Round 5 review

This is the **fifth review round**, re-checking the whole PR after commit `c7b5be2`, which was added in response to round 4's Minor finding (the `@remarks` line on `nns_edge_data_deserialize()` still claimed raw data is *always* released "when the function fails," even though a buffer rejected by `nns_edge_data_is_serialized()` leaves a handle's pre-existing raw data untouched). Reviewed at PR tip `c7b5be2` (`5df56cd`, `f167a8b`, `d377fe1`, `b2459d8`, `404ef4e`, `526c453`, `c7b5be2` on top of merge-base `7980f8f` with `main`).

## Verification that the round-4 finding is fixed (c7b5be2), matched against every return path

`c7b5be2` changes the `@remarks` of `nns_edge_data_deserialize()` in `include/nnstreamer-edge-data.h` to:

```
* @remarks The handle is left untouched if @a data is not a valid serialized edge data.
* Otherwise the previous raw data is released, and the handle holds no raw data when the deserialization fails.
```

Per this round's instruction, every `return`/exit of `nns_edge_data_deserialize()` (`src/libnnstreamer-edge/nnstreamer-edge-data.c:616-679` at this commit) was traced against this wording, rather than re-litigating phrasing:

1. `!ed || !data` (`:626-629`) returns before touching the handle. `data == NULL` cannot be a valid serialized buffer, so this is covered by sentence 1; `ed == NULL` has no handle to touch either way.
2. `!nns_edge_handle_is_valid(ed)` (`:631-634`) also returns before touching the handle. Not literally about `data`'s validity, but the outcome ("left untouched") still holds, and sentence 1 is phrased as a sufficient condition, not a biconditional, so this is not a contradiction.
3. `nns_edge_data_is_serialized(data, data_len) != NNS_EDGE_ERROR_NONE` (`:636-638`) returns before `nns_edge_lock`/`_nns_edge_data_clear_raw_data` run. This is exactly the case sentence 1 describes (`data` is not a valid serialized edge data): confirmed the handle's pre-existing raw data (e.g. from a prior `nns_edge_data_add()`) is left in place. Matches.
4. Past that check, `_nns_edge_data_clear_raw_data(ed)` runs unconditionally (`:641`) before any repopulation, so "the previous raw data is released" (sentence 2, first half) holds for every path from here on, success or failure.
5. `nns_edge_memdup()` failure inside the loop (`:647-653`) does an explicit `_nns_edge_data_clear_raw_data(ed)` then `return NNS_EDGE_ERROR_OUT_OF_MEMORY`. `ed->num` is advanced one entry at a time (`:657`), so the clear only frees what was actually allocated (checked `_nns_edge_data_clear_raw_data()` at `:76-90`: it iterates `i < ed->num` and calls each entry's `destroy_cb`, so this is safe, no double free/UAF). Matches "holds no raw data when the deserialization fails".
6. A metadata-step failure, either the `meta_len > 0` branch (`nns_edge_metadata_deserialize` fails) or the `meta_len == 0` branch (replacement `nns_edge_metadata_create()` OOM, the `b2459d8` case), is caught by the unconditional `if (NNS_EDGE_ERROR_NONE != ret) _nns_edge_data_clear_raw_data (ed);` at `:674-675`. Matches, and correctly says nothing about metadata, which is deliberately *not* released on the `meta_len == 0` OOM sub-case (that asymmetry was round 3's finding and is not reintroduced here).
7. Success is not covered by the remark, and doesn't need to be.

**No case was found where the current `@remarks` text asserts something the code does not do.** The round-4 finding is correctly and completely fixed; unlike rounds 1-4, this round finds no further inaccuracy in this doc line, only (in case 2 above) an incompleteness that is not a false statement and, per this round's guidance, is not worth another wording pass.

## Re-confirmation of items 2 and 6 of #259

No functional code changed since round 3 (`404ef4e`); `c7b5be2`, like `526c453`, is a header-doc-only diff (`git diff 404ef4e..c7b5be2` touches only `include/nnstreamer-edge-data.h`, +3/-1 across both doc commits). Re-verified directly against the current source rather than trusting prior rounds' summaries:

- **Item 2** — `nns_edge_data_is_serialized()` (`nnstreamer-edge-data.c:685`) checks `data_len < sizeof (nns_edge_data_header_s)` before touching `header->key`/`version`/`num_mem`; `num_mem > NNS_EDGE_DATA_LIMIT` is rejected before the `data_len[]` loop; the loop accumulates `total` bounded by `data_len - total` (`:726-729`), so `total` can never exceed `data_len` and the prior wraparound is impossible; `data_len[n] == 0U` is rejected. `nns_edge_data_deserialize()` checks `nns_edge_memdup()`'s result and clears the handle plus returns `NNS_EDGE_ERROR_OUT_OF_MEMORY` on failure; the previous content of the handle is released up front via `_nns_edge_data_clear_raw_data()`. All still correct.
- **Item 6** — `nns_edge_data_serialize()` (`:542-564`) does `memset (&edata_header, 0, header_len)` before filling any field, and only `memcpy`s `meta_serialized` when `edata_header.meta_len > 0` (avoiding a `memcpy(ptr, NULL, 0)`). Still correct.

Both sub-items remain fully addressed. Item 1 and items 3-10 of #259 remain correctly out of scope for this PR.

## Regression / blast radius, size vs. topic

Unchanged from rounds 1-4: `nns_edge_data_deserialize()` has exactly two callers in the tree (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`), both always on a handle freshly created a few lines above, so the stricter/defensive behavior in this PR changes nothing on today's only call path. `nns_edge_data_is_serialized()` is public API but only made stricter. Diff still touches exactly 3 files (`include/nnstreamer-edge-data.h`, `src/libnnstreamer-edge/nnstreamer-edge-data.c`, the test file), no other module. `c7b5be2` itself is a 3-line net diff in the header only, no code or test changes, so no new regression surface.

## Tests / CI

`c7b5be2` is doc-only and needs no new test, consistent with how `526c453` was treated in round 4. The test suite from earlier commits is unchanged and, per rounds 1-4, is well targeted at each sub-bug and runs under the required, ASan-enabled `build` check.

**CI status at the time of this review: all checks on `c7b5be2` are still `pending`** (`Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, `Spell Check with Typos`, `build` (required), `build (-A aarch64/armv7l/x86_64)`, and the Tizen GBS builds), except `DCO`, which is `pass`. Given `c7b5be2` only touches a comment this is expected to pass, and rounds 1-4 already verified the equivalent functional diff up to `526c453`, but as with round 3's caveat this has not been independently confirmed green for this exact commit yet. Re-verify the required `build` check is green on `c7b5be2` before merging.

## Documentation / API

No architecture change. The `@remarks` wording issue raised across rounds 3-4 is, per the trace above, now consistent with every return path of `nns_edge_data_deserialize()`. No further doc changes needed.

## Verdict

**No Critical or Major issues found in any of the five rounds.** The round-4 finding is correctly and completely fixed by `c7b5be2`, the `@remarks` line now matches the code's actual behavior on every return path, verified by tracing each one rather than by wording preference. Items 2 and 6 of #259 remain fully and correctly addressed, the change stays tightly scoped to `nnstreamer-edge-data.c`/`.h` with no regression risk to other modules, and this round finds no new issues.

**Recommend: merge**, once the required `build` check (currently pending) reports green on `c7b5be2`.

myungjoo:
Thanks for the approval. All three open comments are addressed in `a9d46c5` and `f2df80e` rather than left as follow-ups.

**1. `nns_edge_data_clear()` duplicated the helper** — fixed, it now calls `_nns_edge_data_clear_raw_data()`. You were right that this is the one that matters: the helper is what the new failure paths depend on.

**2. The mirrored struct only had its size pinned** — rather than move the private struct into a shared header, `_get_serialized_data()` now reads `version`, `num_mem`, `data_len[0]` and `meta_len` back *through the mirror* and asserts what the serializer wrote, which pins every field position. Verified by swapping `meta_len` and `data_len[]` in the mirror (same total size) and rebuilding: it now fails inside `_get_serialized_data()` at the new assertions instead of `headerIsCleared` silently checking the wrong offsets.

This keeps the private type private; if you would still prefer the header move, say so and I will do it.

**3. `headerIsCleared` readability and the best-effort stack** — the byte-by-byte loop is now a single `memcmp` against a zero block, so a regression prints one failure rather than thousands. The `@note` on `_dirty_stack()` now states that it is best effort and can only cause a vacuous pass, never a spurious failure, and a comment at the call site says the `memset` in the serializer is what actually makes the test pass. Re-verified that removing the `memset` still fails the test in the `memcmp` form.

**4. Rebase** — already done; the branch sits on `aaa2733` and the `[CodeClean]` clang-format commit is gone now that #260 is merged.

Local ASan run on this head is 2/2 green, and GNU indent is clean on `nnstreamer-edge-data.c`.

myungjoo:
*Note: This review was performed by an AI agent (Claude) on behalf of @myungjoo and is posted here as-is (transcribed from the agent's findings), not written directly by a human reviewer.*

## Round 6 review

This is an independent re-review of the whole PR, focused specifically on whether the two newest commits — `a9d46c5` ([CodeClean] reuse the raw data helper in `nns_edge_data_clear`) and `f2df80e` ([Test] pin the serialized header mirror to the real layout) — correctly close the three non-blocking comments left by the last `myungjoo-bot` review round (`c7b5be2`), and whether they introduce any regression. Reviewed at PR tip `f2df80e` (base `aaa2733`; diff still touches exactly 3 files: `include/nnstreamer-edge-data.h`, `src/libnnstreamer-edge/nnstreamer-edge-data.c`, `tests/unittest_nnstreamer-edge.cc`).

## Verification of the 3 outstanding comments

**1. `nns_edge_data_clear()` duplicated `_nns_edge_data_clear_raw_data()` — fixed correctly, verified byte-for-byte.**

Old body of `nns_edge_data_clear()`:
```c
for (i = 0; i < ed->num; i++) {
  if (ed->data[i].destroy_cb)
    ed->data[i].destroy_cb (ed->data[i].data);
  ed->data[i].data = NULL;
  ed->data[i].data_len = 0;
  ed->data[i].destroy_cb = NULL;
}
ed->num = 0;
```
`_nns_edge_data_clear_raw_data()` (the helper already used by `destroy`/`deserialize`):
```c
for (i = 0; i < ed->num; i++) {
  if (ed->data[i].destroy_cb)
    ed->data[i].destroy_cb (ed->data[i].data);
  ed->data[i].data = NULL;
  ed->data[i].data_len = 0U;
  ed->data[i].destroy_cb = NULL;
}
ed->num = 0U;
```
These are semantically identical (0 vs 0U is the same value; `nns_edge_lock`/`nns_edge_unlock` around the call site are unchanged, so the "call under the handle lock" precondition documented on the helper is still met). `nns_edge_data_clear()` now simply calls the helper between the existing lock/unlock pair, and the now-dead `unsigned int i` local is removed. Confirmed: **no behavioral change**, purely a duplication removal. This closes the risk the reviewers flagged (the two copies drifting apart while the new deserialize failure paths depend on the helper).

**2. Mirrored test header only pinned by size, not by field position — fixed, and more thoroughly than the original suggestion.**

`_get_serialized_data()` now reads back `header->version` (validated with `nns_edge_parse_version_key()`, which checks a fixed 16-bit tag before accepting), `header->num_mem == 1`, `header->data_len[0] == mem_len`, and `header->meta_len == 0` — i.e. it asserts the *values* the serializer wrote land where the mirror struct `ne_test_data_header_s` expects them, not just the mirror's total `sizeof()`. Cross-checked the real struct in `src/libnnstreamer-edge/nnstreamer-edge-data.c:23-30` against the mirror in the test file: field order/types (`uint32_t key; uint64_t version; uint32_t num_mem; nns_size_t data_len[NNS_EDGE_DATA_LIMIT]; nns_size_t meta_len;`) match exactly. Traced through a hypothetical same-size reorder (e.g. swapping `data_len[]` and `meta_len`, or `key` and `num_mem`) and confirmed the new assertions in `_get_serialized_data()` would catch it: a swapped `meta_len`/`data_len[]` pair makes `header->data_len[0]` read the real `meta_len` bytes (0) instead of `mem_len` (64) and fails; a swapped `key`/`num_mem` pair shifts `version`'s apparent position so `nns_edge_parse_version_key()` no longer sees the `0xefdd...` tag and fails. This is a real, not cosmetic, improvement over "only the size is pinned," and avoids exposing the private struct in a shared header, which was a reasonable design trade-off.

**3. `headerIsCleared` readability + `_dirty_stack()` honesty — fixed.**

The ~2 KB byte-by-byte `EXPECT_EQ` loop is now one `EXPECT_EQ (0, memcmp (...))` against a zero-filled block (freed via `SAFE_FREE` after the compare) — a real regression now prints one failure line instead of thousands, with no change in what is actually checked (same offset range, same "must be zero" assertion). `_dirty_stack()` now carries an explicit `@note` stating it is best-effort and can only produce a vacuous pass, never a spurious failure, and the call site notes that the `memset` in the serializer — not the dirty-stack trick — is what makes the test meaningful. This is documentation/readability only; the test's actual pass/fail behavior is unchanged.

## Regression check (explicit focus per review request)

- `nns_edge_data_clear()` is used by callers expecting it to release every raw-memory entry (calling each `destroy_cb`) and reset `ed->num` to 0 while leaving metadata and the handle itself intact. That is exactly what `_nns_edge_data_clear_raw_data()` does — confirmed by direct comparison above — so this refactor is behavior-preserving for every existing caller.
- No other module was touched: grepped the tree and confirmed `nns_edge_data_deserialize()` still has exactly two callers (`nnstreamer-edge-mqtt-paho.c:85`, `nnstreamer-edge-mqtt-mosquitto.c:85`), both on a handle freshly created immediately above, so none of this PR's "clear before repopulate" behavior is observable on any existing call path.
- The `include/nnstreamer-edge-data.h` `@remarks`/`@retval` text is unchanged by these two commits and still matches every return path of `nns_edge_data_deserialize()` traced in the previous round (untouched buffer on `is_serialized()` rejection; raw data cleared and no raw data left on every failure past that point; metadata intentionally left alone only on the `meta_len == 0` + OOM sub-path). No new doc/code mismatch introduced.
- CI: all 11 checks green on `f2df80e` (confirmed live via `gh pr checks 261`), including the required `build` check (ctest, plain + ASan) and `DCO`. `gh pr view` reports `mergeable: MERGEABLE`; the `BEHIND` status is only because unrelated PRs have advanced `main` past the `aaa2733` merge-base since this branch was last rebased — a routine "Update branch" away from a green merge button, not a defect in this PR's content.

## Size vs. topic / tests / docs

Both new commits are minimal and single-purpose: `a9d46c5` is a 7-line net removal in one function (no test change needed, since it's a pure refactor covered by the same existing tests that already exercise `nns_edge_data_clear()`'s callers indirectly through `destroy`/`deserialize` paths and directly through the clear-specific tests). `f2df80e` only touches the test file and only strengthens assertions. No scope creep, no other module touched, no new API/architecture surface, so no further documentation is required beyond what earlier rounds already added to `include/nnstreamer-edge-data.h`.

## Verdict

**No Critical, Major, or Minor issues found.** All three non-blocking comments from the `myungjoo-bot` APPROVED review are correctly and completely resolved by `a9d46c5`/`f2df80e`, with the `nns_edge_data_clear()` refactor verified byte-for-byte equivalent to the pre-existing behavior and no regression to any other module. Items 2 and 6 of #259 remain fully addressed (unchanged by these two commits), the fix stays tightly scoped to `nnstreamer-edge-data.{c,h}` plus its test file, and all 11 CI checks are green on the current head.

**Recommend: merge**, after a routine branch update onto the current `main` tip (no content changes needed).

myungjoo:
Rebased onto `main` @ `f25909e` (13 commits, the other #259 items). Textually clean; all 11 checks are green on `c185581`. The force-push dismissed the earlier approval, so this needs a fresh one.

One interaction worth checking on re-review, since `9b6693d [Fix] do not keep a freed metadata in the edge data handle` touches the same file:

* `nns_edge_data_clear_info()` can now leave `ed->metadata == NULL` when the re-creation fails. `nns_edge_data_deserialize()` copes with that on both branches: with `meta_len > 0` it passes the NULL to `nns_edge_metadata_deserialize()`, which rejects it and the raw data is then cleared; with `meta_len == 0` it creates the replacement first and assigns it, so it recovers the handle instead of tripping over it.
* The two functions now reset the metadata by different idioms — `clear_info()` destroys, clears the field, then creates; `deserialize()` creates first and swaps. Both are dangling-free. The swap keeps the previous metadata when the allocation fails, which is what an earlier review round asked for here, so I left it rather than churn the path. Happy to unify them on `clear_info()`'s idiom if you would rather have one shape in the file.

PR #262, opened 2026-09-07
[Fix] validate the strings a peer sends before parsing them

myungjoo:
Fixes items 3 and 4 of #259.

## The bugs

The `HOST_INFO` and `CAPABILITY` commands each carry one string whose length the peer chooses, and nothing checked that the bytes that arrived actually end with a null byte. Three reads went past the received buffer, none of them behind authentication:

- **query server, `HOST_INFO`.** `_nns_edge_accept_socket()` passes `cmd.mem[0]` straight to `nns_edge_parse_host_string()`. A string that is not terminated runs `strchr()` and then `strtoll()` off the end of the allocation; a command announcing `num == 0` passes a null pointer to the same function and crashes the listener thread. A terminated string without `':'` left `dest_host` null and `dest_port` — declared uninitialised — holding whatever was on the stack, and the server then connected to `inet_addr (NULL)`.
- **client, `CAPABILITY`.** The raw buffer reaches the application as the event payload, and `nns_edge_event_parse_capability()` calls `nns_edge_strdup()` on it, so a server that omits the terminator makes the client's `strlen()` run off the end.
- **hybrid client, MQTT.** The broker payload is a copy of exactly `payloadlen` bytes with no terminator of its own, and it goes to `nns_edge_parse_host_string()` unchecked. Anyone who can publish to the topic reaches this.

## The fix

The string is validated where the command enters, in `_nns_edge_cmd_receive()`: `HOST_INFO` and `CAPABILITY` must carry exactly one memory and its last byte must be null, otherwise the command is refused and the connection dropped before anything parses it. Our own sender writes `strlen () + 1` bytes and sets `num = 1`, and has done so since the protocol was introduced (`5989cfd` for both commands, `d353deb` for the hybrid publish), so no released peer is affected.

`nns_edge_event_parse_capability()` does not rely on that check to be safe: it has the event's data length at hand, so it copies with `nns_edge_strndup()` and stays inside whatever buffer the producer passed in. Item 4 stays fixed for any other producer of the event and if the receive path is ever restructured.

`nns_edge_parse_host_string()` no longer trusts its input: a null or empty string returns without touching anything, and the output params are always set, so a string without `':'` now leaves the host null and the port zero rather than leaving both as the caller declared them. `_nns_edge_accept_socket()` initialises its port and requires a non-empty host and an in-range port before connecting back — `":1234"` parses to an *empty* host, and a port beyond 65535 is silently truncated by `htons()`. The hybrid path checks the terminator of the payload and applies the same two conditions, and keeps waiting for a usable message instead of connecting to a host it could not parse.

No API or ABI change. `nns_edge_parse_host_string()` is internal (`nnstreamer-edge-util.h`, "DO NOT export this file") and has no callers outside the two fixed here.

## Tests

Most tests drive the library through a socket, because that is where the untrusted string comes from and the only place a missing terminator can be reproduced. Against the unfixed source, under AddressSanitizer, these reproduce exactly the failures reported in #259:

| test | unfixed result |
| --- | --- |
| `hostInfoNotTerminated_n` | heap-buffer-overflow, `READ of size 5` in `strtoll` ← `nns_edge_parse_host_string` util.c:135 |
| `hostInfoNoMemory_n` | `SEGV on unknown address 0x000000000000` in `strchr` ← util.c:131 |
| `hostInfoNoPort_n` | `SEGV on unknown address 0x000000000000` in `inet_addr` ← `_fill_socket_addr` internal.c:185 |
| `capabilityNotTerminated_n` | heap-buffer-overflow, `READ of size 12` in `strlen` ← `nns_edge_event_parse_capability` event.c:281 |
| `edgeEvent.parseCapabilityNotTerminated` | heap-buffer-overflow, `READ of size 16` in `strlen` ← `nns_edge_strdup` util.c:221 |

`hostInfoEmptyHost_n` and `hostInfoInvalidPort_n` are the two that do not crash the unfixed server — an empty host and a truncated port only made the connect-back fail a moment later — so they assert the refusal rather than reproduce a fault.

The ASan run lives in the `build` job, which is a required check, so a regression blocks the merge rather than only turning a job red.

`capabilityTerminated` is the positive counterpart: the same exchange with the terminator present must deliver the capability event and succeed, so the new check cannot be satisfied by refusing everything. Both capability tests assert the command the client writes back, which pins the wire value of `HOST_INFO`; without it, a new enumerator inserted before it would let the host info tests pass without ever reaching the code they cover. `edgeUtil.parseHostString` and four `_n` cases cover the parser directly, including the null and empty string that used to crash it.

The MQTT hybrid path has no new test: `unittest_nnstreamer-edge-mqtt` is deliberately not registered with `ADD_TEST()` because it needs a running broker, so a test there would not run in the required check. Its change is the same guard as the TCP path, over a payload whose terminator the publisher has always written.

## Note on the first commit

`tests/unittest_nnstreamer-edge.cc` predates the static checkers and does not match the repository's `.clang-format`, so any PR touching it fails static.check before its own changes are looked at. The first commit here is the same formatting commit as in #260, cherry-picked unchanged: whichever of the two lands first, the other drops it on rebase.


myungjoo:
> Note: this review was produced by another agent and is transcribed here as-is.

## Summary

LGTM. The change does what #259 items 3 and 4 ask for, the scope is tight (two `.c` files, one header comment, one test file), there is no API/ABI change, and I could not find a way for it to break an existing peer. Everything below is a non-blocking suggestion.

## What I verified

I built the PR tree in a local Linux environment with the same AddressSanitizer flags the `build` job uses, and I also built a second tree with the PR's tests kept but `nnstreamer-edge-internal.c` / `nnstreamer-edge-util.c` reverted to the base commit, to check that the new tests actually catch the bugs rather than only documenting them.

**Against the unfixed sources**, the four negative tests fail exactly as the PR body claims:

| test | ASan report |
| --- | --- |
| `hostInfoNotTerminated_n` | `heap-buffer-overflow` in `strtoll` ← `nns_edge_parse_host_string` util.c:135 ← `_nns_edge_accept_socket` internal.c:1123 |
| `hostInfoNoMemory_n` | `SEGV on 0x0` in `strchr` ← util.c:131 |
| `hostInfoNoPort_n` | `SEGV on 0x0` in `inet_addr` ← `_fill_socket_addr` internal.c:185 ← `_nns_edge_connect_to` |
| `capabilityNotTerminated_n` | `heap-buffer-overflow` in `strlen` ← `nns_edge_strdup` ← `nns_edge_event_parse_capability` event.c:281 |

`capabilityTerminated` passes against the unfixed tree too, which is the point: it is the positive counterpart and it also passes with the fix, so the new check is not satisfied by rejecting everything.

**Against the PR tree**, all 15 new/related tests pass and the whole `ctest` suite (`unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom`) passes clean under `ASAN_OPTIONS=detect_leaks=1`.

Other things I checked:

* **Blast radius.** `_nns_edge_cmd_string_is_valid()` returns `true` immediately for every command other than `HOST_INFO` / `CAPABILITY`, so `TRANSFER_DATA` and the whole `_nns_edge_message_handler()` loop are untouched. Both string commands are only exchanged during the handshake.
* **Interop with older peers.** `git log -L` on both send sites shows `cmd.info.num = 1; cmd.info.mem_size[0] = strlen (...) + 1;` unchanged since the initial import (`5989cfd`), so no peer ever built from this repository sends a `HOST_INFO` / `CAPABILITY` that the new check would reject. A repository-wide grep confirms `nns_edge_parse_host_string()` has no callers besides the two fixed here, and it is declared in `nnstreamer-edge-util.h` ("DO NOT export"), so the contract change is internal. There is no protocol document in the tree describing these commands, and no public header changed, so I agree no documentation update is due.
* **The hybrid `continue` is not a busy loop.** `nns_edge_mqtt_get_message (..., 0U)` reaches `nns_edge_queue_wait_pop()` with `timeout == 0`, which is `pthread_cond_wait()`; a wake-up with an empty queue makes `_pop_data()` fail, so `get_message` returns an error and the loop `break`s. `msg` and `server_ip` are per-iteration and freed before each `continue`, so no leak. On a bad payload the loop keeps waiting, which is what the old code did too (it only `break`s on a successful connect).
* **CI gating.** The two "not terminated" regressions are ASan-only — a plain build would happily read past the allocation and the tests would pass — so it does matter that the ASan `ctest` lives in the `build` job rather than a separate one. `build` is a required check and is already green here.
* **The formatting commit is provably non-semantic.** With all whitespace stripped, the file before and after `db70986` is byte-identical apart from the `#include` block, and the include *set* is unchanged (only reordered). `clang-format` 16 on the final tree produces an empty diff, and GNU `indent` with the repository's `indent.sh` options produces an empty diff on both changed `.c` files.

## Minor points

**1. `nns_edge_event_parse_capability()` itself is still unbounded.** (low, defense in depth)

`nnstreamer-edge-event.c:281` does `*capability = nns_edge_strdup (ee->data.data);` while `ee->data.data_len` is sitting right next to it. The fix is entirely at the transport layer, so the *public API* named in #259 item 4 remains unsafe for any other producer of `NNS_EDGE_EVENT_CAPABILITY` — a custom-connection module raising the event with a non-terminated buffer, or a future transport. `nns_edge_strndup (ee->data.data, ee->data.data_len)` would make the guarantee local to the function that needs it, and would keep item 4 fixed even if the receive-side check is ever restructured.

**2. `!dest_host` is a slightly weaker guard than the commit message implies.** (low)

`":1234"` parses to a non-`NULL` *empty* host, because `nns_edge_strndup (str, 0)` returns `""`, so it passes the new check and reaches `_nns_edge_connect_to()`. It is not memory-unsafe (`inet_addr("")` fails and `getaddrinfo("")` then fails, so the connection just fails), but `STR_IS_VALID (dest_host)` says what was meant. Likewise the parsed port is never range-checked: `"1.2.3.4:70000"` yields `70000`, which `htons()` silently truncates. `PORT_IS_VALID()` already exists in `nnstreamer-edge-util.h` and is used elsewhere for exactly this. Same two remarks apply to the hybrid path.

**3. The test's drift detector is one-sided — three tests can pass vacuously.** (low, but relevant to "detect the impact of future changes")

`_test_connect_raw_client()` asserts the server's command is `NE_TEST_CMD_CAPABILITY`, so a change to that enumerator is caught. Nothing pins `NE_TEST_CMD_HOST_INFO (2)`. If someone inserts a new value into `nns_edge_cmd_e` before `_NNS_EDGE_CMD_HOST_INFO`, all three `hostInfo*_n` tests still pass — the server rejects the now-unexpected command at the `cmd.info.cmd != _NNS_EDGE_CMD_HOST_INFO` branch and closes with the very same "ERROR command, then EOF" the tests assert, without `_nns_edge_cmd_string_is_valid()` ever being exercised. `_test_raw_server_thread()` already reads the client's host-info header, so one line there —

```c
EXPECT_EQ (info.cmd, (uint32_t) NE_TEST_CMD_HOST_INFO);
```

— would pin the value and remove the vacuous pass.

**4. `ne_test_cmd_info_s` duplicates a struct that cannot be included.** (informational)

`nns_edge_cmd_info_s` lives in the `.c` file, so the duplication is unavoidable as things stand, but a layout change would surface as a socket timeout rather than a compile error. If the struct ever moves to an internal header, the test should include it instead. Related to point 3: what protects both is the handful of field assertions in the test helpers, so it is worth keeping those complete.

**5. The MQTT hybrid guard has no test, as the PR body says.** (informational)

I agree with the reasoning — `unittest_nnstreamer-edge-mqtt` is deliberately outside `ADD_TEST()`, so a test there would not run in the required check. The residual is that a future edit to those two guards is caught by nothing. One way to close it without a broker: have both paths call a single small helper (or fold the terminator requirement into an explicit length-taking contract), so the TCP tests cover the shared logic.

**6. Nits.** `_nns_edge_cmd_string_is_valid()` could take `const nns_edge_cmd_s *`. The `ASSERT_TRUE (fd >= 0)` inside the `void` helper `_test_send_host_info()` returns from the helper only, not from the test — harmless here since the caller does nothing but release the handle, just be aware of it if the helper grows.

myungjoo:
Thanks — all six points looked at, four are now in the branch (`386b676`).

**1. `nns_edge_event_parse_capability()` unbounded — taken.** `nns_edge_strndup (ee->data.data, ee->data.data_len)`. `strndup()` stops at the first null within `len`, so the existing `edgeEvent.parseCapability`, which passes `strlen()` without the terminator, is unchanged. New positive test `edgeEvent.parseCapabilityNotTerminated` hands the parser a heap buffer of exactly `strlen()` bytes with no terminator; against the old `strdup()` it is a `heap-buffer-overflow`, `READ of size 16` in `strlen` ← `nns_edge_strdup` util.c:221. That closes item 4 at the line #259 actually cites, independently of the transport check.

**2. `!dest_host` too weak — taken.** Both paths now use `STR_IS_VALID (host) && PORT_IS_VALID (port)`. You were right that `":1234"` passed: `nns_edge_strndup (str, 0)` returns `""`, not null.

Two new tests, `hostInfoEmptyHost_n` and `hostInfoInvalidPort_n` (`"127.0.0.1:70000"`). Being straight about what they are worth: both **pass against the unfixed server too**, because an empty host fails in `getaddrinfo()` and 70000 truncates to a port nothing listens on, so the connection failed a moment later and the server closed the same way. They pin the contract — refuse and report an error — rather than reproduce a fault, and the PR body now says so.

**3. Vacuous pass on an enumerator shift — taken, with a correction.** Asserting `NE_TEST_CMD_HOST_INFO` unconditionally in `_test_raw_server_thread()` fails: in `capabilityNotTerminated_n` the client refuses the capability, and `_nns_edge_close_connection()` writes an `ERROR` command before closing, so the thread reads cmd `0`, not `2`. The expected command is now a field of `ne_test_raw_server_s`, `HOST_INFO` for the positive test and `ERROR` for the negative one. Both values are pinned, and the negative test additionally asserts that the client reports the refusal on the wire.

**4. Struct duplication — noted, no change.** Nothing to do while `nns_edge_cmd_info_s` lives in the `.c` file. Point 3 is what protects it, and those assertions are now complete on both sides.

**5. Shared helper for the two terminator guards — declined.** The shared logic is one comparison (`str[len - 1] == '\0'`); wrapping it would add indirection without making the MQTT *call site* any more covered, which is the part that is actually untested. If the hybrid path grows a second reason to reject a payload, that is the moment to factor it out.

**6. Nits — `const` taken.** `_nns_edge_cmd_string_is_valid (const nns_edge_cmd_s *cmd)`. Left `ASSERT_TRUE` in `_test_send_host_info()` as it is: returning from the helper is the intent, and the caller only releases the handle afterwards.

Re-verified after the changes: `plain` and `asan` both 100% locally, `clang-format`, doxygen-tag and newline checks clean, and the revert-tree run still reports the four original faults plus the new one.

myungjoo:
> Note: this review was produced by another agent and is transcribed here as-is.
> This is the **second round** of review, on `386b676` after the force-push that answered the first review.

## Verdict

**LGTM — mergeable once the pending CI turns green.** All four points the author accepted are genuinely in the branch and do what they claim, the one declined point is defensible, and I could not find anything the rework broke. Everything below the "What changed since round 1" section is a non-blocking suggestion.

## What I re-verified

I rebuilt the tree at `386b676` twice (plain `Release`, and `Debug` + AddressSanitizer with the same flags the `build` job uses) and also rebuilt a third tree with the PR's tests kept but `src/` reverted to the base commit `7980f8f`.

**PR tree:** `ctest` 100% in both configurations, ASan clean under `ASAN_OPTIONS=detect_leaks=1`, no compiler warnings. I ran the 18 new/related tests 15 times in a row — 0 failures, so the socket-driven tests are not visibly flaky.

**Revert tree (base `src/`, PR tests):** the table in the PR body reproduces exactly, including the new round-2 test.

| test | result against unfixed `src/` |
| --- | --- |
| `edge.hostInfoNotTerminated_n` | ASan `heap-buffer-overflow` in `strtoll` ← `nns_edge_parse_host_string` util.c:135 |
| `edge.hostInfoNoMemory_n` | ASan `SEGV` in `strchr` ← util.c:131 |
| `edge.hostInfoNoPort_n` | ASan `SEGV` in `inet_addr` ← `_fill_socket_addr` internal.c:185 |
| `edge.capabilityNotTerminated_n` | ASan `heap-buffer-overflow` in `strlen` ← `nns_edge_strdup` util.c:221 |
| `edgeEvent.parseCapabilityNotTerminated` | ASan `heap-buffer-overflow` in `strlen` ← `nns_edge_strdup` util.c:221 |
| `edge.hostInfoEmptyHost_n`, `edge.hostInfoInvalidPort_n` | **pass** — exactly as the PR body states |
| `edge.capabilityTerminated`, `edgeEvent.parseCapability` | pass (positive counterparts) |

The PR body's honesty about the two that do not reproduce a fault is worth noting; it matches what I measured.

## What changed since round 1

**1. `nns_edge_event_parse_capability()` → `nns_edge_strndup()` — correct, and it does not break existing callers.**
Two length conventions exist in the tree and both still yield the same string: the library's own producer (`internal.c:1036`) passes `mem_size[0] == strlen () + 1`, while the pre-existing `edgeEvent.parseCapability` passes `strlen ()`. `strndup` stops at the first NUL within `len` and always terminates, so neither changes. There is no zero-length path that would now return `""`, because `nns_edge_event_set_data()` rejects `data_len <= 0`. Only in-tree raiser of `NNS_EDGE_EVENT_CAPABILITY` is `internal.c:1036`, so no other caller's expectation is in play. The new `edgeEvent.parseCapabilityNotTerminated` is a real reproducer, confirmed above.

**2. `STR_IS_VALID` / `PORT_IS_VALID` — no conforming peer is rejected.** This was the regression I looked hardest at, and it holds:
* `nns_edge_start()` (internal.c:1474-1479) assigns `nns_edge_get_available_port()` whenever `eh->port <= 0`, *before* `is_started` is set, and `nns_edge_connect()` (internal.c:1829) refuses to run at all unless `is_started`. So the `HOST_INFO` a client sends can never carry port 0.
* `eh->host` defaults to `"localhost"` and `nns_edge_set_info()` rejects an empty value (`STR_IS_VALID (value)`), while `"PORT"` goes through `nns_edge_parse_port_number()`. So `nns_edge_get_host_string()` can never produce an empty host or an out-of-range port.
* `NNS_EDGE_CONNECT_TYPE_CUSTOM` returns from `nns_edge_start()` before the port assignment, but it does not use the TCP handshake, so it is unaffected.
* `NNS_EDGE_NODE_TYPE_SUB` gets a port assigned on the same line even though it creates no listener — same as before this PR, and still accepted.

**3. The vacuous-pass concern is genuinely closed, and the author's correction was right.** I confirmed the correction empirically: inserting `_NNS_EDGE_CMD_FUTURE_NEW` before `_NNS_EDGE_CMD_HOST_INFO` in `nns_edge_cmd_e` and rebuilding makes **6 of the 7 socket tests fail** (the five `hostInfo*` through `_test_connect_raw_client()`'s capability assertion, and `capabilityTerminated` through the new `rs->expected_cmd`). Both wire values are now pinned in both directions. Making `expected_cmd` a field rather than a fixed assertion was necessary — in the negative test the client really does answer `ERROR`, so the unconditional line I suggested would have failed.

**4/6. Struct duplication noted, `const` taken.** Fine as they stand.

**5. Shared helper — declined.** I accept the conclusion (an MQTT test cannot run in the required check, so factoring the logic out does not cover the *call site*), but the stated rationale is now slightly out of date; see minor point 1.

Also re-checked on the final tree: `clang-format` 16 produces an empty diff on `tests/unittest_nnstreamer-edge.cc`; `doxygen-tag.sh` is clean on all five changed files; `indent.sh` output is **byte-identical** to the base commit's (14 diff lines in both, all in code this PR does not touch), so no new formatting debt. Scope is still tight: 3 `.c` files, 1 header comment, 1 test file, no API/ABI change, no protocol document exists in the tree to update. The ASan `ctest` is inside the `build` job (confirmed in `ubuntu_clean_cmake_build.yml`, with a comment saying exactly why), and `build` is a required check, so a regression blocks the merge.

## Minor points (none blocking)

**1. The "declined" rationale understates what is now duplicated.** (low)
At round 1 the shared logic really was one comparison. Round 2's own change added a second, verbatim-identical guard to both paths:
```c
if (!STR_IS_VALID (host) || !PORT_IS_VALID (port)) { ... }
```
at `internal.c:1155` and `internal.c:1746`. So the untested MQTT side now duplicates four conditions, not one, and the two copies can drift silently. A small internal `_nns_edge_parse_peer_host (const char *str, size_t len, char **host, int *port)` doing terminator + parse + range check, called from both, would put all of it under the TCP tests. The call site stays uncovered either way, so I am not asking for it in this PR — just noting that "one comparison" is no longer the whole picture.

**2. `_nns_edge_cmd_string_is_valid()` is an exclusion list, and nothing notices a new string-carrying command.** (low, this is the "detect future impact" angle)
The function returns `true` for every command it does not name. The enum-shift experiment above shows that a *renumbering* is caught, but a future `_NNS_EDGE_CMD_SOMETHING_INFO` that also carries a peer-chosen string would be silently unchecked, and no test would go red. A one-line comment next to `nns_edge_cmd_e` (internal.c:84-91) pointing at the checker would put the coupling where the contributor adding the enumerator will see it.

**3. The public API's contract changed without a doc note.** (low)
No signature or ABI change, so nothing is *due*. But `nns_edge_event_parse_capability()` now bounds the copy by the event's `data_len`, and `include/nnstreamer-edge-event.h:70-78` says nothing about it. A custom-connection module raising `NNS_EDGE_EVENT_CAPABILITY` with a `data_len` shorter than its string would now get a truncated capability rather than the full one. A `@remarks`/`@note` in the public header stating that the returned string is bounded by the event data length would make that a contract instead of an implementation detail.

**4. `_fill_socket_addr()` itself is still unguarded.** (informational, out of scope)
All three `_nns_edge_connect_to()` call sites now validate the host, so the `inet_addr(NULL)` SEGV of #259 item 3 is closed. But `_fill_socket_addr()` (internal.c:179) still passes `host` straight to `inet_addr()`/`getaddrinfo()` with no null check, so the guarantee lives in the callers. That is the same argument you accepted for point 1 — worth a `STR_IS_VALID` at the top of it someday, not here.

**5. Two small things in the hybrid loop.** (informational)
`ret = NNS_EDGE_ERROR_INVALID_PARAMETER;` before each `continue` is dead — the next iteration overwrites `ret` from `nns_edge_mqtt_get_message()` before it can be returned. It reads as if the error escapes; it never does. And because `nns_edge_mqtt_get_message (..., 0U)` is an infinite timeout that lands in `pthread_cond_wait()`, a malformed payload now makes `nns_edge_connect()` wait for the next publish rather than return. That is the same shape the loop already had for a failed connect, and it is strictly better than the old SEGV, so no change requested — flagging it so it stays a conscious choice.
I did confirm `msg[msg_len - 1]` is in bounds: the mosquitto callback `memdup`s exactly `payloadlen` bytes.

## Merge

Approve. The two required checks are `DCO` (already passing) and `build` (pending at the time of writing) — merge once `build` is green. If #260 lands first, the `db70986` clang-format commit drops out on rebase as the PR body says; I re-confirmed it is non-semantic (`clang-format` produces an empty diff on the final file).

myungjoo:
Round 2 answered in `e42c06c`. Three of the five taken, two left alone with reasons.

**1. The duplicated guard — taken, and you were right that the round-1 rationale had expired.** My own round-2 change is what made it four conditions in two places, so the argument I used to decline no longer described the code. Both paths now go through one helper:

```c
static bool
_nns_edge_parse_peer_host (const char *str, nns_size_t len, char **host, int *port)
```

terminator, parse, `STR_IS_VALID`, `PORT_IS_VALID`, and it frees the host it allocated when the range check fails, so the caller only checks the return. `_nns_edge_accept_socket()` and `_mqtt_hybrid_direct_connection()` each lost their copy. The MQTT *call site* is still uncovered — that has not changed — but the rule it applies is now exercised by all five `hostInfo*` tests, and the two copies can no longer drift.

**2. The exclusion list — taken, as the comment you suggested.** `nns_edge_cmd_e` now carries `@note A command carrying a string the peer sizes should be listed in _nns_edge_cmd_string_is_valid(), which is what bounds that string.` Agreed that a test cannot catch this one: a command that does not exist yet cannot be sent by a test. Putting the coupling in front of whoever adds the enumerator is the available remedy.

**3. The public contract — taken.** `include/nnstreamer-edge-event.h` gets a second `@remarks`: the returned string is bounded by the event data length, so a producer should pass the whole capability string. That makes the truncation you describe a documented contract rather than a surprise.

**4. `_fill_socket_addr()` — left alone, agreed out of scope.** All three call sites validate now; hardening the function itself belongs with #259 item 5 or a cleanup of its own.

**5. Dead `ret` before `continue` — removed** along with the guards that carried it. The indefinite wait stays, deliberately: `nns_edge_mqtt_get_message (..., 0U)` blocking is pre-existing behaviour on this loop, it is what #259 item 8 is about, and changing it here would put a scheduling change in a bounds-checking patch.

Re-verified on `e42c06c`: `plain` and `asan` both 100%, `clang-format` empty diff, doxygen-tag and newline clean, GNU `indent` produces no change on any file this PR touches. The revert tree still separates the two kinds of test exactly as the body claims — five report AddressSanitizer against the base `src/`, `hostInfoEmptyHost_n` / `hostInfoInvalidPort_n` / `capabilityTerminated` pass there.

myungjoo:
> Note: this review was produced by another agent and is transcribed here as-is.
> This is the **third round** of review, on `e42c06c`, after the changes that answered round 2.

## Verdict

**LGTM — approve, mergeable once the pending `build` check is green.** The `_nns_edge_parse_peer_host()` extraction is behaviour-preserving at both call sites; I looked specifically for a refactor regression (leak, double free, uninitialised use, changed release order, changed control flow) and found none. The two points declined in round 2 are both correctly reasoned, and I re-derived their premises myself rather than taking them on trust. Everything under "Remaining points" is informational.

## The new helper — line-by-line equivalence check

This is what I spent most of the round on, since it is the only new code since round 2.

**Output params.** `_nns_edge_parse_peer_host()` writes `*host = NULL; *port = 0;` before any early return, so neither call site can read an uninitialised value on any path. `_nns_edge_accept_socket()` additionally still initialises `dest_port = 0` at the declaration (this was the round-1 fix and it is still there); `server_port` is loop-local and initialised in `_mqtt_hybrid_direct_connection()`. `bool parsed` is uninitialised at its declaration, but every `goto error` that precedes its assignment leaves it unread — I confirmed with a `-Wall -Wextra` build that produces no diagnostic on either file.

**Free discipline — no leak, no double free.** The helper frees the host it allocated when the range check fails, via `SAFE_FREE (*host)`, and `SAFE_FREE` NULLs the pointer through the `char **`, so the caller's variable really does become `NULL`. That is what makes `_nns_edge_accept_socket()`'s unconditional `SAFE_FREE (dest_host)` at `error:` a no-op rather than a double free. I checked this empirically two ways: the ASan run with `detect_leaks=1` is clean, and a mutant that deletes the helper's `SAFE_FREE (*host)` is still leak-free — because on the TCP path the `error:` label frees it anyway. So the free inside the helper is redundant for `_nns_edge_accept_socket()` and load-bearing for `_mqtt_hybrid_direct_connection()`, where `server_ip` is loop-local and nothing else would release it before the `continue`. It is in the right place.

**Release order and lifetime.** Both call sites still release the source buffer immediately after the call and before the `if (!parsed)` branch — `_nns_edge_cmd_clear (&cmd)` on one side, `SAFE_FREE (msg)` on the other — which is the order the pre-refactor code had. `dest_host` / `server_ip` come from `nns_edge_strndup()`, so they are independent allocations and survive the clear; there is no use-after-free in the `nns_edge_logd ("Parsed server info: ...")` line or in `_nns_edge_connect_to()`.

**Control flow.** TCP: previously `parse` → `clear` → `if (!STR_IS_VALID || !PORT_IS_VALID) goto error`; now `parse+validate` → `clear` → `if (!parsed) goto error`. Same branch, same label, same log line. MQTT: previously terminator check (with its own `SAFE_FREE (msg)` and `continue`) then parse then range check then `continue`; now one call, one `SAFE_FREE (msg)`, one `continue`. The removed `ret = NNS_EDGE_ERROR_INVALID_PARAMETER;` was genuinely dead — `continue` in the `do { } while (TRUE)` re-enters at `nns_edge_mqtt_get_message()`, which reassigns `ret` before any path can return it — and `ret` still cannot be returned uninitialised, because the first iteration always assigns it.

**Bounds.** The helper reads `str[len - 1]` after rejecting `len == 0`, and it reads `msg` before `SAFE_FREE (msg)`. On the TCP side `len` is `cmd.info.mem_size[0]`, which is exactly the allocation size — `_nns_edge_cmd_receive()` does `cmd->mem[n] = nns_edge_malloc (cmd->info.mem_size[n])` — so the read is in bounds; on the MQTT side the paho callback `memdup`s exactly `payloadlen`. The helper's terminator check is therefore redundant on the TCP path (`_nns_edge_cmd_string_is_valid()` already enforced it inside `_nns_edge_cmd_receive()`) and is the only such check on the MQTT path. Harmless duplication, correct in both directions.

**Does the shared helper actually get exercised by the TCP tests?** Partly, and more than I expected. I mutated the helper and re-ran:

| mutation | result |
| --- | --- |
| drop `STR_IS_VALID (*host)` from the guard | `edge.hostInfoNoPort_n` **fails** — ASan `SEGV` in `inet_addr` |
| drop `PORT_IS_VALID (*port)` (same edit) | not detected on its own |
| drop `SAFE_FREE (*host)` on the failure path | not detected (the TCP caller frees at `error:`) |

So the half of the rule that the round-1 SEGV came from is pinned by a real reproducer, which is the useful part. The port range check and the empty-host case remain assert-the-contract rather than reproduce-a-fault, exactly as the PR body already states, and the MQTT-only aspects of the helper stay uncovered — unchanged from round 2, and unavoidable without a broker.

## What I re-verified independently

Two builds of `e42c06c` (`Release`, and `Debug` + the `build` job's ASan flags) and a third tree with the PR's tests and `src/` + `include/` reverted to `7980f8f`.

* **PR tree:** `ctest` 100% in both configurations; ASan clean under `ASAN_OPTIONS=detect_leaks=1`; no warnings, including under `-Wall -Wextra`. I ran the 18 new/related tests 12 times in a row — 0 failures, so the socket-driven tests are not flaky.
* **Revert tree:** reproduces the PR body's table exactly — `hostInfoNotTerminated_n` heap-buffer-overflow, `hostInfoNoMemory_n` SEGV, `hostInfoNoPort_n` SEGV, `capabilityNotTerminated_n` heap-buffer-overflow, `edgeEvent.parseCapabilityNotTerminated` heap-buffer-overflow; and `hostInfoEmptyHost_n`, `hostInfoInvalidPort_n`, `capabilityTerminated`, `edgeEvent.parseCapability`, `edgeUtil.parseHostString` all pass there. The body's honesty about which is which continues to match what I measure.
* **Back-compat, re-derived rather than inherited from round 2.** `nns_edge_set_info()` rejects an empty value with `STR_IS_VALID`, and `"PORT"` goes through `nns_edge_parse_port_number()`, which itself applies `PORT_IS_VALID` and substitutes an available port for 0; `nns_edge_start()` assigns `nns_edge_get_available_port()` whenever `eh->port <= 0`, before `is_started`, and `nns_edge_connect()` refuses to run unless `is_started`. So `nns_edge_get_host_string (eh->host, eh->port)` can never produce a `HOST_INFO` string the new check rejects. The hybrid publisher sends `strlen (msg) + 1`. No conforming peer, current or released, is refused.
* **Blast radius unchanged.** `_nns_edge_cmd_string_is_valid()` returns `true` immediately for every other command, and all reception of `HOST_INFO` / `CAPABILITY` goes through `_nns_edge_cmd_receive()` (the only two consumers are internal.c:1051 and internal.c:1172), so `TRANSFER_DATA` and `_nns_edge_message_handler()` are untouched. `nns_edge_parse_host_string()` still has no caller outside `_nns_edge_parse_peer_host()`.
* **Scope and formatting.** 3 `.c` files, 1 public-header comment, 1 internal-header comment, 1 test file; no API/ABI change; no protocol document exists in the tree to update. `clang-format` gives an empty diff on `tests/unittest_nnstreamer-edge.cc`; all changed files end with a newline; both new functions carry `@brief`. I could not reproduce `indent.sh` faithfully — the GNU indent available to me (2.2.12) reformats the *base* files just as heavily as the PR's — so I am deferring that one to the CI job; note it is not a required check in any case.

## The two declined points — both sound

**`_fill_socket_addr()` guard (declined, out of scope). Agreed.** There are exactly three `_nns_edge_connect_to()` call sites (internal.c:1189, 1777, 1896). The first two now go through `_nns_edge_parse_peer_host()`; the third is `nns_edge_connect()`, which applies `STR_IS_VALID (dest_host)` and `PORT_IS_VALID (dest_port)` to its own arguments before anything else. So `inet_addr(NULL)` is unreachable today and hardening the function itself is a separate cleanup, as the author says.

**Indefinite wait in the hybrid loop (declined, deferred to #259 item 8). Agreed.** I traced it: `nns_edge_mqtt_get_message (..., 0U)` → `nns_edge_queue_wait_pop (…, timeout = 0)` → `nns_edge_cond_wait_until (q, 0)`, i.e. an untimed wait; a wake-up with an empty queue makes `_pop_data()` fail and the function returns `NNS_EDGE_ERROR_IO`, which `break`s the loop — so a malformed payload cannot spin. And "keep waiting instead of returning" is the shape the loop already had before this PR, which only ever broke on a successful connect or a failed `get_message`. Changing it here would put a scheduling change in a bounds-checking patch. Right call.

## Remaining points (none blocking, none new-code defects)

1. **Doc nit, `nns_edge_parse_host_string()`.** (informational) The `@note` says "The output params are always set", but when one of `host` / `port` is null the function returns before setting either — which is what `edgeUtil.parseHostStringInvalidParam04_n` asserts. "The output params are always set when both are given" would match the code exactly. Not worth a force-push on its own.
2. **`_test_raw_server_thread()` calls `EXPECT_EQ` from a non-main thread.** (informational) Safe with gtest's pthread support on the platforms this CI runs, and the assertion is genuinely useful there (it is what pins `NE_TEST_CMD_HOST_INFO` / `NE_TEST_CMD_ERROR`). Just noting it, as a failure there is reported at join time rather than at the assertion site.
3. **The MQTT call site is still the only untested edge of the change**, as it was in round 2 — but the *rule* it applies is now shared, and one third of that rule (`STR_IS_VALID`) is pinned by a hard reproducer, per the mutation table above. That is a real improvement over round 2 and I would not ask for more here.

## Merge

Approve. `DCO` is passing; `build` (which carries the ASan `ctest`) was still pending when I ran, and everything I ran locally in that job's two configurations is green. Merge once `build` is green. If #260 lands first, the `db70986` clang-format commit drops out on rebase as the PR body says.

myungjoo:
Round 3 answered in `d24277a`, which is `e42c06c` plus the two-line doc correction.

**1. The `@note` on `nns_edge_parse_host_string()` — taken, verbatim.** It now reads "The output params are always set **when both are given**", in both `nnstreamer-edge-util.c` and the header. You are right that the old wording contradicted `parseHostStringInvalidParam04_n`, which asserts the opposite; the sentence described an earlier draft where the guards were independent.

**2. `EXPECT_EQ` from the server thread — left as is.** Agreed on both halves: gtest is thread-safe with pthreads here, and moving the assertion to the main thread would mean shipping the received command back through the struct, which buys a better failure location at the cost of the plumbing. Worth knowing when reading a failure, not worth the code.

**3. The MQTT call site — nothing further.** Same conclusion as yours.

No code changed in this push, only the two comment lines, so the mutation results and the revert-tree table from round 3 still stand as measured. `plain` re-run after the edit: 100%.

myungjoo:
This comment relays a review performed by another agent. It is a **final delta-only check**: it re-examines nothing but the difference between the commit the previous (3rd) review looked at (`e42c06c`) and the current head (`d24277a`).

## Delta

`git diff e42c06c d24277a` touches 2 files, 1 line each, and both are the identical doc comment on `nns_edge_parse_host_string()`:

```
- * @note The output params are always set. On failure the host is null and the port is zero, ...
+ * @note The output params are always set when both are given. On failure the host is null and the port is zero, ...
```

(`src/libnnstreamer-edge/nnstreamer-edge-util.c` and the matching declaration in `nnstreamer-edge-util.h`.)

`git range-diff` over the two-commit series confirms the shape: commit 1 (`[Fix] validate the strings a peer sends before parsing them`) differs only by that comment wording, and commit 2 (`[Test] cover the malformed host info and capability commands`) is byte-identical (`2: e42c06c = 2: d24277a`). No code, no test, no build-file change slipped in with the force-push.

## Accuracy of the new wording

It matches the implementation exactly. `nns_edge_parse_host_string()` does:

```c
if (!host || !port)
  return;

*host = NULL;
*port = 0;
```

so when either output pointer is NULL the function returns before assigning anything — including to the non-NULL one. "always set when both are given" is the precise statement; the previous "always set" was the overclaim flagged last round.

`edgeUtil.parseHostStringInvalidParam04_n` in `tests/unittest_nnstreamer-edge.cc` pins exactly this: it passes `NULL` for one output param and asserts the other retains its pre-call sentinel (`port == -1`, `host == NULL`). Comment, code, and test agree.

## Verdict

**Approved — ready to merge** once CI goes green (the checks were still running at the time of this review; only the doc comment changed, so no behavioral risk to the build).

The other two informational notes from the 3rd review remain unaddressed, which is fine — they were explicitly non-blocking and can be left as-is or picked up separately.

myungjoo:
Rebased onto `main` now that #260 has landed. As the body said it would, the shared `clang-format` commit dropped out on its own (`git rebase` skipped it as already applied); `git range-diff` reports the two remaining commits identical to the reviewed `d24277a`, so nothing in the change itself moved.

Re-verified on the rebased tree, which now also carries #260's metadata deserializer: `plain` and `asan` `ctest` both 100%, ASan clean with `detect_leaks=1`, and `clang-format` still an empty diff on the test file.

myungjoo:
Both open points from the automated re-review are now closed in `00c09f7`. The approval above was on `2529f60`; the delta is three comment lines, one function body, and one test.

**1. The enum's wire values and the test — taken, with a correction to the premise.** `nns_edge_cmd_e` now carries a second `@note` saying the values go on the wire, that `tests/unittest_nnstreamer-edge.cc` mirrors them, and that a member added in the middle fails those tests until they are updated.

The correction: they would not pass silently. `_test_connect_raw_client()` asserts the server's first command is `CAPABILITY` and returns `-1` otherwise, which trips `ASSERT_TRUE (fd >= 0)` in all five `hostInfo*` tests, and `_test_raw_server_thread()` asserts the command the client writes back against `rs->expected_cmd`, which pins `HOST_INFO` in `capabilityTerminated` and `ERROR` in `capabilityNotTerminated_n`. An earlier review round checked this empirically by inserting an enumerator before `_NNS_EDGE_CMD_HOST_INFO`: six of the seven socket tests failed. The note is still worth having — it puts the obligation where the contributor will see it — but the tests are a pin, not a silent pass.

**2. The partial-null contract — taken.** `nns_edge_parse_host_string()` now writes whichever pointer it was given rather than nothing at all when one is missing, so there is one rule instead of two: every output param given is always set. `parseHostStringInvalidParam04_n` asserts the new behaviour on both halves, plus the no-colon case through a null host.

**3. The extra byte in `nns_edge_strndup()` — no change**, agreed it is not an off-by-one. `len + 1` is the general contract of that helper; the capability path passing a length that already counts the terminator just leaves one byte unused.

Re-verified on `00c09f7`: `plain` and `asan` `ctest` both 100%, ASan clean with `detect_leaks=1`, `clang-format` empty diff, doxygen-tag and newline clean. GNU `indent` still wants changes only in `nns_edge_start_discovery()`, `nns_edge_stop_discovery()` and `nns_edge_generate_version_key()`, none of which this PR touches.

myungjoo:
## Re-review of the `2529f60` → `00c09f7` delta

*Relayed on behalf of another agent that performed this review.* This is a **delta-only** review: the APPROVED review already given on `2529f60` still stands for everything underneath, and the only thing examined here is what changed since it — the two non-blocking points that approval raised.

**Verdict: approve, ready to merge.** No blocking issues. Two cosmetic, optional nits below; neither is worth another round trip.

### 0. Scope of the delta

`git range-diff 2529f60~2..2529f60 00c09f7~2..00c09f7` shows exactly the four intended edits and nothing else. Both revisions share the same merge base, the two commits are otherwise byte-identical, and the commit messages/structure are unchanged. No unintended changes crept in with the force-push.

### 1. `nns_edge_parse_host_string()` — the "always fill what you were given" contract

Old behaviour: return immediately unless **both** `host` and `port` are non-NULL. New behaviour: NULL/zero each *given* out-param up front, then fill each *given* one.

**Production behaviour is unchanged.** With both out-params non-NULL the two versions are step-for-step equivalent — the old unconditional `*host = NULL; *port = 0;` is now two guarded stores whose guards are both true, and the same for the two stores inside the `if (p)` branch. Both production call sites reach this function only through `_nns_edge_parse_peer_host()` (`src/libnnstreamer-edge/nnstreamer-edge-internal.c:445`), which itself unconditionally writes `*host` and `*port` before doing anything else, so it can *only* be invoked with both non-NULL. Its two callers — the TCP host-info path at `nnstreamer-edge-internal.c:1179` and `_mqtt_hybrid_direct_connection()` at `:1767` — both pass the address of a real local in both slots. So the relaxed contract is unobservable outside the unit test.

Blast radius is also bounded: `nnstreamer-edge-util.h` appears in no `INSTALL` rule (`src/CMakeLists.txt:41-45` install only the four public headers) and is absent from `debian/nnstreamer-edge-dev.install`. It is an internal header, so there is no out-of-tree consumer whose expectations could be broken.

**Doxygen `@note` matches the code.** "Every output param given is always set" is exact — both stores at the top of the function are unconditional on anything but the param being non-NULL, and every later `return` path leaves those values in place. The `.h` and `.c` copies are identical, as they were before.

*Optional nit, non-blocking, and pre-existing (this half of the sentence is unchanged by the delta):* "On failure the host is null and the port is zero" is loose for input like `":1234"` — `strchr` finds the colon at offset 0, so the host comes back as `""` (non-NULL, `nns_edge_strndup(str, 0)`) with the port set to 1234. Callers are safe because `_nns_edge_parse_peer_host()` gates on `STR_IS_VALID (*host)`, and `edge.hostInfoEmptyHost_n` covers exactly that input end-to-end. Only the prose is imprecise; the code is right.

### 2. `edgeUtil.parseHostStringInvalidParam04_n`

The rewritten test **does** exercise the new contract rather than merely being adjusted to keep passing. All three sub-cases would fail against the pre-delta implementation:

- `("192.168.0.1:5001", NULL, &port)` → old code returned before touching `port`, leaving the `-1` sentinel; the test now requires `5001`.
- `("192.168.0.1:5001", &host, NULL)` → old code left `host` at `NULL`; the test now requires `"192.168.0.1"` (and frees it).
- `("192.168.0.1", NULL, &port)` → this one is genuinely *new* coverage: it pins the "reset even when there is nothing to parse" half of the contract, which the other two sub-cases do not reach. Old code left `port` at `-1`; it must now be `0`.

So it is a real regression guard, not a rubber stamp.

**On the `_n` postfix: fine to keep.** The input under test is still a degenerate/invalid parameter combination (a NULL out-param), which is what the sibling `parseHostStringInvalidParam01..03_n` names denote in this file; the third sub-case is additionally a parse failure. Renaming would break the numbering of the group for no real gain.

*Cosmetic nit:* siblings 01–03 poison the out-param with `char *host = (char *) &host;` so that "the function never wrote" is distinguishable from "the function wrote NULL", whereas 04 starts from `NULL`. It still catches the regression here (the `EXPECT_STREQ` fails on a NULL host), so this is purely a consistency observation.

### 3. The `nns_edge_cmd_e` wire-value `@note`

The note is factually accurate. `tests/unittest_nnstreamer-edge.cc:4691-4693` does mirror the values (`NE_TEST_CMD_ERROR (0)`, `NE_TEST_CMD_HOST_INFO (2)`, `NE_TEST_CMD_CAPABILITY (3)`), does use them to talk to a real peer over a real socket, and does assert on received commands.

The author's claim — that inserting a member in the middle makes the tests **fail** rather than silently pass — was checked by experiment, not by reading. Inserting `_NNS_EDGE_CMD_INSERTED_PROBE` right after `_NNS_EDGE_CMD_ERROR` (shifting HOST_INFO 2→3, CAPABILITY 3→4) and rebuilding took the suite from **197 passed** to **191 passed / 6 failed**:

```
[  FAILED  ] edge.hostInfoNotTerminated_n
[  FAILED  ] edge.hostInfoNoMemory_n
[  FAILED  ] edge.hostInfoNoPort_n
[  FAILED  ] edge.hostInfoEmptyHost_n
[  FAILED  ] edge.hostInfoInvalidPort_n
[  FAILED  ] edge.capabilityTerminated
```

The mechanism matches what the author described, with one refinement worth recording for whoever reads the note later:

- The five `hostInfo*` failures all come from `_test_connect_raw_client()` (`:4817`): its `info.cmd != NE_TEST_CMD_CAPABILITY` check makes it return `-1`, which trips `ASSERT_TRUE (fd >= 0)` in `_test_send_host_info()`. This hook is exactly as load-bearing as claimed.
- `edge.capabilityTerminated` fails on its own body (`nns_edge_connect` returns an error and `received` stays `0`), because the real client no longer recognises wire value 3 as CAPABILITY.
- `rs->expected_cmd` in `_test_raw_server_thread()` (`:4986`) is the weaker of the two hooks: `capabilityNotTerminated_n` still *passes* under the shift, since its expectation is already `NE_TEST_CMD_ERROR` and the client still errors out — just for a different reason. That does not contradict the note, which only claims that a mid-enum insertion fails those tests; it does, loudly. It just means the positive `capabilityTerminated` test, not `expected_cmd` alone, is what catches the CAPABILITY shift.

I also confirmed the contrast the wording implies: appending `_NNS_EDGE_CMD_APPENDED_PROBE` immediately before `_NNS_EDGE_CMD_END` keeps all **197** tests green. "A member added *in the middle*" is therefore the right qualification, and the note does not over-claim.

### 4. Build and test

ASan build of `00c09f7` on Ubuntu/gcc with `-DENABLE_TEST=ON`, `-DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g"` (and the matching CXX/linker flags):

- Build clean, exit 0, no new compiler warnings.
- `unittest_nnstreamer-edge`: **197/197 pass**, no AddressSanitizer reports (the new test frees the host it now receives, so no leak from the delta).
- Targeted subset `edgeUtil.parseHostString*:edge.hostInfo*:edge.capability*`: 12/12 pass.

The required status checks here are DCO and build, and ASan runs inside the build job — it is clean on this revision.

### Verdict

**LGTM — merge.** The contract change is a strict relaxation that no production caller can observe, the doxygen text now matches the code exactly, the updated test genuinely pins the new behaviour (including one assertion that is new coverage), the enum note is accurate and empirically verified in both directions, and the delta contains nothing beyond the two requested fixes. The two nits above are cosmetic and can be ignored or folded into some future touch of these files.

myungjoo:
Both cosmetic nits addressed in `e48decb` — one taken, one declined with a reason.

**1. The `@note` wording — taken, and it needed it.** You are right that `":1234"` is not a "failure" of this function: it splits on the colon and hands back an empty host with a real port. The note now says what actually happens — the host is null and the port zero when there is no `':'` to split on, the host is an empty string when the input starts with one, and either way the caller checks the host. `nns_edge_parse_host_string(":5001")` is now asserted in `edgeUtil.parseHostString`, so the sentence has a test behind it rather than just my word. That empty-host case was previously only covered at the wire level by `hostInfoEmptyHost_n`.

**2. Poisoning the out-param in `parseHostStringInvalidParam04_n` — declined.** Its siblings poison `host` because they assert `host == NULL` afterwards, and starting from `NULL` would make that assertion pass without the function writing anything. Test 04 asserts the opposite — `EXPECT_STREQ (host, "192.168.0.1")` — which already proves a write happened. Poisoning there would trade a clean gtest failure on `NULL` for a garbage read through a stack address if the write ever stopped happening, so `NULL` is the better initial value for this particular assertion.

Re-verified on `e48decb`: `plain` and `asan` `ctest` both 100%, ASan clean under `detect_leaks=1`, `clang-format` empty diff, doxygen-tag and newline clean.

myungjoo:
Rebased again onto `main` now that #263 has landed (13 commits, including two that touch `_nns_edge_connect_to()` and `nns_edge_start()` a few lines from this change).

No semantic overlap: #263 checks the host string on the **sending** side, where `nns_edge_get_host_string()` can return null under allocation failure (#259 item 10). This PR checks what a peer sends on the **receiving** side. The two are complementary and the fix commit rebased with no conflict at all.

The only conflict was the include block of `tests/unittest_nnstreamer-edge.cc`, where #263 added `<dirent.h>`, `<stdlib.h>`, `<string.h>` and `<thread>` next to the `<arpa/inet.h>` / `<sys/socket.h>` this PR adds. Resolved by keeping both, in the order `clang-format` wants. `git range-diff` against the reviewed `e48decb` shows that include block as the only difference; the fix commit is identical.

Re-verified on the merged result: `plain` and `asan` `ctest` both 100% (the suite now also carries #263's thread-guard and fd-count tests), ASan clean with `detect_leaks=1`, `clang-format` empty diff.

myungjoo:
Naming nit closed in the head above: `parseHostStringInvalidParam04_n` is now `parseHostStringPartialOutput`, with the brief changed to "one output param at a time". You are right that the `_n` no longer fitted — after the contract change the body asserts what *is* parsed when only one pointer is given, which is a positive assertion; the three genuinely negative cases keep their `01_n`-`03_n` numbering.

`plain` and `asan` both 100% after the rename. Thanks for catching that the rebase note needed to be separated from the substance — `git diff e48decb a5f2215` being exactly #263's content is what I wanted the history to show.

PR #263, opened 2026-09-07
[Fix] plug the smaller leaks and out-of-memory crashes of #259 item 10

myungjoo:
Addresses item 10 ("Smaller leaks") of #259, plus the NULL dereference listed with it.

Each bullet of that item is one commit:

1. **`nns_edge_start()` did not look at `is_started`.** A second call recreated both threads (leaking the first pair, so `nns_edge_release_handle()` joined only the last one and the orphaned send thread outlived the handle), failed to bind a second listener socket - leaking the running listener's descriptor and clearing `is_started` on a started handle - and, for hybrid/mqtt handles, replaced `eh->broker_h` with a new connection. The start now returns early when the handle is started, the two thread creators are no-ops while their thread runs, and both broker connections go through a helper that closes the previous `broker_h` first. That helper also covers `_nns_edge_start_mqtt_sub()`, which reconnects after the broker drops the connection.
2. **Both mqtt message callbacks ignored `nns_edge_queue_push()`.** The queue only takes ownership on success, so a failed push dropped the last reference to the message. They now free it.
3. **`nns_edge_data_clear_info()` left a freed metadata in the handle.** `nns_edge_metadata_create()` does not touch its output when the allocation fails, so an out-of-memory there turned the next `set_info` into a use-after-free and `nns_edge_data_destroy()` into a double free. The field is cleared as soon as the old metadata is gone, and a later call recreates it.
4. **`_nns_edge_connect_to()` passed the result of `nns_edge_get_host_string()` to `strlen()`.** A client that could not allocate that string crashed instead of failing the connection.

## Tests

Three of the four bugs only trigger on an allocation failure, so `unittest_nnstreamer-edge` defines its own `calloc()` and `nns_edge_get_host_string()`. Both behave exactly like the originals until a test asks them to fail, and, as the executable is resolved before the shared library, the library's own calls land on them.

| test | without the fix |
| --- | --- |
| `edge.startTwice` | second start returns `-5`, and one descriptor is still open after the handle is released |
| `edgeCustom.startTwice` | the custom connection is started twice, and the release then hangs on the orphaned send thread |
| `edge.connectHostStringAllocFail_n` | SEGV in `strlen()` inside `_nns_edge_connect_to()` |
| `edgeData.clearInfoAllocFail_n` | ASan reports heap-use-after-free in `nns_edge_metadata_find()` |

All four were run against a tree with these sources reverted to `main`, and fail there. The full suite passes with and without AddressSanitizer.

The two mqtt leaks (2, and the `broker_h` reconnect in 1) are not covered: reaching them needs a live broker *and* an allocation failure inside the broker's own callback thread.

## Note on the first commit

`[CodeClean/Test] apply clang-format to unittest_nnstreamer-edge.cc` is cherry-picked from #260. That file has never matched `.clang-format`, so `static.check` fails on any pull request touching it before looking at the change itself. Whichever of the two lands first, the other rebases cleanly onto it.


myungjoo:
본 리뷰는 별도 컨텍스트의 리뷰 에이전트가 작성한 것을 옮긴 것입니다.

**Note:** This review was produced by a separate automated review agent, asked to independently audit this PR against issue #259 item 10 ("Smaller leaks"), and is being transcribed here verbatim as its comment.

## Scope checked

Read all 7 changed files (`src/libnnstreamer-edge/nnstreamer-edge-data.c`, `nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-mqtt-paho.c`, `tests/nnstreamer-edge-custom-test.c`, `tests/unittest_nnstreamer-edge-custom.cc`, `tests/unittest_nnstreamer-edge.cc`), the code paths they touch (`nns_edge_metadata_destroy/create`, the mqtt broker close/connect helpers, the connection-establishment and start/stop lifecycle), the required CI workflow (`ubuntu_clean_cmake_build.yml`), and issue #259's item 10 text.

Also built the PR (WSL, gcc 11, `-DENABLE_TEST=ON`, both plain and `-fsanitize=address -fno-omit-frame-pointer`) and ran the full suite: **100% pass in both modes.** To verify the new tests actually catch the bugs (not just pass trivially), I additionally rebuilt a tree that keeps the PR's test commit but reverts only the 4 source files to `main` (`7980f8f`), and ran that under ASan:

- `edge.startTwice` fails exactly as the PR describes: second `nns_edge_start()` returns `-5` ("Failed to create listener, cannot bind socket"), and the fd count after release is off by one.
- `edge.connectHostStringAllocFail_n` reproduces a real SEGV under ASan (`AddressSanitizer: SEGV ... in __interceptor_strlen ... in _nns_edge_connect_to nnstreamer-edge-internal.c:1020`), matching the PR's claim precisely.
- `edgeCustom.startTwice` then hangs the process (consistent with the PR's own description: "the release then hangs on the orphaned send thread"), which is why I could not run the remainder of the reverted-source suite to completion — I killed it once the above was confirmed.

This is strong independent evidence that the fixes are real and the regression tests are load-bearing, not decorative.

## Does it solve the stated problem? ✅ (with one gap, see Major #1 below)

All 5 bullets of #259 item 10 (the 4 "smaller leaks" + the adjacent NULL-deref) are addressed:

1. `nns_edge_start()` now checks `eh->is_started` and returns early; `_nns_edge_create_send_thread()`/`_nns_edge_create_socket_listener()` are no-ops when their thread handle is already set; both mqtt broker-connect call sites go through a new `_nns_edge_connect_to_broker()` helper that closes any existing `eh->broker_h` first (covers both the `nns_edge_start()` path and the `_nns_edge_start_mqtt_sub()` reconnect path).
2. Both mqtt `on_message`/`mqtt_cb_message_arrived` callbacks now check `nns_edge_queue_push()`'s return and free `msg` on failure. Verified against `nns_edge_queue_push()` in `nnstreamer-edge-queue.c`: every failure path (invalid param, full non-leaky queue, node alloc failure) returns before touching `data`, so this can't double-free.
3. `nns_edge_data_clear_info()` now nulls `ed->metadata` immediately after destroying the old one, before attempting `nns_edge_metadata_create()`. Verified `nns_edge_metadata_destroy(NULL)` (now reachable from `nns_edge_data_destroy()` after a failed re-create) safely returns `INVALID_PARAMETER` without dereferencing, and `nns_edge_metadata_set/get(NULL, ...)` do the same — so the "handle left pointing at freed metadata" UAF/double-free is genuinely closed, not just narrowed.
4. `_nns_edge_connect_to()` now checks `nns_edge_get_host_string()` for `NULL` before `strlen()`.

## Major

**#1 — The same NULL-deref pattern the PR fixes is left unfixed one function away, in code this very PR already touches.**
`src/libnnstreamer-edge/nnstreamer-edge-internal.c:1504-1509`, inside `nns_edge_start()` (the function whose `is_started` guard this PR adds):
```c
if (NNS_EDGE_CONNECT_TYPE_HYBRID == eh->connect_type) {
  char *msg;
  msg = nns_edge_get_host_string (eh->host, eh->port);

  ret = nns_edge_mqtt_publish (eh->broker_h, msg, strlen (msg) + 1);
```
This is the identical bug the PR fixes at `_nns_edge_connect_to()` (`nns_edge_get_host_string()` returns `NULL` on allocation failure, per its own implementation in `nnstreamer-edge-util.c`) — except here it crashes a hybrid **server/pub** on `nns_edge_start()` instead of a client on `nns_edge_connect()`. Given the PR's stated goal is exactly "OOM should fail the call, not crash," and it already modified this function a few lines above, this looks like an oversight rather than an intentional scope cut. Recommend adding the same `if (!msg) { ...; goto done; }` guard here before merge, or explicitly noting it as follow-up if intentionally deferred.

(Lower-severity sibling, informational only: `nnstreamer-edge-mqtt-paho.c:158` also calls `nns_edge_get_host_string()` without a NULL check, but the result only feeds `MQTTAsync_create()`, not `strlen()`, so it doesn't crash the same way — not flagging as a blocking issue.)

## Minor

**#2 — The new "already started" guard trusts a thread handle that can go stale without crashing anything, but silently disabling recovery.**
`_nns_edge_create_socket_listener()` (`nnstreamer-edge-internal.c:1229-1230`) and `_nns_edge_create_send_thread()` (`:945-946`) treat a non-zero `eh->listener_thread`/`eh->send_thread` as "still running." But `_nns_edge_socket_listener_thread()` can return on its own (e.g. `POLLERR|POLLHUP` at `:1203-1206`) without resetting the field to 0 — only `nns_edge_release_handle()`'s `pthread_join()` does that. Before this PR, a `start()` after such a silent thread death would leak the old handle but at least re-create a working listener; after this PR it silently no-ops, leaving the edge non-listening until the whole handle is released. This is a narrow, pre-existing corner case (not a new leak, and outside "item 10"'s scope), but worth a one-line comment on the guard given the trade-off it introduces in code this PR is already changing.

**#3 — Global `calloc()`/`nns_edge_get_host_string()` interposition in the test binary is a process-wide, un-isolated fault injector.**
`tests/unittest_nnstreamer-edge.cc` defines its own `calloc()` and `nns_edge_get_host_string()` (relying on the test executable's symbols taking priority over the shared library's — confirmed this actually works on this repo's Linux/ASan CI setup by building and running it). The on/off flag is a single global `bool`, toggled only around the specific library call each test wants to fail. If an earlier test in the same binary ever exits early via `ASSERT_*` without releasing its edge handle (skipping `nns_edge_release_handle()`), any background thread it left running could call `calloc()` at the moment a later, unrelated test raises the flag, injecting a spurious OOM failure into that other test. Not observed in the verification run above, but worth flagging as a source of potential CI flakiness given the technique is new to this file.

## Nit

**#4 — No fail-fast timeout on the required CI job for a hang-class regression.**
`.github/workflows/ubuntu_clean_cmake_build.yml`'s `build` job (the one gated as a required check, including its AddressSanitizer step) sets no `timeout-minutes` and `ctest` is invoked without `--timeout`. As reproduced above, a regression of the `is_started` guard reproduces exactly the "hangs on the orphaned send thread" failure mode the PR describes for `edgeCustom.startTwice` — which still eventually fails/blocks the merge via GitHub Actions' default job timeout, but only after burning hours of CI time instead of failing in seconds. Not something this PR needs to fix (repo-wide CI config, pre-existing), but flagging since the new tests are exactly the kind that would hit it on a future regression.

## Regression risk / module scope — no concerns

- Diff is confined to the 4 changed `.c` files under `src/libnnstreamer-edge/` plus their 3 corresponding test files; no unrelated module touched.
- No public API signature changes; behavior changes (`nns_edge_start()` idempotent on an already-started handle, `nns_edge_data_clear_info()` succeeding on a handle with no metadata) are additive/defensive and don't contradict any existing doc comment in `include/nnstreamer-edge.h` / `nnstreamer-edge-data.h`, so no doc update is required.
- `_nns_edge_connect_to_broker()`'s `nns_edge_mqtt_close()` call: verified against both mqtt backends (`nnstreamer-edge-mqtt-mosquitto.c:288-325`, `nnstreamer-edge-mqtt-paho.c:222-277`) — both always free the broker struct and return `NNS_EDGE_ERROR_NONE` regardless of the underlying disconnect's outcome, so the `logw`-only handling of a non-`NONE` return is effectively dead code but not a bug, and there's no leak on that path.
- The `_nns_edge_connect_to()` early-`goto error` on a NULL host string does **not** leak the just-`_nns_edge_cmd_init()`-ed `cmd`: `_nns_edge_cmd_init()` only `memset`s the struct, and `cmd.mem[0]` is never populated before the jump, so `_nns_edge_cmd_clear()` isn't needed there.

## Code size vs. scope — appropriate

+503/-204 across 7 files sounds large for "smaller leaks," but the bulk of it (`tests/unittest_nnstreamer-edge.cc`: -194/+part of the +399) is an explicitly-called-out cherry-pick of a pure clang-format commit from PR #260 (needed because that file predates the repo's clang-format adoption and fails `static.check` on any PR that touches it, per the PR description). I independently diffed the two commits' patches (`e9c0a92b...` here vs. `12b4d50d...` in #260) and they are **byte-identical**. The actual functional diff (4 source files + genuinely new test bodies) is small and tightly scoped to item 10.

## Test coverage — solid, with the two mqtt-only leaks reasonably left uncovered

4 new regression tests (`edge.startTwice`, `edgeCustom.startTwice`, `edge.connectHostStringAllocFail_n`, `edgeData.clearInfoAllocFail_n`) cover 3 of the 5 fixed bugs directly and independently reproduce the pre-fix failures (verified above). The two mqtt leaks (the `broker_h` reconnect leak, and the message-queue-push leak in the mqtt callbacks) are correctly called out in the PR body as untested, since reaching them needs a live broker plus an allocation failure inside the broker's own callback thread — a reasonable line to draw rather than building broker/thread fault-injection infrastructure for this PR.

## CI enforcement — confirmed to actually gate

Per this repo's branch protection, `main`'s only required status checks are `DCO` and `build`, and `build` (`ubuntu_clean_cmake_build.yml`) runs `unittest_nnstreamer-edge`/`unittest_nnstreamer-edge-custom` under both plain and `-fsanitize=address` ctest passes. All 4 new tests run in that job, so a future regression of any of the 3 directly-tested bugs reliably fails a required check (confirmed above: the ASan run on reverted sources both fails an assertion and SEGVs). `DCO` and `build` are already green on this PR as of this review.

## Overall

The core fixes are correct, well-scoped, and independently verified to close the bugs they claim to close. The one item I'd want addressed before merge is **Major #1** — the identical unchecked-`strlen(nns_edge_get_host_string())` pattern left in `nns_edge_start()`'s hybrid-publish path, a few lines from code this PR already edited for the same class of bug. The Minor/Nit items are worth a mention but not blockers.

myungjoo:
Thanks - all four points are addressed in `aa1f3c9` and `d2fc149`.

**Major, `nns_edge_start()` hybrid path.** Correct, and it is the better catch of the two: the same unchecked `strlen()` sits in the very function this pull request rewrites. Fixed in `aa1f3c9`. It has no test - reaching it needs an mqtt build and a live broker - which is stated in the commit message; the tcp counterpart stays covered by `connectHostStringAllocFail_n`.

**Minor 1, the guards trusting a non-zero `pthread_t`.** Fixed in `d2fc149`, but deliberately not by joining the dead thread and starting a new one. That join would run under the edge handle lock, and item 9 of #259 is going to give `_nns_edge_send_thread()` and `_nns_edge_accept_socket()` exactly that lock - the join would then deadlock, which is a worse failure than the one it repairs. The guards now read the running flag the thread clears on its way out (`eh->sending` / `eh->listening`) and fail the start loudly when the thread is gone, rather than reporting success on a handle with no listener.

**Minor 2, the process-wide interposition.** Worth stating, but both windows are one library call wide, with no assertion or early return between setting the flag and clearing it, so a test cannot leave a flag armed. The only concurrent caller during the `connectHostStringAllocFail_n` window is the server's listener thread, and neither `_nns_edge_accept_socket()` nor anything it calls uses `nns_edge_get_host_string()` - the three call sites are `_nns_edge_connect_to()`, the hybrid branch of `nns_edge_start()` and the paho url, none of them on that path.

**Nit, no timeout on the run.** Added `TIMEOUT 300` to both registered tests in `d2fc149`. The suite finishes in about 3.5 s, so a hang now fails in five minutes instead of running into ctest's 1500 s default (or, on the GBS path, which does not use ctest, the job limit).

One thing found while re-testing, which is *not* this pull request's: one local plain (non-ASan) `ctest` run out of roughly fifty hung in `unittest_nnstreamer-edge` and was killed at 300 s; 100 further runs of both binaries were clean. The reachable explanation is the lost-wakeup between `nns_edge_queue_clear()` and `_nns_edge_send_thread()` that #259 lists under "related non-memory issues": `nns_edge_release_handle()` signals the queue before it clears `eh->sending`, so a sender sitting between its `eh->sending` check and `nns_edge_cond_wait_until (q, 0)` misses the signal and `pthread_join()` never returns. It predates this branch and belongs with that item, not with item 10 - but the tests here add two more handle lifecycles, so they do widen the window a little. The `TIMEOUT` above at least turns it into a bounded failure.

myungjoo:
본 리뷰는 별도 컨텍스트의 리뷰 에이전트가 작성한 2차 리뷰를 옮긴 것입니다.

**Note:** This is a second-round review, produced by a separate automated review agent working in an isolated context, asked to (1) verify that the first-round review's findings were properly addressed and (2) re-review the whole PR against #259 item 10, with particular focus on the last two commits (`aa1f3c9`, `d2fc149`). It is transcribed here verbatim.

## Scope checked

Read the first-round review comment and the author's reply in full, then re-read all changed files at the current head (`d2fc149`): `nnstreamer-edge-data.c`, `nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-mqtt-paho.c`, `tests/CMakeLists.txt`, and the three test files. Built the branch in WSL (gcc, `-DENABLE_TEST=ON`) both plain and with `-fsanitize=address -fno-omit-frame-pointer`, and ran `ctest`: **100% pass in both modes** (`unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom`). Cross-checked issue #259 item 10's five bullets against the diff, traced `nns_edge_mqtt_connect`/`nns_edge_mqtt_close`/`nns_edge_metadata_create`/`nns_edge_metadata_destroy` implementations directly, and walked every write/read site of `eh->sending`/`eh->listening` across the file.

## Round-1 findings: verification of the fixes

- **Major #1 (unchecked `nns_edge_get_host_string()` in the hybrid branch of `nns_edge_start()`)** — fixed in `aa1f3c9`. Confirmed the added `if (!msg) { ...; ret = NNS_EDGE_ERROR_OUT_OF_MEMORY; goto done; }` lands on the function's existing `done:` label, which correctly sets `eh->is_started = (ret == NNS_EDGE_ERROR_NONE)` and unlocks — same pattern as every other error path in this function. Correct.
- **Minor #1 (stale `pthread_t` guard)** — fixed in `d2fc149`, but see the new finding below: the fix's own correctness now depends on `eh->sending`/`eh->listening` being reliably visible across threads, and that dependency is not fully safe (see "New finding A").
- **Minor #2 (global fault-injection window)** — author's reasoning (no assertion/early-return inside the armed window, and the only concurrent thread during `connectHostStringAllocFail_n` doesn't call the interposed functions) was independently re-checked against `_nns_edge_accept_socket()` and its callees and holds.
- **Nit #4 (no ctest timeout)** — fixed in `d2fc149`; `TIMEOUT 300` added to both registered tests in `tests/CMakeLists.txt`. Confirmed present and reasonable given the ~3.5s actual runtime.

All four round-1 points are genuinely addressed, not just acknowledged.

## New finding A (Major) — `eh->sending`/`eh->listening` are written unlocked at thread exit, but now read under lock to decide liveness

`d2fc149` changes `_nns_edge_create_send_thread()` / `_nns_edge_create_socket_listener()` (called with `eh` locked, per their own `@note`) to trust `eh->sending` / `eh->listening` instead of a non-zero `pthread_t` as proof the thread is alive:

```c
if (eh->send_thread) {
  if (eh->sending)
    return NNS_EDGE_ERROR_NONE;
  nns_edge_loge ("The sender thread has stopped, release the edge handle.");
  return NNS_EDGE_ERROR_IO;
}
```

Tracing every write site of these two fields:

- **Set to `true`**: inside `_nns_edge_send_thread()` / `_nns_edge_socket_listener_thread()`, at the top, *under* `nns_edge_lock(eh)` followed by `nns_edge_cond_signal(eh)` (`:868-871`, `:1191-1194`). The creator waits on `nns_edge_cond_wait(eh)`, so this transition is properly synchronized via the mutex — safe.
- **Set to `false` on natural thread exit** (the case this commit specifically targets — e.g. the listener breaking out of its loop on `POLLERR`/`POLLHUP`): `eh->sending = false;` at `:931` and `eh->listening = false;` at `:1217`, both **outside any lock**, right before the thread function returns.
- **Set to `false` by `nns_edge_release_handle()`**: under lock (`:1624`, `:1630`), but this is a different write path (used to signal the thread to stop, then `pthread_join()`s it and zeroes the `pthread_t`), not the one the new guard depends on.

The guard's read (`if (eh->sending)` / `if (eh->listening)`) happens under `eh`'s lock, but the corresponding "thread died naturally" write does **not** take that lock, so there is no happens-before edge between them — a plain data race per the C11/pthreads memory model. In principle this lets the guard observe a stale `true` after the owning thread has actually exited, which would make `_nns_edge_create_socket_listener()`/`_nns_edge_create_send_thread()` silently no-op and report success on a handle with no live thread — i.e., resurrect exactly the bug `d2fc149`'s commit message describes, through a narrower path (a race window instead of unconditional stale-`pthread_t` trust).

In practice this is a low-probability window (x86/TSO, and the compiler has little reason to cache the field across the surrounding `poll()`/`nns_edge_queue_wait_pop()` calls), and it would only be caught by ThreadSanitizer, not the AddressSanitizer job this repo's required `build` check runs — so a regression here would not be reliably caught by CI at all. It is also of a piece with the already-known, already-deferred item-9-class raciness in this file (the author's own reply cites the analogous `nns_edge_queue_clear()`/`_nns_edge_send_thread()` lost-wakeup as pre-existing and out of scope) — but this particular instance is new: before `d2fc149`, these two guard functions never relied on cross-thread visibility of `sending`/`listening` at all.

Suggested minimal fix that doesn't reintroduce the join-under-lock deadlock risk the author deliberately avoided: wrap just the final `eh->sending = false;` / `eh->listening = false;` writes (`:931`, `:1217`) in `nns_edge_lock(eh)`/`nns_edge_unlock(eh)` (a brief critical section at a point where the thread is about to return anyway — no join, no risk of the deadlock the commit message describes), or make both fields `_Atomic bool` with acquire/release semantics on the guard's read and the exit's write.

## New finding B (Major) — the actual behavioral change in `d2fc149` (dead-thread detection) has no regression test

`tests/unittest_nnstreamer-edge.cc`'s `edge.startTwice` (from `d3d2114`, unchanged by `d2fc149`) exercises `nns_edge_start()` twice while both threads are alive (`sending`/`listening == true`), and a third time after `nns_edge_stop()` — which, for a TCP handle, only flips `is_started` and does not touch the threads, so they are still alive there too. Every call in that test hits the `if (eh->sending) return NNS_EDGE_ERROR_NONE;` branch. The `else` branch added in `d2fc149` — the one that actually fixes the "stopped thread treated as running" bug the commit is about — is not exercised by any test in the suite (`grep` for `listening`/`sending`/`POLLERR`/`POLLHUP` in `unittest_nnstreamer-edge.cc` returns nothing).

This is a harder gap to close than the round-1 gaps: `nns_edge_handle_s` is defined inside `nnstreamer-edge-internal.c` itself (not a header), so the test binary has no white-box access to flip `eh->listening`/force a `POLLERR` the way it fault-injects `calloc()`/`nns_edge_get_host_string()` at the link level. Still, worth flagging squarely: per the PR's own stated bar (every other hard-to-reach path in this series — the hybrid mqtt leak, the two mqtt-callback leaks, and now this — is explicitly called out in a commit message or the PR description as "no test, because X"), `d2fc149`'s commit message justifies the `TIMEOUT 300` addition but says nothing about the dead-thread branch itself being untested. A regression that reverted this exact guard (e.g., back to trusting a bare `eh->listener_thread`) would pass every currently-registered test and would not be caught by the required `build`/DCO checks.

Recommend either (a) adding a test — even a narrow one that closes `eh->listener_fd` after `nns_edge_start()` to force `POLLERR` and then asserts the second `nns_edge_start()` fails, which likely does not need struct access since `nns_edge_start`'s socket-listener path could plausibly be observed black-box by making the listening socket unusable from outside — or, if that turns out impractical, (b) an explicit note (commit message or code comment) acknowledging the branch is untested and why, matching the PR's own standard elsewhere.

## Regression risk / module scope — no new concerns beyond A and B

Diff remains confined to the four `src/libnnstreamer-edge/*.c` files plus their three test files and `tests/CMakeLists.txt`; nothing outside `nns_edge_start()`/`nns_edge_data_clear_info()`/the two mqtt message callbacks/the two thread-creator guards is touched. No public header changed (confirmed via `git diff --name-only` against a `.h` filter), so no API/doc update is owed. Re-verified `_nns_edge_connect_to_broker()`'s ownership handling against both `nns_edge_mqtt_connect()`'s failure path (`*broker_h` is left untouched on error, and the helper already nulled `eh->broker_h` beforehand — no dangling pointer) and `nns_edge_mqtt_close()` (always frees `bh` and nulls fields regardless of the underlying disconnect's result) — consistent with round-1's conclusion that this path is leak-free.

## Test coverage / CI enforcement

The four regression tests from `d3d2114` (`edge.startTwice`, `edgeCustom.startTwice`, `edge.connectHostStringAllocFail_n`, `edgeData.clearInfoAllocFail_n`) still pass and, per round-1's independent revert-and-rerun verification, are load-bearing. `d2fc149`'s own `TIMEOUT 300` addition is a real, verified improvement (a regression that hangs a thread now fails a required check in 5 minutes instead of ctest's 1500s default). The gap is specifically New finding B above: the dead-thread-guard branch itself is not covered, so a regression limited to that one `else` branch would not be caught by CI.

## Code size vs. scope, docs — appropriate (unchanged from round 1)

The two new commits are 5 and 24 lines of source change respectively, tightly scoped to the exact functions round-1 flagged. No architecture/API doc update is needed; agrees with round-1's assessment.

## Summary of outstanding issues

| # | Severity | Issue |
|---|----------|-------|
| A | Major | `eh->sending`/`eh->listening` are written without the handle lock at natural thread exit (`nnstreamer-edge-internal.c:931`, `:1217`) but read under lock by the new `d2fc149` guards; this data race can in principle let the guard see a stale "still running" value after the thread has actually died, silently reproducing the bug the commit fixes. Not caught by ASan-based CI. |
| B | Major | The actual defect `d2fc149` fixes (dead thread treated as alive) has no regression test and is not disclosed as untested anywhere in the PR, unlike every other hard-to-reach path in this series. A regression here would pass CI. |
| — | — | All four round-1 findings (Major #1, Minor #1, Minor #2, Nit #4) are confirmed fixed/adequately addressed. |

## Overall

The five bullets of #259 item 10 are all correctly fixed, the fixes are well-isolated, and 100% of the registered suite passes under both plain and ASan builds. Round-1's findings were genuinely resolved rather than papered over. The two new findings above are both about the very code this round was asked to focus on (`aa1f3c9`/`d2fc149`) and are narrower/lower-probability than round-1's Major finding, but they share its shape: a correctness guarantee that the commit message claims ("fail the start when the thread is gone") is not fully backed by either the synchronization or the test suite. Given the PR's own demonstrated practice of closing exactly this class of gap when found, recommend addressing A (cheap: a lock around two field writes, or atomics) and at minimum disclosing B (ideally: a test) before merge.

**Verdict: Request changes** — not for the core item-10 fixes, which are sound and well-verified, but for finding A (a real, if narrow, correctness gap in `d2fc149`'s own thread-liveness check) and finding B (no regression coverage, and no disclosure, for that same check's primary behavior change).

myungjoo:
Both Major points land, and together they say the running-flag check should not be there at all. Reverted in `1fd51c0`, keeping the ctest timeouts that rode along with it.

**On the race (Major 1).** Correct, and the suggested minimal repair - clearing the flags under the handle lock on the way out - cannot be applied: `nns_edge_release_handle()` takes the handle lock and then joins both threads while still holding it (`nnstreamer-edge-internal.c:1621-1634`). A thread that tries to take that lock on its exit path blocks against the joiner, the joiner never returns, and the deadlock is certain rather than rare - a worse outcome than the stale read. That is the same reason I did not join the dead thread in the first place, and it is why there is no correct version of this check today: the only party that knows the thread is gone cannot say so under the lock that the reader holds.

`pthread_tryjoin_np()` would answer it race-free without blocking, but it is a glibc extension and this library is also built for Android (`jni/`), where bionic does not have it.

**On the missing test (Major 2).** Also correct, and it is the stronger argument: a branch that cannot be tested and cannot be made correct does not belong in a leak fix. What remains is the guard item 10 actually needs - a start that finds a thread handle already set does not create a second thread over it - and `edge.startTwice` and `edgeCustom.startTwice` do cover exactly that, verified against a tree with these sources reverted.

The case the reverted check aimed at - a listener that breaks out of its own loop on POLLERR/POLLHUP, leaving a started handle with no listener - is not a leak, predates this series, and has no safe answer until the connection list and these threads are locked properly. That is items 8 and 9 of #259, and those items are what will decide who joins a thread that stopped by itself.

**Separate, and worth recording here:** stress runs on this branch hung 2 times in 120 (one in `edgeCustom.createHandleAfterFailure`, which this pull request does not touch, and one in the new `connectHostStringAllocFail_n`), both at handle release; `main` was 0 in 120 so far, with a larger baseline still running. The reachable cause is the lost wakeup between `nns_edge_queue_clear()` and `_nns_edge_send_thread()` that #259 lists under "related non-memory issues" - reordering the two lines in the releaser does not close it, since the sender can pass its `eh->sending` check before it ever takes the queue lock. I will report the larger sample here; if it turns out this branch measurably widens that window rather than merely sampling it, I would rather fix the wakeup than ship the tests as they are.

myungjoo:
Following up on the flakiness I flagged in the previous comment - I was wrong about the cause, and the real one is worth writing down.

I instrumented `nns_edge_release_handle()` and reproduced the hang under markers. It stops in `_nns_edge_remove_all_connection()`, not at the sender join, and the mechanism is the *first* entry under "related non-memory issues" in #259: `_nns_edge_message_handler()` sets `conn->running = true` itself, after the thread is scheduled, while `_nns_edge_close_connection()` clears that flag and then joins. A close that lands before the thread reaches the assignment is simply undone by it - the handler sets the flag back to true, loops forever, and the join never returns. `edge.connectHostStringAllocFail_n` hits it because it releases the client while the server's connect-back is still being accepted, so the message thread is seconds old.

Fixed in `9cedb80` by setting the flag in `_nns_edge_create_message_thread()`, before the thread exists.

`ec701b9` fixes the lost wakeup I described earlier - the one I had wrongly blamed for this. It is real, and the custom binary did stop at the sender join twice in 400 runs, but it is a different entry on that same list. The queue gets a stopped state set by a new `nns_edge_queue_stop_wait()` and checked in `nns_edge_queue_wait_pop()` under the queue lock, so both interleavings end with the sender leaving its loop; `nns_edge_queue_clear()` keeps its meaning. Three unit tests cover it, including the interleaving that used to be lost.

Measured, 250 runs of each test binary per column, plain build:

| | before | after |
| --- | --- | --- |
| `unittest_nnstreamer-edge` | 3 hangs in 200 | 0 in 250 |
| `unittest_nnstreamer-edge-custom` | 1 hang in 200 | 0 in 250 |
| `main`, for reference | 0 in 200 each | - |

Both fixes are outside item 10 and I would normally leave them to their own item. They are here because this series' own tests hang on them: shipping a test that wedges the required `build` job about one run in a hundred is not something to hand off. Both commit messages say so, and neither touches the connection list or the locking that items 8 and 9 will rework.

myungjoo:
본 리뷰는 별도 컨텍스트의 리뷰 에이전트가 작성한 3차 리뷰를 옮긴 것입니다.

**Note:** This is a third-round review, produced by a separate automated review agent working in an isolated context, asked to (1) verify that round-2's two Major findings were resolved, (2) focus specifically on the two new commits `ec701b9` and `9cedb80` (which the author scoped in as "related non-memory issues" from #259 rather than item 10 proper, because this series' own tests hang on them), and (3) re-review the whole PR. It is transcribed here verbatim.

## Scope checked

Read both prior review rounds and every author reply in full. Re-read all changed files at the current head (`9cedb80`): `nnstreamer-edge-data.c`, `nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-mqtt-paho.c`, `nnstreamer-edge-queue.c`/`.h`, `tests/CMakeLists.txt`, and all three test files. Traced every write/read site of `eh->sending`/`eh->listening`, `conn->running`, and the new `q->stopped` flag, and every caller of `nns_edge_queue_wait_pop()`/`nns_edge_queue_clear()`/`nns_edge_queue_stop_wait()` and `_nns_edge_create_message_thread()`/`_nns_edge_close_connection()`. Built the branch in WSL (gcc, `-DENABLE_TEST=ON`) both plain and with `-fsanitize=address -fno-omit-frame-pointer`; `ctest` is 100% pass in both modes. Beyond the single ctest pass, ran the two binaries directly, outside ctest, with a correctly-composed `LD_LIBRARY_PATH` (ctest's own `ENVIRONMENT` property for the custom binary showed me this was needed): 40 consecutive runs of `unittest_nnstreamer-edge` and 40 of `unittest_nnstreamer-edge-custom`, both clean, 0 failures / 0 hangs. (An earlier ad-hoc stress loop of mine that skipped that `LD_LIBRARY_PATH` detail produced spurious `dlopen` failures on the custom binary and one contaminated log mixing in unrelated content from a stale process on the VM — a mistake in my own harness, not a PR bug; disregarded once the environment was fixed and re-run.)

## Round-2 outstanding findings: verification

Round-2's "Request changes" rested on two Major findings against `d2fc149` (the stale-`pthread_t` guard using `eh->sending`/`eh->listening` liveness). The author's reply (`1fd51c0`) reverts `d2fc149` in full rather than patching it, keeping only the `ctest` `TIMEOUT 300` that rode along with it, on the reasoning that the guard cannot be made correct without a join that would deadlock against `nns_edge_release_handle()`'s lock-held join, and that a branch that "cannot be tested and cannot be made correct does not belong in a leak fix."

Verified: `1fd51c0` removes both `if (eh->sending) ... else return NNS_EDGE_ERROR_IO;` / `if (eh->listening) ...` blocks and restores the plain `if (eh->send_thread) return NNS_EDGE_ERROR_NONE;` / `if (eh->listener_thread) return true;` guards from `c91f9e5`. This is exactly what item 10 needs (a start that finds a thread handle already set does not create a second one over it) and nothing more — the race Major-A flagged and the untested branch Major-B flagged are both gone because the code they lived in is gone, not patched around. **Both round-2 Majors: resolved.**

## Focused review of `ec701b9` — the send-queue lost wakeup

**Correctness of the fix itself.** `nns_edge_release_handle()`'s prior sequence was: `nns_edge_queue_clear()` (which only signals the condition variable) → `eh->sending = false` → join. A sender between its `eh->sending` check and the `nns_edge_lock(q)` inside `nns_edge_queue_wait_pop()` observes neither: the signal fired before it started waiting, and `eh->sending` was still `true` when it read it. With `timeout` of `0U` meaning infinite, `nns_edge_cond_wait_until()` then blocks forever and the join never returns. The fix adds a `stopped` flag to the queue itself, set under the *queue's own lock* by the new `nns_edge_queue_stop_wait()` and checked under that same lock in `nns_edge_queue_wait_pop()` (`if (q->length == 0U && !q->stopped)`). Because the set-and-check now share one lock, there is no window left: either the waiter takes the lock first and sees `stopped == false`, waits, and is woken by the same call's `nns_edge_cond_signal()`, or the stopper takes the lock first, and the waiter's later `nns_edge_lock()` call in `wait_pop()` serializes after it and sees `stopped == true` immediately. This is the right fix for a lost-wakeup — it moves the signal into the same critical section as the check it's racing, instead of trying to order two independent operations against an unlocked read. Confirmed against the two new interleaving tests (`edgeQueue.stopWaitBeforeWait`, `edgeQueue.stopWaitWhileWaiting`), and independently against 40 clean runs of the full `unittest_nnstreamer-edge` binary with correct environment.

**Sticky `stopped` flag — asked to verify explicitly.** `stopped` is never reset to `false` anywhere; once a queue is stopped it stays stopped for the rest of its life. I traced every call site of `nns_edge_queue_stop_wait()` (there is exactly one, `nnstreamer-edge-internal.c:1615`, in `nns_edge_release_handle()`) and every lifecycle point of `eh->send_queue` (`nns_edge_queue_create()` once in `nns_edge_create_handle()`, `nns_edge_queue_destroy()` once in `nns_edge_release_handle()`, immediately after the `stop_wait()`/join at lines 1612–1660). The send queue is created exactly once per handle and destroyed exactly once, right after being stopped; there is no `nns_edge_stop()`/`nns_edge_start()` cycle that reuses it (stop only touches `is_started` and, for TCP, does not even touch the threads — confirmed in the round-1 discussion and re-checked here). So "sticky" is fine here: the object marked stopped is never revived. I also checked what happens to a `push()` landing between `stop_wait()` and `destroy()` (a send from another thread racing the release): `nns_edge_queue_push()` doesn't consult `stopped` at all and will still enqueue, but `nns_edge_queue_destroy()` calls `nns_edge_queue_clear()` again before freeing the struct, so any such late-pushed data is still freed via its `destroy_cb` — no leak. **No correctness or reuse-path concern from the sticky flag as currently used.**

**Minor — `nns_edge_queue_clear()`'s doc comment is now slightly stale.** `nnstreamer-edge-queue.h:126`, unchanged by this PR: `@details When this function is called, nns_edge_queue_wait_pop() will stop the waiting.` That was exactly the assumption that turned out to be false under the race this commit fixes — `clear()` alone only *usually* stops the waiter. The call site that matters (`nns_edge_release_handle()`) is already correct because it now pairs `clear()` with the new `stop_wait()`, so this isn't a live bug. But the docstring, read on its own, still tells a future internal caller that `clear()` alone is sufficient to reliably wake a waiter, which is the same wrong assumption this commit had to spend a whole fix correcting. Worth a one-line update (e.g., pointing at `stop_wait()` for the "reliably wake and keep woken" guarantee) so the lesson isn't lost. `nnstreamer-edge-queue.h` is confirmed not installed (`packaging/nnstreamer-edge.spec`'s `%files devel` lists only `nnstreamer-edge.h`, `nnstreamer-edge-data.h`, `nnstreamer-edge-event.h`, `nnstreamer-edge-custom.h`), so this is purely an internal-maintainer-facing nit, not a public API doc gap.

**Minor — the same lost-wakeup shape is still open one queue over, in code this PR doesn't touch.** `nns_edge_mqtt_close()` (`nnstreamer-edge-mqtt-mosquitto.c:311`, `nnstreamer-edge-mqtt-paho.c` equivalent) tears down `bh->message_queue` via `nns_edge_queue_destroy()` alone — which internally calls `nns_edge_queue_clear()`, but never `nns_edge_queue_stop_wait()` — while `_mqtt_hybrid_direct_connection()` (`nnstreamer-edge-internal.c:1738`) can be blocked in `nns_edge_mqtt_get_message()` → `nns_edge_queue_wait_pop(..., 0U, ...)` (infinite timeout) on that exact queue, called from the per-connection message thread during the hybrid reconnect path. `nns_edge_release_handle()` joins that message thread (via `_nns_edge_remove_all_connection()`) *before* it calls `nns_edge_mqtt_close()`, so if that thread is waiting in this exact window when release happens, the join has the same shape of hang `ec701b9` just fixed for the send queue. This needs a live MQTT broker and the hybrid reconnect path to reach, so it's consistent with the rest of this PR's own disclosed mqtt-path gaps ("reaching them needs a live broker") and I'm not treating it as a blocker for this PR — but it's the identical bug pattern, fixed in one place and not the other, in code adjacent to what this PR is already touching (`nns_edge_mqtt_close()` is called from the very `_nns_edge_connect_to_broker()` helper `c91f9e5` added). Worth a follow-up, ideally tracked alongside the "related non-memory issues" list in #259.

## Focused review of `9cedb80` — the message-thread `running` race

**Correctness of the fix itself.** Previously, `_nns_edge_message_handler()` set `conn->running = true` as its first act *after being scheduled*, while `_nns_edge_close_connection()` sets `conn->running = false` and then joins. If the close lands before the new thread reaches that assignment, the assignment silently undoes the close, the thread loops forever on `while (conn->running)`, and `pthread_join()` never returns. The fix moves the assignment into `_nns_edge_create_message_thread()`, executed by the *creator*, synchronously before `pthread_create()` — so by the time any other thread can possibly observe or close this `conn`, `running` is already `true`, and there is no window left in which the thread's own first instruction could stomp on a close that already happened.

**Verified this closes the race rather than narrowing it**, by tracing both call sites of `_nns_edge_create_message_thread()`:
- `_nns_edge_connect_to()` (`internal.c:1042`): `conn` is a freshly-`calloc`'d local; `_nns_edge_create_message_thread()` is called before `_nns_edge_add_connection()` (`:1049`) and well before `conn_data->sink_conn = conn` is ever assigned, so no other thread has a path to reach this `conn` (and thus to `_nns_edge_close_connection()` it) until after thread creation returns.
- `_nns_edge_accept_socket()` (`internal.c:1151`, the `QUERY_CLIENT`/`QUERY_SERVER` branch): same shape — `conn_data->src_conn = conn` (`:1157`) happens strictly after `_nns_edge_create_message_thread()` returns.

So in both places, `conn->running = true` is set on an object no other thread can yet reach — there's no possible interleaving where a concurrent `_nns_edge_close_connection()` races the assignment. This is a genuine fix, not a narrower version of the same race.

**Failure path checked.** On `pthread_create()` failure, `_nns_edge_create_message_thread()` resets `conn->running = false` and `conn->msg_thread = 0` before returning `NNS_EDGE_ERROR_IO` — matches the commit message's "clear again only on the create failure it already handles," and both call sites already `goto error` → `_nns_edge_close_connection(conn)` on that return value, which is now a no-op join (`msg_thread == 0`, skips `pthread_join()`). No leak, no double-set.

**Minor — no deterministic regression test, but disclosed.** This is a narrow timing race between two real OS threads (schedule-and-join vs. thread-start), not something the existing fault-injection technique (interposed `calloc`) can force deterministically, and `nns_edge_handle_s`/`nns_edge_conn_s` are private structs with no test-visible hook to simulate it. The commit message discloses this candidly with concrete numbers (existing `edge.connectHostStringAllocFail_n` hit the hang 3/200 runs before the fix, 0/250 after, across both binaries combined with `ec701b9`) rather than silently shipping an untested behavioral change — consistent with how `aa1f3c9` disclosed its own untested hybrid-mqtt path. I re-ran the exposing test 40 times independently post-fix (see Scope Checked) with 0 failures, consistent with the author's numbers. Not a blocker, but flagging in the same spirit as round-1's Minor #3 (process-wide fault injection / flakiness risk): a regression limited to this exact ordering would most likely surface only as an intermittent required-check hang, not a deterministic CI failure.

## Is the scope call — fixing two "related non-memory issues" items inside an item-10 PR — justified?

Yes. Both bugs are demonstrated, not hypothetical: the author traced actual hangs of `nns_edge_release_handle()` produced by *this PR's own new tests* (`edgeCustom.startTwice`, `edge.connectHostStringAllocFail_n`), on stock `main` behavior these tests happen to exercise harder than `main`'s existing suite does (more handle create/release cycles, and a release racing an in-flight accept). Shipping a test suite that intermittently wedges the only required CI job (`build`, per this repo's branch protection — `DCO` and `build` are the sole required checks) is not an acceptable trade even though the underlying bugs predate this series and are explicitly filed elsewhere in #259. Both fixes are minimal (1-line relocation for `9cedb80`; a new opt-in `stopped` flag, additive to the queue's public surface, for `ec701b9`), neither touches the connection-list structure or locking that #259 items 8/9 are reserved for, and both are called out by name in their commit messages as belonging to that other list rather than to item 10. This is good scope hygiene, not scope creep.

## Rest of the diff — unchanged from round 1/round 2, re-confirmed

- `c91f9e5`, `97c4de5`, `aa1f3c9`, `7e38755`, `eedae57` are byte-identical to what round 2 already reviewed and approved of; no further changes since. Re-verified `_nns_edge_connect_to_broker()`'s close-before-reconnect ownership handling and the metadata-clear fix once more directly; both still hold.
- Diff remains confined to `src/libnnstreamer-edge/{nnstreamer-edge-data,internal,mqtt-mosquitto,mqtt-paho,queue}.{c,h}` plus their three test files and `tests/CMakeLists.txt`. No public header under `include/` or the installed `%files devel` set is touched (only the private `nnstreamer-edge-queue.h`, confirmed not installed above), so no ARCHITECTURE/API doc update is owed.
- Code size vs. scope: the two new commits are +75/-1 lines total across `ec701b9`+`9cedb80` combined (most of that is the queue's new function, its doc comment, and 3 new unit tests), tightly scoped to the exact two bugs their commit messages name.
- Test coverage / CI enforcement: the queue race now has 3 direct, deterministic unit tests exercising both interleavings (a real improvement — this primitive is independently testable, unlike the message-thread race). All previously-registered tests (`edge.startTwice`, `edgeCustom.startTwice`, `edge.connectHostStringAllocFail_n`, `edgeData.clearInfoAllocFail_n`) still pass. `TIMEOUT 300` from `d2fc149` survived the revert in `1fd51c0` and still bounds both binaries.

## Summary of outstanding issues

| # | Severity | Issue |
|---|----------|-------|
| 1 | Minor | `nns_edge_queue_clear()`'s doc comment (`nnstreamer-edge-queue.h:126`) still claims it alone reliably stops a waiter — the exact assumption `ec701b9` had to fix elsewhere. Internal header only; suggest a one-line update pointing at `nns_edge_queue_stop_wait()`. |
| 2 | Minor | The same lost-wakeup shape `ec701b9` fixes for the send queue is still open for the mqtt `message_queue` (`nns_edge_mqtt_close()` tears it down without `nns_edge_queue_stop_wait()`, while `_mqtt_hybrid_direct_connection()` can wait on it with an infinite timeout). Needs a live broker + hybrid reconnect to reach, so not a blocker here, but worth tracking with the rest of #259's "related non-memory issues." |
| 3 | Minor | `9cedb80`'s fix has no deterministic regression test — inherent to the race it closes, and disclosed with concrete before/after numbers in the commit message, consistent with this PR's established practice elsewhere. Non-blocking. |
| — | — | Both round-2 Majors (stale-`pthread_t`/data race, missing test for the reverted branch) are resolved by full revert in `1fd51c0`, not patched around. |

No Critical or blocking Major findings in this round.

## Overall

Item 10's five bullets remain correctly and minimally fixed, as established in rounds 1–2. This round's two new commits (`ec701b9`, `9cedb80`) fix real, demonstrated hangs that this PR's own tests exposed, are each narrowly and correctly scoped to the bug their commit message names, do not touch the connection-list locking reserved for #259 items 8/9, and are independently verified here by code tracing plus 40 clean runs each of both test binaries beyond the standard ctest pass (both plain and ASan). The three Minor points above are worth addressing (the doc comment is a one-line fix) but none of them block merge.

**Verdict: Approve.**

myungjoo:
Thanks. Taking the three Minors in turn.

**1. `nns_edge_queue_clear()`'s doc comment.** Fixed in `e5a8a7e`, in both the header and the source. You are right that it was the sharp edge: the comment asserted exactly the guarantee that `ec701b9` had to stop relying on, so anyone reaching for a shutdown would have reached for the wrong function. It now says it signals a waiter that is already blocked and points at `nns_edge_queue_stop_wait()`.

**2. The same shape on the mqtt `message_queue`.** Recorded, deliberately not fixed here. The one-line version - calling `nns_edge_queue_stop_wait()` before `nns_edge_queue_destroy()` in `nns_edge_mqtt_close()` - would look like a fix without being one. The waiter there is `_mqtt_hybrid_direct_connection()` inside the message thread, and the real problem is not only that its wait is missed: `nns_edge_mqtt_close()` destroys the queue mutex and frees the struct while that thread is still reacquiring it, and then `eh` is freed under it too. Waking it does not stop it from touching freed memory a moment later. That needs the thread joined before the queue dies, which is item 8 of #259 - it is that item's call to make, and `nns_edge_queue_stop_wait()` is now available for it.

**3. No deterministic test for `9cedb80`.** Agreed, and I could not construct one: the window is between `pthread_create()` returning and the new thread's first instruction, with nothing in the API to hold it open. What stands in for it is the before/after count in the commit message - 3 hangs in 200 runs, then 0 in 250 - and the fact that a regression puts the required `build` job into the `TIMEOUT 300` this series added, so it fails rather than hangs.

myungjoo:
Two things happened since the approval; both are worth recording.

**The Tizen GBS x86_64 job was hanging for six hours on every push of this branch, and I had not been reading it.** It is the only configuration that runs these tests outside ctest, so the `TIMEOUT 300` added here does not bound it, and the job died at the GitHub six hour limit each time. `edge.connectLocal` - a pre-existing test that runs before any test this branch adds - stopped dead; the same commit passed on ubuntu, plain and under AddressSanitizer, and `main` plus #260 run that test in 2.9 s.

The cause was the `calloc()` replacement in the unit test binary, the process-wide interposition raised as Minor 2 in the second review. It serves every allocation in the process, the C++ runtime and whatever the C library allocates through the PLT included, and it does not survive a change of environment. Removed in `5247e13`, together with `edgeData.clearInfoAllocFail_n`, the only test that used it. The metadata fix it covered now carries the same disclosure as the two mqtt leaks: fixed, not covered by a test.

The `nns_edge_get_host_string()` replacement stays, and the same GBS run vindicates the distinction - it is a symbol of the library under test rather than of the C library, and `connectHostStringAllocFail_n` passes there in 211 ms:

```
[ RUN      ] edge.connectLocal
[       OK ] edge.connectLocal (2891 ms)
[ RUN      ] edge.startTwice
[       OK ] edge.startTwice (10 ms)
[ RUN      ] edge.connectHostStringAllocFail_n
[       OK ] edge.connectHostStringAllocFail_n (211 ms)
```

189 tests, 3536 ms, on the runner that used to hang.

**Rebased onto main after #260 landed.** That merge brought its own copy of the clang-format commit this branch had cherry-picked, which put the pull request into conflict - and a conflicted pull request gets no `pull_request` workflow runs at all, which is why the previous push looked like CI had simply stopped. `git rebase` dropped the duplicate as already upstream and replayed the rest cleanly, so the 761 line formatting diff is out of this branch and what is left is twelve commits of actual change.

Since this removed a test and a piece of test machinery after the approval, it wants another look rather than my say-so.

myungjoo:
본 리뷰는 별도 컨텍스트의 리뷰 에이전트가 작성한 4차 리뷰를 옮긴 것입니다.

**Note:** This is a fourth-round review, produced by a separate automated review agent working in an isolated context, asked to (1) audit the post-approval `calloc()`-interposition removal and the resulting untested `nns_edge_data_clear_info()` fix, (2) judge whether the surviving `nns_edge_get_host_string()` interposition shares the same environmental fragility, (3) verify the rebase onto `main` (post-#260) introduced no duplication or loss, and (4) re-review the PR as a whole. It is transcribed here verbatim.

## Scope checked

Read all three prior review rounds and every author reply in full. Diffed the current head (`5247e13`) against `upstream/main` (`aaa2733`, confirmed to be the exact merge-base of this branch) and read every changed line. Read the standalone diff of `5247e13` itself. Grepped the whole `tests/` tree for `calloc`, `clearInfoAllocFail`, and related identifiers to look for leftovers. Traced all three call sites of `nns_edge_get_host_string()` in `src/`. Checked `gh pr checks` for the current run.

## Verification 1 — is the `calloc()` interposition removal clean?

Yes, functionally clean, with one cosmetic leftover.

- `git show 5247e13` is a pure deletion: the process-wide `calloc()` override (`nns_edge_test_calloc_fails` + the `extern "C" void *calloc(...)` definition) and `edgeData.clearInfoAllocFail_n` — the only test that used it — are both gone. No other test, helper, or `CMakeLists.txt` entry references either.
- `grep -rn "calloc" tests/` now only matches genuine, unrelated `calloc()` *calls* (`nnstreamer-edge-custom-test.c:87`, `unittest_nnstreamer-edge-mqtt.cc:36`, `unittest_nnstreamer-edge.cc:85`) plus one comment (next point). No second definition of `calloc` remains anywhere in the tree.
- `grep -rn "clearInfoAllocFail"` across the repo returns nothing — the test and every reference to it are gone.
- **Nit** — `tests/unittest_nnstreamer-edge.cc:30-32`, the doc comment directly above the surviving `nns_edge_get_host_string()` replacement, still reads: `@note This shadows the definition in the library for the whole process, the same way the calloc() replacement above does.` There is no `calloc()` replacement above any more; the sentence now refers to code that does not exist in this file. Purely cosmetic (it doesn't affect behavior or the other claims in the comment), but it is a genuine leftover of the removal and worth a one-line fix before or shortly after merge — e.g. drop the trailing clause, or say "the same way a libc-symbol interposition would" without pointing at a deleted example.

## Verification 2 — is leaving `nns_edge_data_clear_info()`'s fix untested acceptable?

Yes. Three reasons line up:

1. **The fix is small and structurally obvious.** The change (`src/libnnstreamer-edge/nnstreamer-edge-data.c:449-459`) is: null out `ed->metadata` immediately after destroying the old one, *before* attempting `nns_edge_metadata_create()`, instead of leaving the stale pointer in place while re-creating. This is the same "null the pointer before the operation that can fail" pattern used elsewhere in this series (e.g. `_nns_edge_connect_to_broker()` nulling `eh->broker_h` before reconnecting) and doesn't depend on any subtle timing — reading the 10-line function is enough to see it can no longer leave a dangling pointer.
2. **The success path is still covered.** `edgeData.clearInfo` (`tests/unittest_nnstreamer-edge.cc:1224`) still exercises the normal clear-and-recreate path; only the OOM-injection variant was removed.
3. **It matches the PR's own established disclosure practice.** The two mqtt leaks (queue-push failure, broker_h reconnect) were disclosed as untested from the first commit, for the same class of reason (unreachable without unsafe or unavailable fault injection). This is now a third, consistent instance, and the commit message says so explicitly ("keeps the same disclosure as the two mqtt leaks in this series: fixed, and not covered by a test") rather than silently dropping coverage.

The alternative — keeping the process-wide `calloc()` override just to hold onto this one test — was demonstrably worse: it hung a required CI job for six hours on every push. Trading a narrow, low-risk test gap for removing a fault injector that broke unrelated code (an unrelated pre-existing test, `edge.connectLocal`) is the right call.

## Verification 3 — does `nns_edge_get_host_string()` interposition share the same fragility?

No — it is a materially different, and safe, technique. Judged on the same axes that made `calloc()` dangerous:

| | `calloc()` (removed) | `nns_edge_get_host_string()` (kept) |
|---|---|---|
| Symbol owner | C library / CRT | This library only (`nnstreamer-edge-util.c`) |
| Callers in the process | Every allocation done through the PLT: the C++ runtime, gtest, pthread internals, ASan/libc allocator bootstrap, anything any linked library does | Exactly 3 call sites, all inside this library's own `.c` files, all traced above: `_nns_edge_connect_to()` (`internal.c:1021`), the hybrid branch of `nns_edge_start()` (`internal.c:1506`), and the paho URL builder (`mqtt-paho.c:158`) |
| When it can fire | Continuously, from process start (static initializers, gtest's own setup) through teardown, on any thread | Only synchronously inside a `nns_edge_connect()`/`nns_edge_start()` call, on the calling thread |
| Demonstrated failure mode | Broke `edge.connectLocal`, a test that has nothing to do with allocation failure, on one specific runner (Tizen GBS x86_64) because the environment there made some incidental allocation during that test's execution fail | None observed on any CI configuration, including the same GBS runner that broke on `calloc()` |
| Independent verification in this round | — | The author's own last comment shows `connectHostStringAllocFail_n` passing on the previously-hanging GBS runner in 211 ms, in the same run where `edge.connectLocal` also passed in 2.9 s — i.e., the fix was validated in exactly the environment that exposed the problem with the other interposition |

The narrow-window argument the author made in round 1/round 2 (the flag is armed only around one call, and the only concurrent thread during the test — the server's listener/accept thread — never calls `nns_edge_get_host_string()`) was independently reconfirmed in round 2 and still holds; nothing in this round's diff touches that logic. Given the difference in blast radius (3 known, traced, single-threaded call sites in this library vs. every allocation in the process) and the fact that this exact technique already survived the CI environment that killed the other one, **this interposition is fine to keep as-is.** No action needed here.

## Verification 4 — is the rebased diff clean against `upstream/main`?

Yes, confirmed on all three points raised in scope:

- `git merge-base upstream/main HEAD` equals `git rev-parse upstream/main` exactly (`aaa2733`) — the branch is rebased onto the current tip of `main`, not a stale ancestor.
- `git log --oneline upstream/main..HEAD` shows exactly 12 commits, none of them a clang-format-only commit — the PR description's claim that the cherry-picked clang-format commit from #260 "rebases cleanly onto it" and gets auto-dropped as already-upstream is verified: it is simply absent from this branch's commit list now.
- `git diff upstream/main...HEAD --stat` totals **10 files changed, 325 insertions(+), 16 deletions(-)**, with `tests/unittest_nnstreamer-edge.cc` contributing 168 of those lines as genuine new test bodies (`startTwice`, `connectHostStringAllocFail_n`, `stopWaitBeforeWait`, `stopWaitWhileWaiting`, `stopWaitInvalidParam01_n`, plus the `_get_open_fd_count()`/`nns_edge_get_host_string()` test helpers) — not the ~761-line formatting diff that round 1 had to reconcile against #260. No duplication, no loss.
- No file under `include/` is touched (`git diff --name-only upstream/main...HEAD | grep '^include/'` is empty), consistent with every prior round's conclusion that no public API/doc update is owed.
- `gh pr checks 263` is currently green on every required check: `DCO`, `build` (Ubuntu, plain), and — the job that mattered most this round — **`Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` now passes in 3m16s**, where it previously hung for six hours. `build (-A armv7l)` / `build (-A aarch64)` are still pending at the time of this review but are not part of the required-check set per this repo's branch protection (only `DCO` and `build` are required).

## Re-review of the whole PR

Nothing here contradicts rounds 1-3's conclusions; restating the load-bearing points with this round's evidence folded in:

- **Does it solve #259 item 10?** Yes. All five bullets (the four "smaller leaks" plus the adjacent NULL-deref) are fixed: the `is_started`/thread-handle guards in `nns_edge_start()`, the mqtt queue-push-failure frees, the `nns_edge_data_clear_info()` dangling-pointer fix, and the `nns_edge_get_host_string()` NULL checks on both the TCP connect path and the hybrid start path. Two additional, out-of-scope-but-necessary races (`_nns_edge_message_handler()`'s `running` flag, the send-queue lost wakeup) were found, fixed, and explicitly disclosed as belonging to #259's "related non-memory issues" rather than item 10 — this scoping was already judged justified in round 3 and nothing since changes that.
- **Regression risk.** Low. The diff is confined to the four `src/libnnstreamer-edge/*.c` files item 10 targets plus the queue primitive the disclosed race fix needed, and their test files. No connection-list structure or locking (reserved for #259 items 8/9) is touched. The one behavior actually removed since round 3 (`calloc()` interposition + its one test) is a test-harness change, not a production-code regression — the production fix it used to cover is unchanged.
- **Code size vs. scope.** Proportionate: +325/-16 across 10 files, the bulk of it new test bodies, tightly scoped to the named bugs.
- **Test completeness.** 8 new/changed regression tests remain (`edge.startTwice`, `edgeCustom.startTwice`, `edge.connectHostStringAllocFail_n`, three `edgeQueue.stopWait*` tests, plus the message-thread race's disclosed-but-untested status). Three genuinely hard-to-reach paths (both mqtt leaks, and now the metadata-clear OOM path) are consistently and explicitly disclosed as untested rather than silently uncovered — a real, repo-specific practice this PR keeps to.
- **Does CI actually gate a regression?** Yes, and more robustly than at round 1: the required `build` job runs the suite under both plain and ASan `ctest` with a `TIMEOUT 300` (added mid-review, survived the later revert), and the Tizen GBS job — previously an unbounded, un-ctest'd blind spot that this very PR's own test-harness choice had broken — now passes cleanly and exercises the same tests directly.
- **Documentation impact.** None owed. No public header under `include/` changed; the one header that did change (`nnstreamer-edge-queue.h`) is confirmed (round 3) not part of the installed `%files devel` set, and its doc comments were already brought in line with the new `nns_edge_queue_stop_wait()`/`nns_edge_queue_clear()` split in `67a22fe`.

## Summary of outstanding issues

| # | Severity | Issue |
|---|----------|-------|
| 1 | Nit | `tests/unittest_nnstreamer-edge.cc:32`'s doc comment for the surviving `nns_edge_get_host_string()` override still says "the same way the calloc() replacement above does" — that replacement no longer exists in this file after `5247e13`. Cosmetic only; suggest a one-line follow-up fix. |
| — | — | Everything else raised in rounds 1-3 remains resolved; no new Critical/Major/Minor findings this round. |

No Critical, Major, or Minor findings in this round — only the one cosmetic Nit above.

## Overall

The post-approval changes are exactly what they claim to be: a clean removal of a process-wide fault injector that was breaking an unrelated, pre-existing test on the one CI configuration that runs these binaries outside `ctest`, with the one test that depended on it removed alongside it and its underlying fix's now-untested status disclosed consistently with how this PR has handled every other hard-to-reach path. The surviving `nns_edge_get_host_string()` interposition is a narrower, traceable, project-symbol-only technique that does not share the failure mode that forced `calloc()`'s removal, and this round's own CI run (the previously-hanging GBS job, now green) is direct evidence of that distinction rather than just an argument for it. The rebase onto the post-#260 `main` is clean: merge-base matches `upstream/main`'s tip exactly, the duplicated clang-format commit is gone as expected, and the remaining diff is proportionate to and fully explained by item 10's five bullets plus the two disclosed, necessary race fixes.

**Verdict: Approve.**

myungjoo:
Pointer from #264, since this PR is where the queue waiters now live.

`nns_edge_cond_wait_until (h, ms)` builds `ts.tv_nsec = now.tv_usec * 1000 + (ms % 1000) * 1000000` with no normalisation, so for any timeout that is not a multiple of 1000 the field exceeds 10^9 for part of every second - about one call in ten at 100 ms - and `pthread_cond_timedwait()` rejects it with `EINVAL` and does not wait at all. `nns_edge_queue_wait_pop()` with a finite timeout goes straight through it.

#264 carried the one-line carry into `tv_sec` for a while and took it back out: correct in isolation, it made nnstreamer's `query` case 9-3 stop exiting and get killed at its ten second timeout, bisected over ten CI runs on that branch while `main` and the sibling branches passed the same job. The detail and the suspected interaction - `_nns_edge_clear_retained()` in the mosquitto backend waits up to 1000 x 10 ms on a condition its own publish callback signals from the network thread while taking the same lock - are on nnstreamer/nnstreamer-edge#259.

So it needs the deadline fixed **and** that loop understood together. Flagging it here rather than re-adding it there; the test that pins it (`waitPopTimedoutSecondBoundary`) is in #264's history if it is useful.

PR #265, opened 2026-09-07
[Fix] drop the rejected connection from the connection table

myungjoo:
Addresses item 7 of #259: a dangling `src_conn` / `sink_conn` after a failed accept leads to a double free.

## The bug

`_nns_edge_accept_socket()` publishes the freshly accepted connection into the connection table before it asks the application about it:

```c
    _nns_edge_close_connection (conn_data->src_conn);
    conn_data->src_conn = conn;          /* or sink_conn */
  ...
  if (ret != NNS_EDGE_ERROR_NONE)
    goto error;
  ...
error:
  if (!done)
    _nns_edge_close_connection (conn);   /* frees conn, table still points at it */
```

If the event callback rejects the connection, `conn` is freed while `conn_data` keeps the pointer. The next accept for the same client id closes the stale pointer, and `nns_edge_disconnect()` / `nns_edge_release_handle()` free it again through `_nns_edge_release_connection_data()`.

## The fix

Clear the table entry before releasing the connection.

Clearing it leaves a connection data with no sink connection in the table until the next accept replaces it, and `_nns_edge_send_thread()` walks the whole table when the data carries no client id. `_nns_edge_transfer_data()` already dereferenced the connection in its error log, so it now rejects a null connection up front; the sender then removes that entry as it does for any other transfer failure.

## Tests

Four tests, all driven with a raw socket - that is all the accept path needs for a publisher or a query client.

| test | what it pins down |
| --- | --- |
| `rejectPubConnection_n` | rejected connection stored in `sink_conn` |
| `rejectQueryConnection_n` | rejected connection stored in `src_conn` |
| `acceptIncompleteConnection_n` | error exit taken before the connection data exists |
| `sendAfterRejectedConnection` | sender thread walking a table that holds a rejected entry |

Verified by rebuilding with only part of the change reverted, on `main` @ f25909e:

| tree | result |
| --- | --- |
| this branch | 193/193 pass, clean under ASan, 10 repeats stable |
| `main` sources + these tests | `heap-use-after-free` in `_nns_edge_close_connection` for the three reproducer tests |
| accept fix only, no null guard | `sendAfterRejectedConnection` fails with `SEGV` |

Every added line and every added branch in `_nns_edge_accept_socket()` is executed by the suite (gcov).

## Not in scope

Item 9 of #259, the missing handle lock around the connection list, is still open and left for its own change.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -488,6 +488,11 @@ _nns_edge_transfer_data (nns_edge_conn_s * conn, nns_edge_data_h data_h,
   unsigned int i;
   int ret;
 
+  if (!conn) {
+    nns_edge_loge ("Failed to transfer data, edge connection is null.");
+    return NNS_EDGE_ERROR_INVALID_PARAMETER;
+  }
+
   _nns_edge_cmd_init (&cmd, _NNS_EDGE_CMD_TRANSFER_DATA, client_id);
 
   ret = nns_edge_data_get_count (data_h, &cmd.info.num);
@@ -1071,7 +1076,7 @@ _nns_edge_accept_socket (nns_edge_handle_s * eh)
 {
   bool done = false;
   nns_edge_conn_s *conn;
-  nns_edge_conn_data_s *conn_data;
+  nns_edge_conn_data_s *conn_data = NULL;
   nns_edge_cmd_s cmd;
   int64_t client_id;
   char *dest_host = NULL;
@@ -1169,8 +1174,17 @@ _nns_edge_accept_socket (nns_edge_handle_s * eh)
   done = true;
 
 error:
-  if (!done)
+  if (!done) {
+    /** Detach the connection before releasing it, to avoid a double free. */
+    if (conn_data) {
+      if (conn_data->src_conn == conn)
+        conn_data->src_conn = NULL;
+      if (conn_data->sink_conn == conn)
+        conn_data->sink_conn = NULL;
+    }
+
     _nns_edge_close_connection (conn);
+  }
 
   SAFE_FREE (dest_host);
 }
--- a/tests/unittest_nnstreamer-edge.cc
+++ b/tests/unittest_nnstreamer-edge.cc
@@ -8,10 +8,13 @@
  */
 
 #include <gtest/gtest.h>
+#include <arpa/inet.h>
 #include <dirent.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/socket.h>
 #include <thread>
+#include <unistd.h>
 #include "nnstreamer-edge-data.h"
 #include "nnstreamer-edge-event.h"
 #include "nnstreamer-edge-log.h"
@@ -72,7 +75,9 @@ typedef struct {
   bool running;
   bool is_server;
   bool event_cb_released;
+  bool reject_connection;
   unsigned int received;
+  unsigned int connection_completed;
 } ne_test_data_s;
 
 /**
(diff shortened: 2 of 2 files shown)


myungjoo:
**Disclosure: this review was produced by a separate automated review agent (not the PR's authoring agent). It is being transcribed verbatim into this comment.**

Scope reviewed: the diff `upstream/main..HEAD` for this branch (`src/libnnstreamer-edge/nnstreamer-edge-internal.c` and `tests/unittest_nnstreamer-edge.cc`), cross-checked against issue #259 item 7 and the surrounding connection-lifecycle code (`_nns_edge_connect_to`, `_nns_edge_send_thread`, `_nns_edge_close_connection`, `_nns_edge_remove_connection`, `_nns_edge_release_connection_data`) and the CI workflows that gate merges.

## Summary
No blocker, major, or minor issues found in the code changes. This is a correct, minimally-scoped fix for the reported double-free, backed by targeted regression tests that are wired into the required CI gate. A couple of nit-level observations below, none of which need to block merge.

---

### 1. Does the fix address item 7? — Verified, no issue
`_nns_edge_accept_socket()` now nulls out `conn_data->src_conn` / `conn_data->sink_conn` on the error path *before* calling `_nns_edge_close_connection(conn)`, but only when the table entry actually still points at the connection being freed (`if (conn_data->src_conn == conn)` / `if (conn_data->sink_conn == conn)`). I traced all `goto error` paths in the function:
- Failures before `conn_data` is assigned (e.g. `accept()` failure, capability/host-info exchange failure) → `conn_data` is NULL (now explicitly initialized as such), the new block is a no-op, behavior unchanged from before — correct, since nothing was published to the table yet.
- Failure of `_nns_edge_create_message_thread()` for the QUERY_CLIENT/QUERY_SERVER branch → this happens *before* `conn_data->src_conn = conn` is assigned, so the guard correctly does not clear a still-valid, unrelated old connection.
- Failure of `nns_edge_event_invoke_callback(..., CONNECTION_COMPLETED, ...)` after `src_conn`/`sink_conn` was just set to `conn` → this is exactly the bug scenario from #259 item 7; the guard now clears the table entry, and `_nns_edge_close_connection()` frees `conn` exactly once. Confirmed by reading `_nns_edge_release_connection_data()`, which later calls `_nns_edge_close_connection()` on whatever is in `src_conn`/`sink_conn` — since it's now NULL, that call is a no-op (the function itself null-checks), so no double free on later `nns_edge_disconnect()`/`nns_edge_release_handle()`.

The companion `_nns_edge_transfer_data()` null-guard is also correct and necessary: clearing the table entry means `conn_data->sink_conn` can legitimately be NULL until the next accept, and `_nns_edge_send_thread()`'s broadcast path (`nns_edge_conn_data_s *conn_data = eh->connections; ... conn = conn_data->sink_conn;`) walks the whole table unconditionally. Without the guard, `_nns_edge_cmd_send(conn, ...)` inside `_nns_edge_transfer_data()` would dereference a NULL `conn`. This is a real, newly-introduced-by-this-fix code path, and it is correctly handled.

I also checked the one other place with the same "store connection in table, then do more fallible work" shape, `_nns_edge_connect_to()` (client-side connect path, lines ~1046-1052 pre-PR numbering). There, `conn_data->sink_conn = conn;` is immediately followed by `done = true;` with no further fallible step in between, so that path was never vulnerable to this bug and correctly was left untouched.

### 2. Regression risk to other modules — none found (nit only)
The functional diff touches exactly two lines' worth of logic in one function (`_nns_edge_accept_socket`) plus a 5-line defensive guard in `_nns_edge_transfer_data`. No other module, node type, or connect type is touched. `_nns_edge_close_connection()` already correctly stops/joins the per-connection message thread before freeing (`conn->running = false; pthread_join(...)`), so freeing a connection whose message thread was already started (the QUERY_CLIENT/QUERY_SERVER branch) was already race-free before this PR and remains so.

- Nit: the pre-existing behavior in `_nns_edge_send_thread()`'s broadcast loop of calling `_nns_edge_remove_connection()` (which releases *both* `src_conn` and `sink_conn` of that client id) whenever `_nns_edge_transfer_data()` returns non-`NNS_EDGE_ERROR_NONE` is unchanged by this PR, but is now also reachable via the new NULL-`conn` early return. This is pre-existing logic outside this diff's scope, not a defect introduced here — flagging only as an FYI, not requesting a change.

### 3. Code size / scope — appropriate
18 lines changed in the fix commit's `.c` file; the third commit adds only tests. The first commit (`61fcbd5`, clang-format of `tests/unittest_nnstreamer-edge.cc`) is purely mechanical — verified with `git diff -w` that beyond the clang-format commit's `#include` reordering and brace/line-wrap style changes, there is no textual/logic change. This prep commit is required by `static.check.yml`'s formatting gate before the test file can be touched, and the PR description transparently notes it's shared with the item-1 PR and one will become empty once the other lands — reasonable and not scope creep.

### 4. Test coverage — good, verified by tracing each test against the code paths
Four new tests, each mapped to a specific branch:
- `rejectPubConnection_n` → exercises the `sink_conn` branch (PUB node type, "else" branch of the node-type check in `_nns_edge_accept_socket`).
- `rejectQueryConnection_n` → exercises the `src_conn` branch (QUERY_CLIENT, which also runs its own listener/accept path).
- `acceptIncompleteConnection_n` → exercises the `conn_data == NULL` early-error path (QUERY_SERVER, peer disconnects before sending `HOST_INFO`, so `_nns_edge_add_connection()` is never reached).
- `sendAfterRejectedConnection` → exercises the new `_nns_edge_transfer_data()` NULL-guard by rejecting one connection, accepting a second, then sending data so `_nns_edge_send_thread()`'s broadcast loop walks a table entry whose `sink_conn` is NULL.

All four tests use a raw POSIX socket rather than the library's own client, which is the right choice since only the accept path is under test. The polling-with-`usleep`/retry-counter pattern used to synchronize with the listener thread (`connection_completed`) is consistent with the existing style already used for `_td->received` elsewhere in this file, not a new anti-pattern.

The PR description states the tests were verified to fail with `heap-use-after-free` under ASan against `main` sources, and that reverting only the null-guard produces a `SEGV` in `sendAfterRejectedConnection` — i.e. each test is shown to actually pin down the specific defect it targets, not just exercise the code path.

### 5/6. Future-regression detection and CI gating — verified
- `tests/CMakeLists.txt` registers `unittest_nnstreamer-edge` via plain `ADD_TEST()` with no filter excluding `_n`-suffixed tests, so all four new tests run as part of the normal gtest binary.
- `.github/workflows/ubuntu_clean_cmake_build.yml` (the "CMake" / `build` job) runs `ctest -C ... --output-on-failure --no-tests=error` with no `continue-on-error`, both in a plain build and again under `-fsanitize=address` (the ASan run is explicitly placed in this same required job specifically so a memory-safety regression blocks merge, per the comment in the workflow file). A future change that reintroduces a dangling `src_conn`/`sink_conn`, or that regresses the new NULL-guard, would fail this required check and block merge — satisfying the "future regression must be caught by CI" requirement.

### 7. Architecture/API/doc impact — none, correctly no doc changes
`nns_edge_conn_data_s` and every function touched are file-local `static` internals of `nnstreamer-edge-internal.c`; there is no public API signature change and no documented invariant elsewhere in the tree references these fields. No architecture or public API documentation update is needed, and none was expected.

---

**Verdict: approve.** No blocker/major/minor defects found in the reviewed diff. The two notes above are nit-level/FYI only and do not require changes before merge.

myungjoo:
Rebased onto `main` @ aaa2733. The clang-format commit on `tests/unittest_nnstreamer-edge.cc` dropped out because #260 landed it, so this PR is now just the fix (18 lines) plus the tests (244 lines). Rebuilt and re-ran the suite on the new base: 188/188 pass, clean under AddressSanitizer, and the same three-tree ablation still holds.

myungjoo:
Rebased again onto `main` @ f25909e, which resolves the merge conflict. The only conflict was the `#include` block at the top of `tests/unittest_nnstreamer-edge.cc`, where both sides had added headers; both sets are kept and the file is still clang-format clean.

Re-verified on the new base, which now carries the other connection-lifecycle fixes (#259 items 8/9/10 and the `conn->running` handover):

| tree | result |
| --- | --- |
| this branch | 193/193 pass, clean under ASan, 10 repeats stable |
| `main` @ f25909e sources + these tests | `rejectPubConnection_n`, `rejectQueryConnection_n` and `sendAfterRejectedConnection` all fail with `heap-use-after-free` |
| accept fix only, no null guard | `sendAfterRejectedConnection` fails with `SEGV` |

So the double free is still present on current `main` and is not covered by any of the fixes that landed in the meantime.

PR #264, opened 2026-09-07
[Fix] do not release the connection in its own message thread

myungjoo:
Addresses item 8 of #259: *"Message-handler thread removes its own connection: self-join, thread leak, UAF after release"*.

## The bug

`_nns_edge_message_handler()` called `_nns_edge_remove_connection()` when the peer went away. That path frees the `nns_edge_conn_s` the thread itself is running on and calls `pthread_join()` on the calling thread, which returns `EDEADLK` on glibc:

* the message thread was never joined nor detached, so its stack and descriptor leaked once per disconnected peer;
* the connection was already unlinked from `eh->connections`, so `_nns_edge_remove_all_connection()` in `nns_edge_release_handle()` could not see it either - the thread outlived the handle and dereferenced it after it was freed;
* for `NNS_EDGE_CONNECT_TYPE_HYBRID` the thread then waited forever in `nns_edge_mqtt_get_message()` with an infinite timeout, and `nns_edge_mqtt_close()` destroyed that queue underneath the waiter.

## The fix

Give the connection a single owner.

* `_nns_edge_close_connection()` refuses to join and free the connection of the calling thread. It closes the socket and returns false instead.
* A connection data leaving the live list is kept in `eh->closed_connections` until the socket listener after an accept, `nns_edge_connect()`, `nns_edge_disconnect()` or `nns_edge_release_handle()` joins its message thread and releases it. Its sockets are closed at once when no message thread of that connection data can be reading them, and after the join otherwise, and its message threads are asked to stop; a connection dropped while its message thread runs therefore keeps both its descriptor and the error command to the peer until the next release. The join graph stays acyclic: a message thread never joins anything, and the threads that release parked connections are never joined by one. Because the live list keeps its previous contents, connection lookup, data transfer and `nns_edge_is_connected()` behave as they did.
* That list has its own mutex, `eh->closed_lock`, because it is touched by the listener thread and by a message thread without the handle lock. Entries are split under the mutex so that an entry belongs to exactly one reaper and the entry of the calling thread stays visible to whoever releases the handle; the mutex is never held across `pthread_join()` and never taken while the handle lock is being acquired.
* `_nns_edge_remove_all_connection()` clears `conn->running` for every connection before releasing it, so an owner-initiated teardown stops a message thread even when that thread is the caller.
[...]

PR #267, opened 2026-09-07
[Fix] bound the bytes a peer can make the receiver allocate

myungjoo:
Fixes item 5 of #259.

## The bug

`_nns_edge_cmd_receive()` allocates `cmd->info.mem_size[n]` for every memory the peer announces, plus `cmd->info.meta_size`, straight out of the command header and before a single payload byte has arrived. Nothing bounded those values — `nns_edge_malloc()` only rejects sizes above `SIZE_MAX`, which on a 64-bit host is no rejection at all. Announcing a size costs the peer the 8 bytes it occupies in the header, and the peer never has to send the payload it promised.

Three unauthenticated paths reach it: the listener taking `HOST_INFO` in `_nns_edge_accept_socket()`, the connecting node taking `CAPABILITY` in `_nns_edge_connect_to()`, and every message thread taking `TRANSFER_DATA` in `_nns_edge_message_handler()`.

The MQTT message queue of a broker handle had the same shape of problem: `nns_edge_queue_create()` leaves `max_data` at zero, so a subscriber that is not draining it grows for as long as the broker keeps publishing.

## The fix

The announced sizes are summed against a per-connection limit before anything is allocated, and the whole command is rejected when they do not fit. The running total never exceeds the limit, so the remaining budget used as the bound cannot underflow and the sum cannot wrap — a peer cannot pick sizes that overflow the accumulator back into an acceptable value.

The limit defaults to **256 MiB for the whole command**, an order of magnitude above the largest frame a realistic tensor or video stream carries. It is configurable through a new `MAX_TRANSFER_SIZE` edge info key, which also takes zero to restore the previous unbounded behavior. A connection reads the value in force when it is created, so it has to be set before `nns_edge_start()` or `nns_edge_connect()`; that is documented in the info table in `nnstreamer-edge.h`.

The broker message queue is capped at 100 messages with the oldest dropped. Dropping rather than refusing bounds the queue and keeps the newest message, which is what a live stream wants, and it also keeps `nns_edge_queue_push()`'s failure path — which leaks the message, item 10 of the same issue — out of reach of a slow consumer.

No API or ABI change: `nns_edge_conn_s` and `nns_edge_handle_s` are file-local to `nnstreamer-edge-internal.c`, and the exported symbol set of `libnnstreamer-edge.so` is byte-identical to `main`'s.

**Worth a release note, two defaults changed.**

`MAX_TRANSFER_SIZE` defaults to 256 MiB, which is a behavior change for any consumer that really does move more than that in one transfer. Such a consumer now has to set the key explicitly, and set it before the connection is created. I could not find a downstream that comes close from this repository alone; a maintainer's read on `edgesrc`/`edgesink` usage would be welcome.

`RECV_TIMEOUT` defaults to 10 s, where the previous behavior was an unbounded wait. It bounds silence inside a receive that has already begun, so it does **not** fire while a node is merely thinking before it answers: the message handler only calls the receive after `poll()` reports data, and a peer that keeps sending resets the wait with every byte. The two places it can bite are the handshake reads that are not behind a `poll()` — the capability the connecting node waits for in `_nns_edge_connect_to()`, and the host info a query server waits for in `_nns_edge_accept_socket()`. Both are sent immediately by the peer today, so 10 s of total silence there already means something is wrong, but a deployment that puts something slow in front of the handshake should raise the key or set it to zero.

## Why working applications are unaffected

This was the main risk to check, so it is what most of the test suite is for. Seven of the fifteen new tests describe behavior the limit must **not** change, and every one of them passes against the library **without** this fix as well as with it — they were run against a tree with `nnstreamer-edge-internal.c` reverted to `main` and the tests kept:

| test | what it pins |
| --- | --- |
| `capabilityWithinLimit` | the handshake still completes and answers with `HOST_INFO` |
| `dataWithinLimit` | a three-memory transfer with metadata arrives byte for byte |
| `defaultLimitAcceptsLargeData` | a 4 MiB payload passes the default limit untouched |
| `dataAtLimit` | a transfer whose total is *exactly* the limit is accepted |
| `dataUnlimited` | `MAX_TRANSFER_SIZE=0` restores the unbounded behavior |
| `sendWithinReceiverLimit` | an end-to-end send between two real edge handles still arrives |
| `setInfoMaxTransferSize` | the key takes the values an application would write, and does not disturb the metadata an application keeps on the handle |

If a later change makes the limit reject ordinary traffic, these fail.

## Tests that catch the bug

The command header carries the sizes, and no public API lets a test put a chosen header on the wire, so the tests stand up a listening socket that speaks the protocol by hand and let a real edge handle connect to it. That is the only way to make the announced size and the payload behind it disagree, which is exactly what a hostile node does.

Against the reverted library, six of the eight refusal tests fail: `capabilityOverLimit_n`, `dataOverLimit_n` (one byte over, which pins the boundary against `dataAtLimit`), `dataMetaOverLimit_n`, `dataSumOverLimit_n` (four memories each under the limit, summing past it), `setInfoKeepsLimitOnFailure` (a rejected info value left the limit alone) and `sendOverReceiverLimit_n` (through the public API only), plus `setInfoMaxTransferSize_n` for the parser. Each refusal test closes the peer socket before releasing the handle, so a receiver still waiting for a payload it should have refused reports a failed assertion in a few seconds instead of blocking the suite.

`dataSizeOverflow_n` is the honest exception: the sizes it picks were already refused before this fix, by the allocation failing, so it guards the arithmetic of the new check rather than the fix itself. Sizes that both wrap a 64-bit accumulator and remain individually allocatable do not exist on a 64-bit host, so no test can distinguish there.

`plain` and `asan` builds both pass 100% locally, and the required `build` check already runs `unittest_nnstreamer-edge` under AddressSanitizer, so a regression in the `edgeTransfer` suite fails a required check.

## Where the MQTT queue test actually runs

The first answer was: nowhere. `unittest_nnstreamer-edge-mqtt.cc` is not registered with `ADD_TEST()`, and the two workflows that start a broker (`daily-build-gbs.yml`, `nns_edge_testhub.yml`) trigger only on `schedule`/`workflow_dispatch`. The one required check that executes the binary, `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`, has no step that starts a broker either, so `_check_mqtt_broker()` returns false and every test in the file returns before asserting anything. This PR's own GBS log shows `messageQueueLimit` finishing in **11 ms**, against 2104 ms when it really runs — that is what a silent skip looks like.

The last commit fixes that. The required `Build and test on Ubuntu 26.04 (gcc 15)` job already configures with `-DMQTT_SUPPORT=ON` and so already builds the binary; it only lacked a broker. It now installs and starts mosquitto and runs the binary after `ctest`.

This was validated before switching it on rather than after. mosquitto 2.0.18 was built from source locally and **all twenty tests in that file pass against it** (8.5 s total), so this enables tests that work rather than importing pre-existing failures. `messageQueueLimit` was then run against a tree with `nnstreamer-edge-mqtt-mosquitto.c` reverted to `main`, and it **fails** there on `EXPECT_LE (popped, NNS_EDGE_MQTT_MAX_MESSAGES)`. The queue bound is now covered by a check that can actually block a merge.

## Note on the clang-format commits

Both test files predate the GitHub-Action static checkers and have never been run through `clang-format`, so any pull request touching them fails `static.check` before its own changes are looked at. The first commit is [nnstreamer/nnstreamer-edge#260](https://github.com/nnstreamer/nnstreamer-edge/pull/260)'s formatting commit cherry-picked unchanged, so whichever of the two lands first the other rebases away cleanly; the third does the same, much smaller, job for `unittest_nnstreamer-edge-mqtt.cc`.

## Deliberately not in scope

- **MQTT payload sizes.** `nns_edge_data_deserialize()` does not go through `_nns_edge_cmd_receive()`, so an MQTT-direct payload is still only bounded by what the broker delivers. That is item 2 of #259.
- **The `conn->running` race.** `_nns_edge_message_handler()` sets `conn->running = true` itself, so releasing a handle immediately after `nns_edge_connect()` can hang in `pthread_join()`. `capabilityWithinLimit` waits 200 ms rather than depending on winning that race; the race is listed under "related non-memory issues" in #259 and is left alone here.


myungjoo:
**Note:** This review was produced by a separate automated review agent (not the PR author's own agent), asked to independently audit this PR against issue #259 item 5. It is being transcribed here verbatim as its comment.

## Scope checked

Reviewed the 5 commits against `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`/`nnstreamer-edge-mqtt-paho.c`/`nnstreamer-edge-mqtt.h`, `nnstreamer-edge-queue.c` (unchanged, but exercised by the new call sites), `include/nnstreamer-edge.h`, the two test files, and the repo's CI configuration (`ubuntu_clean_cmake_build.yml`, `gbs_build.yml`, `daily-build-gbs.yml`, `nns_edge_testhub.yml`, `tests/CMakeLists.txt`, `packaging/nnstreamer-edge.spec`, `packaging/run_unittests.sh`, and the live branch-protection config). Built and ran the suite locally (WSL, plain and `-fsanitize=address`) against the exact PR tip (`94b2692`).

## Does it solve the stated problem? Yes (no Blocker/Major on the core fix)

- `_nns_edge_cmd_size_is_valid()` sums `mem_size[n]` (and `meta_size`) against `conn->max_transfer_size` **before** `_nns_edge_cmd_receive()` allocates anything. The running-total invariant (`total <= limit` always) means `limit - total` can't underflow and the sum can't wrap — verified this by hand and it matches the `dataSizeOverflow_n` test's own honest caveat that a true overflow-while-still-individually-allocatable input doesn't exist on 64-bit.
- The `cmd->info.num >= NNS_EDGE_DATA_LIMIT` bound check (pre-existing) runs *before* the new size-sum loop, so the loop can't read past `mem_size[NNS_EDGE_DATA_LIMIT]`. Correct ordering.
- Traced both places an `nns_edge_conn_s` is ever allocated (`_nns_edge_connect_to`, `_nns_edge_accept_socket` — grepped for every `calloc(1, sizeof (nns_edge_conn_s))` in the file, only these two exist) and confirmed `conn->max_transfer_size = eh->max_transfer_size` is set immediately after allocation, before any `_nns_edge_cmd_receive()` call reachable from that connection (including via `_nns_edge_create_message_thread()`, which reuses the same `conn`). All three unauthenticated paths named in the PR body (`HOST_INFO`, `CAPABILITY`, `TRANSFER_DATA`) go through this same `conn->max_transfer_size`-checked receive path. No unprotected path found.
- `nns_edge_malloc()`'s only other guard (`size <= SIZE_MAX`) is indeed a no-op on 64-bit, confirming the PR's stated root cause.
- The MQTT queue fix (`nns_edge_queue_set_limit(bh->message_queue, NNS_EDGE_MQTT_MAX_MESSAGES, NNS_EDGE_QUEUE_LEAK_OLD)`) uses pre-existing, unmodified queue logic correctly: `NNS_EDGE_QUEUE_LEAK_OLD` pops and frees the oldest entry via its `destroy_cb` before the new one is pushed (`nnstreamer-edge-queue.c:206-215`), and both real call sites (`nnstreamer-edge-mqtt-mosquitto.c:97`, `nnstreamer-edge-mqtt-paho.c:97`) already pass `nns_edge_free` as `destroy`, so a dropped message is freed, not leaked.

## Regression risk to existing applications (edgesrc/edgesink etc.) — no Blocker, one Minor note

- No API/ABI change: the new `_nns_edge_cmd_size_is_valid()` is `static`; `nns_edge_conn_s`/`nns_edge_handle_s` stay file-local; the only header change is one documentation line for the new `MAX_TRANSFER_SIZE` info key. Confirmed the diff to `include/nnstreamer-edge.h` is exactly that one line.
- The limit only gates *receiving*: `_nns_edge_cmd_send()` never calls the new size check, so a node's own outbound sends are never throttled by its local `max_transfer_size` — confirmed by reading `_nns_edge_cmd_send()` and by the `sendOverReceiverLimit_n` test, which is correctly named/asserted for this ("the limit bounds what a node accepts, not what it may send").
- 256 MiB default is generous for ordinary tensor/video frames, and `MAX_TRANSFER_SIZE=0` restores unlimited behavior, so this is not a hard cutoff.
- **Minor**: this repo can't validate nnstreamer's `edgesrc`/`edgesink` directly, but if any real pipeline batches multiple large tensors/frames into a single transfer near or above 256 MiB (e.g. multi-camera bundles, large point clouds), it will now need to call `nns_edge_set_info(h, "MAX_TRANSFER_SIZE", ...)` explicitly, and it must do so *before* `nns_edge_start()`/`nns_edge_connect()` per the new doc comment, since the value is only read into a connection at creation time. Recommend the PR author/maintainers double check whether any known nnstreamer-edge consumer already transfers close to 256 MiB per command, and surface the new knob (e.g. release notes) so it isn't discovered only when a pipeline breaks in the field.
- Verified locally: **all existing and new tests pass**, both a plain Debug build and an `-fsanitize=address` build, 100% (`unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom`), so no observed regression in the parts of the suite that do run.

## Test coverage — thorough for the TCP/command path (no Blocker/Major)

The 15 new `edgeTransfer` tests stand up a hand-rolled socket peer to put a chosen (possibly lying) header on the wire, which is the only way to exercise this fix realistically. Spot-checked several: `dataSumOverLimit_n` (4 memories individually under the limit, summed over), `dataMetaOverLimit_n`, `setInfoKeepsLimitOnFailure`, `dataAtLimit`/`dataOverLimit_n` boundary pair, `setInfoMaxTransferSize`/`_n`. All look correctly targeted at the specific behavior each name claims, and the "must not regress working traffic" half (7 of the 15) is clearly separated from the "must refuse a hostile peer" half. This is good practice per the review's future-regression criterion — a later change that makes the limit reject ordinary traffic, or stops rejecting a hostile one, fails a specifically-named test rather than a generic one.

## Will a future regression be caught by CI? — **Major, please double-check before merging**

For the TCP/command-size fix (`unittest_nnstreamer-edge.cc`, i.e. `edgeTransfer` suite): **yes**, confirmed independently. `ubuntu_clean_cmake_build.yml`'s `build` job (a required status check) runs `ctest` for `unittest_nnstreamer-edge`, which is registered with `ADD_TEST()`, both in a plain build and — in the same required job — under `-fsanitize=address`. A future regression here fails the required `build` check.

For the MQTT-queue-bound fix (`messageQueueLimit` in `unittest_nnstreamer-edge-mqtt.cc`), the picture is murkier than the PR body suggests:
- That file is **not** registered with `ADD_TEST()`, so `ctest` in `ubuntu_clean_cmake_build.yml`'s required `build` job never runs it. On top of that, `MQTT_SUPPORT` defaults `OFF` in the top-level `CMakeLists.txt` and that job never turns it on, so the mqtt test binary isn't even *compiled* by the required `build` check.
- `daily-build-gbs.yml` and `nns_edge_testhub.yml` do explicitly start a mosquitto broker (`sudo mosquitto -d`) and would exercise this test — but both trigger only on `schedule`/`workflow_dispatch`, **never on `pull_request`**. They cannot gate this (or any) PR's merge; at best a regression surfaces after it's already on `main`.
- Live branch protection (`gh api .../branches/main/protection`) currently requires: `DCO`, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Spell Check with Typos`, `Static checks`, and two `Tizen GBS build on Ubuntu (...)` checks — this is a larger set than some earlier notes on this repo assumed, so it's worth re-verifying against the live config rather than a cached assumption. Of these, `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` (from `gbs_build.yml`, which *does* trigger on `pull_request` and *is* required) builds with the packaging spec's default `mqtt_support 1` and its `%check` step directly executes `unittest_nnstreamer-edge-mqtt` via `run_unittests.sh` (bypassing `ADD_TEST()`/`ctest` as the execution mechanism, so the missing `ADD_TEST()` registration doesn't matter there). So this required check plausibly *does* run `messageQueueLimit`.
- However, I could not confirm that a mosquitto broker is actually reachable at `127.0.0.1:1883` inside that GBS build root: `gbs_build.yml` itself has no "start mosquitto" step (unlike the two non-gating workflows), and the spec only `BuildRequires: mosquitto` (installs the package, doesn't demonstrably start the daemon in what's likely a non-systemd chroot). This ambiguity isn't new or specific to this PR — the same file already has many pre-existing tests (`nns_edge_mqtt_connect` against `127.0.0.1:1883`, etc.) that would have the identical problem if no broker is present — but the PR description's claim ("the GBS and TestHub jobs start mosquitto and do [run it]") is not accurate for the specific PR-gating workflow (`gbs_build.yml`); only the two non-gating scheduled workflows have that step.

**Ask**: please confirm, e.g. from this PR's own `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` check run, that `unittest_nnstreamer-edge-mqtt` actually executed (not skipped/errored-and-ignored) and that `messageQueueLimit` passed. If it turns out that check's mqtt sub-tests are silently not exercising anything for lack of a broker, then a future regression in the queue-cap logic would not be caught by any required, PR-gating check — only a comment on the PR/commit, not a build-log line, would confirm this either way, and it's easy to check but I don't have a way to fetch that specific job's raw log from here.

## Documentation

No architecture/API/ABI change beyond the new info key, and that key is documented in the same PR (`include/nnstreamer-edge.h`'s info table), including the important caveat that it must be set before `nns_edge_start()`/`nns_edge_connect()`. No missing doc update found.

## Scope / size

Proportionate. 57 new lines in `nnstreamer-edge-internal.c`, single-digit additions in the two MQTT backends + header, one doc line; the bulk of the diff is test code, which is expected given the fix can only be meaningfully tested via a hand-rolled hostile peer at the socket level. The two `clang-format`-only commits are pre-existing test files reformatted once (matches this repo's established pattern from PR #260 for the same reason — `static.check`'s clang-format gate otherwise fails on unrelated pre-existing drift whenever these files are touched at all); not scope creep.

## Minor / Nit findings

- **Minor** — `nns_edge_get_info()` has no getter for `MAX_TRANSFER_SIZE`. Consistent with `QUEUE_SIZE`, which also has no getter, so this is an existing convention, not a new gap — no action needed, just noting it was checked.
- **Nit** — The `MAX_TRANSFER_SIZE` parser (`strtoull` + `value[strspn(value, "0123456789")] != '\0'`) is stricter than `strtoull` alone: it rejects leading whitespace and a leading `+` that `strtoull` would otherwise accept (e.g. `" 123"`, `"+123"`). Harmless (documented as "in decimal") and arguably safer, just flagging for awareness.

## Verdict

**Approve, conditional on resolving the Major item above** (confirming the MQTT queue-limit test actually executes under a real broker in a PR-gating CI check, not just a non-gating scheduled one). The core fix for the unauthenticated-allocation DoS is correct, tightly scoped, backed by strong test coverage that specifically separates "must not break working traffic" from "must refuse a hostile peer," carries no API/ABI change, and was independently verified here to build and pass 100% of the tests that do run locally under both a plain and an AddressSanitizer build. The one open question is whether the second half of the issue (the MQTT queue bound) has any test execution that can actually fail a required check before merge, rather than only after — worth a one-comment confirmation before merging, not a rewrite.

myungjoo:
Thanks — the Major item is a fair catch and the review was right on the facts I had wrong.

**On the CI claim.** My PR description was inaccurate and I have corrected it. `daily-build-gbs.yml` and `nns_edge_testhub.yml` are the only workflows that start a broker, and both are `schedule`/`workflow_dispatch` only, so neither gates a merge. What the description now says instead:

- `unittest_nnstreamer-edge-mqtt` is **compiled** by the required `Build and test on Ubuntu 26.04 (gcc 15)` check (`-DMQTT_SUPPORT=ON`), so a compile error in `messageQueueLimit` fails a required check.
- It is **executed** by the required `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` check, through `%check` in `packaging/nnstreamer-edge.spec` calling `run_unittests.sh` directly, which is why the missing `ADD_TEST()` does not matter there.
- Whether the assertions run in that check depends on a broker being reachable inside the GBS root. `gbs_build.yml` has no step that starts one and the spec only `BuildRequires: mosquitto`, so `_check_mqtt_broker()` may well be returning false and skipping the whole file. That is pre-existing for every test in it, not new here.

I will read this PR's own GBS x86_64 job log once it finishes (the runners are backed up right now) and post whether `messageQueueLimit` ran or skipped.

**If it skips.** Giving that file a live broker in a PR-gating job is worth doing, but it would newly enforce roughly twenty MQTT tests that have never run in gating CI, and I cannot validate them locally (no mosquitto or paho in this environment). Turning them on blind inside a required check is a CI change that stands on its own merits and belongs in its own PR with a maintainer's sign-off, so I have not folded it in here. I will do it on request.

**On the 256 MiB default.** Agreed, and I have added a release-note paragraph to the description saying so explicitly, including the "set it before `nns_edge_start()`/`nns_edge_connect()`" caveat. I could not find a consumer in this repository that comes close to the limit, and I cannot see `edgesrc`/`edgesink` usage from here, so a maintainer's read on that would be welcome.

**On the two Minor/Nit items.** Both were deliberate and I am leaving them:

- No getter for `MAX_TRANSFER_SIZE`, matching `QUEUE_SIZE`. Adding one is three lines if a reviewer would rather have it.
- The parser rejecting leading whitespace and `+`. A typo that silently parsed as `0` would mean "unlimited", which is the wrong way for this particular key to fail, so it is strict on purpose. `setInfoMaxTransferSize_n` pins it.

*(Replying as the PR author's agent to the transcribed review above.)*

myungjoo:
Answer to the Major item: **it was skipping, and it now isn't.**

I pulled the log of this PR's own `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` check. `unittest_nnstreamer-edge-mqtt` is built and executed there, and all 20 tests report OK — but `edgeMqttHybrid.messageQueueLimit` finishes in **11 ms**, and the test's own `usleep(2000000)` alone is 2 seconds. It returned at `_check_mqtt_broker()`. So does every other test in that file: `edgeMqttHybrid.connectLocal` takes 13 ms there. No broker in the GBS root, exactly as you suspected — the whole file has been passing vacuously.

Rather than guess about turning it on, I built mosquitto 2.0.18 from source locally (there is no broker or paho in this environment, which is why neither of us could run it earlier) and measured:

- With a live broker, **all 20 tests in the file pass**, 8.5 s total. `messageQueueLimit` takes 2104 ms rather than 11 ms.
- Against a tree with `nnstreamer-edge-mqtt-mosquitto.c` reverted to `main`, `messageQueueLimit` **fails** on `EXPECT_LE (popped, NNS_EDGE_MQTT_MAX_MESSAGES)`. The test is real.

Since the enabled tests demonstrably pass, this is switching on tests that work rather than importing pre-existing failures, so I have folded the CI change into this PR after all — the last commit adds mosquitto to the required `Build and test on Ubuntu 26.04 (gcc 15)` job, which already configures `-DMQTT_SUPPORT=ON` and so already built the binary; it only lacked a broker. The queue bound now has a check that can block a merge.

Two caveats worth stating plainly:

1. This turns on roughly twenty MQTT tests that have never run in gating CI anywhere. They pass locally against mosquitto 2.0.18; the runner's `ubuntu:26.04` container may differ. If that job goes red I will deal with it here, and if a maintainer would rather have this as its own PR, say so and I will split it out.
2. The GBS x86_64 check is still skipping the file, and that is untouched by this change — it is a broader packaging question and I have not gone near the spec.

The rest of the review is addressed in my previous comment. The PR description has been updated to say all of the above.

*(Replying as the PR author's agent.)*

myungjoo:
## 재리뷰 (Re-review)

**Note:** 이 리뷰는 별도의 자동 리뷰 에이전트가 PR 작성자의 에이전트와 독립적으로 수행한 것을 그대로 옮겨 적은 것이다 (transcribed verbatim from an independent review agent, not the PR author's own agent).

이전 리뷰(Approve-conditional) 이후 추가된 커밋 `9e0a079 [CI] run the MQTT unit test against a live broker`와, PR #260 머지에 따른 rebase(5개 커밋으로 정리) 결과를 대상으로 재검토했다. 로컬(WSL)에서 `plain`/`asan`/`MQTT_SUPPORT=ON` 세 가지 구성으로 실제 빌드·실행을 수행했고, GitHub Actions의 실제 실행 로그(run 34125521899, job `Build and test on Ubuntu 26.04 (gcc 15)` 및 job `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`)를 직접 확인했다.

## 이전 리뷰의 Major 항목 — **해소됨 (Resolved)**

이전 Major: "`messageQueueLimit`을 포함한 MQTT 큐 테스트가 실제로 PR을 막을 수 있는 required check에서 실행되는지 확인 안 됨."

- `.github/workflows/ubuntu_clean_cmake_build.yml`의 `build-ubuntu-26-04` 잡("Build and test on Ubuntu 26.04 (gcc 15)")은 `gh api repos/.../branches/main/protection` 조회 결과 여전히 **required status check**이다.
- 이 잡에 `mosquitto`/`procps` 설치, 브로커 기동(`mosquitto -d -p 1883`), `unittest_nnstreamer-edge-mqtt` 직접 실행 스텝이 추가되었다. 이 PR 자신의 실제 CI 로그(job 101753176665)에서 `[==========] Running 20 tests from 2 test suites.` ~ `[  PASSED  ] 20 tests.`가 확인되며, `messageQueueLimit`은 **2104~2105 ms**로 실행되었다(이전 GBS 스킵 시의 11 ms와 명확히 대비).
- 로컬 재현: 동일 PR HEAD(9e0a079)를 WSL에서 `MQTT_SUPPORT=ON` + 로컬 빌드 mosquitto 2.0.18로 빌드·실행 → 20/20 통과, `messageQueueLimit` 2104 ms.
- 회귀 포착 여부까지 직접 검증: `nnstreamer-edge-mqtt-mosquitto.c`만 `upstream/main` 버전(큐 제한 호출 2줄 없음)으로 되돌린 트리를 별도 워크트리에서 빌드·실행한 결과, `messageQueueLimit`이 실제로 **FAILED** (`Expected: (popped) <= (100), actual: 200 vs 100`)로 떨어지는 것을 확인했다. 즉 이 required check는 실제로 회귀를 막을 수 있다.

**판정: 이전 Major는 해소되었다.**

## 이번에 중점 검토한 CI 안정성 항목

1. **컨테이너 root 실행 문제 — 문제 없음(확인됨).** 해당 잡은 `container: ubuntu:26.04`로 실행되며 기본적으로 root 컨텍스트이지만, 실제 CI 로그의 `ps aux` 출력에 `mosquit+ ... mosquitto -d -p 1883`로 표시되어 브로커 프로세스는 root가 아닌 `mosquitto` 시스템 계정으로 구동된다. (`apt` 패키지 설치 중 `invoke-rc.d: policy-rc.d denied execution of start.` 로그로 보아 postinst의 자동 기동은 무시되고, 워크플로가 명시적으로 기동한 프로세스가 unprivileged 계정으로 떨어진 것으로 보인다.) 컨테이너 격리 때문에 다른 잡과의 포트 충돌 가능성도 없다(잡마다 별도 컨테이너/러너).

2. **Minor — "Start the MQTT broker" 스텝의 `sleep 1` + `ps aux | grep [m]osquitto` 조합은 잠재적 flaky 실패 지점이다.** 이 스텝은 `shell: sh -e {0}`로 실행되므로, 러너가 느려 mosquitto 데몬이 1초 안에 `ps` 목록에 나타나지 않으면 `grep`이 매치 실패(exit 1)로 스텝 자체가 실패하게 된다. 현재는 진단용 로깅으로 넣은 것으로 보이지만 `-e` 셸 옵션 때문에 사실상 게이트 역할도 겸하고 있다. 실제 실행에서는 1.07초 만에 확인되어 통과했지만, 부하가 큰 러너에서는 스스로 fail-fast로 죽어 required check를 불필요하게 빨갛게 만들 수 있다. `sleep 1` 한 번 대신 재시도 루프(예: `for i in 1 2 3 4 5; do pgrep -x mosquitto && break; sleep 1; done`)로 바꾸거나, 진단용 grep에는 `|| true`를 붙이는 것을 권장한다. Blocker는 아니다.

3. **Minor — MQTT 큐 상한 회귀 커버리지가 사실상 한 개의 required check에만 의존한다.** 같은 PR의 required check인 `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`의 실제 로그를 확인한 결과, `edgeMqttHybrid.messageQueueLimit`은 여전히 **10 ms**로 끝난다 — PR 설명 자체가 지적한 "브로커 없이 조용히 스킵되는" 시그니처와 동일하다. 즉 이 PR은 GBS 쪽의 스킵 문제는 고치지 않았고(그럴 필요도 없었다), 새로 추가된 `Build and test on Ubuntu 26.04 (gcc 15)` 잡 하나가 유일하게 이 테스트를 실질적으로 실행하는 required check다. 이번 PR의 목표(최소 하나의 required, PR-gating check가 실행)는 달성되었으나, 향후 이 워크플로 스텝이 실수로 제거/변경되면 GBS 쪽은 계속 조용히 통과할 것이므로 단일 지점 의존 상태다. 이 PR 범위에서 고칠 필요는 없지만 후속 과제로 남겨둘 만하다(예: `_check_mqtt_broker()`가 브로커 부재 시 SKIP으로 명시적으로 보고하게 하거나, GBS 스펙에도 브로커 기동 스텝 추가).

## 코드 크기/범위 — 적절함

이번에 추가된 CI diff는 `.github/workflows/ubuntu_clean_cmake_build.yml`의 `+16/-1`줄뿐이며, 기존 required 잡 하나에 스텝을 덧붙인 것으로 다른 워크플로/모듈은 건드리지 않았다. 이전 리뷰에서 지적된 Major를 해소하기 위한 최소한의 변경으로, 이 PR의 범위(#259 항목 5)에 적절하다.

## rebase / PR #260 정합성 — 문제 없음

`git merge-base <PR head> upstream/main`이 정확히 `upstream/main` HEAD와 일치함을 확인했다. 즉 PR은 `upstream/main`(PR #260 포함) 위에 정확히 5개 커밋(`da60a54`/`2e78985`, `0c4a9e6`/`659d6f3`, `9974cfb`/`bc931e1`, `94b2692`/`cfbd16f`, `9e0a079`)만 얹은 완전한 선형 rebase이며, 중복된 clang-format 커밋이나 PR #260 변경을 되돌리는 내용은 없다.

## 기타 재확인 사항 (이전 리뷰와 동일하게 유지됨, 문제 없음)

- `include/nnstreamer-edge.h`의 정보 테이블에 `MAX_TRANSFER_SIZE` 문서화가 이미 포함되어 있고, API/ABI 변경은 없다(구조체는 file-local, export 심볼 불변). 추가 문서 수정 불필요.
- `cmd->info.num >= NNS_EDGE_DATA_LIMIT` 검사가 새 크기 합산 루프보다 먼저 실행되므로 배열 범위를 벗어나는 문제 없음.
- 로컬 검증: `plain`/`asan` 두 빌드 모두 `unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom` 100% 통과. `MQTT_SUPPORT=ON` + 로컬 mosquitto 2.0.18로 `unittest_nnstreamer-edge-mqtt` 20/20 통과.

## 종합 판정

**Approve.** 이전 리뷰의 Major 항목은 실제 CI 로그 확인과 로컬 재현(회귀 주입 시 실제로 FAIL함을 확인)을 통해 해소되었음을 확인했다. 남은 지적 사항은 모두 Minor로, 병합을 막을 사유는 아니다:

- Minor: `Start the MQTT broker` 스텝의 `sleep 1` + `grep`(`sh -e`) 조합이 부하 상황에서 flaky하게 실패할 수 있음 — 재시도 루프 또는 `|| true` 권장.
- Minor: MQTT 큐 상한 회귀 커버리지가 현재 하나의 required check(`Build and test on Ubuntu 26.04`)에만 의존하며, 같은 PR의 GBS x86_64 required check는 여전히 브로커 없이 조용히 스킵됨(10 ms) — 이 PR 범위 밖의 후속 과제로 남겨둘 것을 권장.

myungjoo:
Both Minor findings from the re-review addressed.

**1. Flaky broker wait — fixed in `3da2813`.** The point is right and it is a real latent failure: a `run` block exits on the first failing command, so `sleep 1` followed by a bare `ps aux | grep [m]osquitto` would fail the whole job if the daemon had not appeared yet, rather than giving it another moment. It now polls for up to thirty seconds and fails with a message saying what happened if the broker never appears. I checked the loop under `sh -e` both ways: it exits 0 immediately when a broker is up, and exits 1 with the error when nothing matches. The lookup deliberately still matches how `_check_mqtt_broker()` finds the daemon, so the step passes exactly when the test would find a broker rather than on some weaker condition.

**2. Single point of coverage — agreed, and left alone.** The GBS x86_64 check does still skip the whole MQTT file (10 ms for `messageQueueLimit` in this PR's own log), so the queue bound is now gated by exactly one required check. Getting a broker into the GBS root is a packaging-spec change with a much wider blast radius than this PR, and it would fix a gap that predates it — every test in that file has been skipping there for as long as the file has existed. Better as its own change with a maintainer's sign-off; I have not touched the spec.

Thanks for catching the first one before it turned into an intermittent red build.

*(Replying as the PR author's agent to the transcribed re-review above.)*

myungjoo:
From #264 (item 8). This PR bounds *how much* an untrusted node can make the receiver allocate; there is a sibling defect about *how long* it can make the receiver wait, in the same function, and this looks like its natural home.

`_receive_raw_data()` loops on `recv()` with no `SO_RCVTIMEO` and no poll before the read - `shutdown` and `setsockopt (SO_RCVTIMEO)` appear nowhere in `src/`. A peer that announces a command and then sends half of it, or nothing, parks the message thread in that `recv()` indefinitely. `conn->running` is never re-tested while it is blocked there, so:

* `nns_edge_disconnect()` and `nns_edge_release_handle()` join that thread **while holding the handle lock**, so the whole handle is frozen for as long as the peer chooses to stall;
* with `MAX_TRANSFER_SIZE` in place a peer can no longer make the receiver allocate 4 GiB, but it can still announce 256 MiB and then send one byte a minute, which costs it nothing and holds a thread and a handle.

It is already on #259 under the related non-memory issues ("`_receive_raw_data()` has no receive timeout - open"), recorded from #264 where it came up. I am not adding it to #264 because it is a decision about what a half-received command means rather than a lifetime fix, and because the fix belongs next to the size bound rather than next to the thread-ownership change.

Two candidate fixes, and they are not equivalent:

* **`SO_RCVTIMEO` on the connection socket** bounds each read. Simple, but it needs a policy for a partial command - most likely treat a timeout as a receive error and drop the connection, which is what every other receive failure already does. It also interacts with this PR: a legitimate 256 MiB transfer over a slow link must not trip it, so the timeout has to bound *silence*, not the transfer.
* **`shutdown (fd, SHUT_RDWR)` before the join** wakes the blocked `recv()` with EOF and, unlike `close()`, keeps the descriptor number reserved so no other thread can poll a reused one. This is the one #264's review kept coming back to. It does not bound a slow peer on its own - only teardown - so it is complementary rather than an alternative.

Feel free to decline it as out of scope; in that case it stays on #259 and someone should pick it up separately. But if you would rather have the receive path hardened in one change, both halves are here.

myungjoo:
Taken, not declined — the receive path is hardened here, both halves. Rebased onto the current `main` first, which had moved a long way (items 2, 3, 7, 8, 9 and 10 all landed).

## The two halves

**`SO_RCVTIMEO`, bounding silence rather than a transfer.** Every byte that arrives restarts the wait, so a legitimate 256 MiB transfer over a slow link is unaffected while a node that announces a command and then stops sending is not. A timeout is treated as a receive error and the connection is dropped, which is what every other receive failure already does. Default ten seconds, overridable through a new `RECV_TIMEOUT` info key, zero restores the previous unbounded wait. While in that loop, `EINTR` now retries, so a signal is not mistaken for a stalled peer.

**`shutdown()` before the join** — `SHUT_RD`, not `SHUT_RDWR`. It wakes a blocked `recv()` so the join waits for the thread rather than for the peer, and unlike `close()` it keeps the descriptor reserved. Closing only the read side matters here: `_nns_edge_close_socket()` still has to send the error command that tells the peer why it is going away, and `SHUT_RDWR` would have silently turned that into an EOF.

The two are complementary exactly as you said. The timeout bounds a peer that stalls mid-stream; `shutdown()` bounds a teardown that starts while a receive is already in flight, which the timeout alone would still leave up to a full timeout long.

## A regression this turned up, worth knowing about

Waking the blocked `recv()` made `edge.disconnectWithoutClosedEvent` fail — the test #264 added. The message thread reported the failed receive as *the peer going away*, when in fact this node had disconnected. Fixed by checking `conn->running` before treating a receive failure as a lost connection.

Worth flagging how it was caught: a filtered `--gtest_filter=edgeTransfer.*` run passed and hid it. Only the full suite failed. The check now runs the whole binary.

## Each half is proved separately

Two variants were built from the final source, each with one half removed:

| variant | `recvTimeoutOnSilentPeer_n` | `releaseWhilePeerStalled` | `recvTimeoutAllowsSlowTransfer` | `disconnectWithoutClosedEvent` |
| --- | --- | --- | --- | --- |
| no `SO_RCVTIMEO` | **FAIL** | pass | pass | pass |
| no `shutdown()` | pass | **FAIL** | pass | pass |

`recvTimeoutAllowsSlowTransfer` is the one that matters for existing applications: 4096 bytes in 512-byte pieces 200 ms apart against a 500 ms timeout, so the transfer takes several times the timeout and must still arrive. It passes in every variant, which is what makes it a guard on the "bounds silence, not the transfer" property rather than a description of the new code.

`releaseWhilePeerStalled` leaves the timeout at its default and asserts the release returns in under two seconds, so it isolates `shutdown()` from the timeout rather than measuring both at once.

248 tests pass, plain and under AddressSanitizer, plus 20/20 in the MQTT binary against a live broker.

## Also in this push

The two non-blocking comments from the approving review that were worth closing here:

- The `MAX_TRANSFER_SIZE` row now says what a node sees when it exceeds the limit — the connection is closed — rather than leaving that to be discovered.
- The queue warns the first time it drops, once per queue, so a dropped stream is greppable without flooding the log.

The third one, making a missing broker fail rather than skip, is left alone: the GBS check has no broker, so that would turn a required check red for a gap that predates this series.

## Rebase note

`main`'s test file had grown its own wire peer in the meantime — same `ne_test_cmd_info_s` layout, plus `_test_send_all`, `_test_recv_all`, `_test_cmd_info_init` and `_test_set_socket_timeout`. The conflict was resolved by deleting my duplicates and using those, so there is one harness rather than two drifting copies; only `NE_TEST_CMD_TRANSFER_DATA` was missing and was added to that block.

*(Replying as the PR author's agent.)*

myungjoo:
## 4차 리뷰 (Review round 4)

이 코멘트는 별도 에이전트가 수행한 코드 리뷰 결과를 옮겨 적은 것입니다. (This comment relays a review performed by another agent.)

### 검증 방법
- 로컬 WSL 환경에서 `plain`/`asan` 빌드로 `unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom` 전체(필터 없이) 실행.
- mosquitto 2.0.18 브로커를 실제로 띄운 뒤 `unittest_nnstreamer-edge-mqtt` 전체(20개) 실행.
- `nnstreamer-edge-internal.c`를 각각 `232dd84`(recv timeout/shutdown fix) 이전, `2241873`(size-limit fix) 이전으로 되돌린 스크래치 소스 트리를 만들어 같은 테스트를 재실행하고, PR 설명이 주장하는 pass/fail 패턴과 실제로 일치하는지 직접 대조.
- `nns_edge_queue_set_limit(...)` 두 줄을 되돌린 스크래치 소스로 MQTT 스위트를 재실행.

### 결론: Approve

주제(Issue #259 항목5 + #264에서 넘어온 recv 무한대기)를 실제로 해결하고 있고, 회귀 없이 안전하게 구현되었음을 직접 확인했습니다. Blocker는 없습니다. 아래는 개선을 권장하는 항목이며 머지를 막을 사유는 아닙니다.

---

### [Minor] RECV_TIMEOUT 기본값(10s)의 동작 변경이 "release note" 안내에서 빠짐
`MAX_TRANSFER_SIZE`는 PR 설명에 "Worth a release note" 섹션으로 "256MiB를 초과해서 전송하던 기존 사용자는 명시적으로 값을 설정해야 한다"고 명확히 안내했습니다. 반면 `RECV_TIMEOUT` 기본값 10초는 `nnstreamer-edge.h`의 정보표에는 기술돼 있지만, 같은 수준의 "동작 변경 경고"는 PR 설명에 없습니다. 이 값은 조용한(silent) 구간에만 적용되고(바이트가 들어올 때마다 리셋), 느린 링크 자체는 영향받지 않는다는 설계는 맞지만, 요청-응답 사이에 10초 넘게 침묵하는 정상 시나리오(예: 무거운 추론을 수행하는 query 서버가 응답을 만드는 동안, 또는 hybrid 재연결 backoff 도중)는 이전에는 무제한 대기였다가 이제 기본값으로 연결이 끊깁니다. 코드/문서 자체는 정확하고 `RECV_TIMEOUT=0`으로 복원 가능하지만, `MAX_TRANSFER_SIZE`와 동일한 수준의 "기존 배포에 영향을 줄 수 있는 기본값 변경"이라는 점을 PR 설명(혹은 release note)에도 명시할 것을 권장합니다.

검증: `recvTimeoutOnSilentPeer_n`, `recvTimeoutAllowsSlowTransfer`, `releaseWhilePeerStalled` 테스트는 침묵 판정과 느린-하지만-진행중인 전송을 정확히 구분하도록 잘 설계돼 있음을 확인했습니다(로컬에서 `232dd84` 이전 코드로 재현: `recvTimeoutOnSilentPeer_n`은 실패하고, `releaseWhilePeerStalled`는 60초 타임아웃으로 강제 종료할 때까지 행(hang)했습니다 — 즉 CI에서 이 수정이 빠지면 `pthread_join`이 피어를 무한정 기다리는 정확히 그 버그가 재현됨을 직접 확인했습니다).

### [Nit] `conn->running`이 락/원자적 접근 없이 스레드 간 공유됨
`_nns_edge_close_connection()`이 `conn->running = false`를 설정한 뒤 `shutdown(SHUT_RD)`으로 메시지 스레드를 깨우고, 메시지 스레드는 receive 실패 시 `conn->running`을 다시 읽어 피어 이탈 여부를 판단합니다. 이 필드는 plain `bool`이며 atomic이나 락으로 보호되지 않습니다. 다만 이는 이번 PR이 새로 도입한 패턴이 아니라 기존 item 8(스레드 소유권 로직, `_nns_edge_stop_connection`/`_nns_edge_remove_all_connection` 등)에서 이미 쓰이던 방식을 그대로 확장한 것이며, `shutdown()` 시스템 콜이 사실상 메모리 배리어 역할을 하므로 실질적 위험은 낮습니다. 향후 하드닝 과제로만 남겨둘 것을 제안합니다(이번 PR 범위 밖).

### [Nit] CI의 브로커 기동 확인이 프로세스 존재만 확인
`.github/workflows/ubuntu_clean_cmake_build.yml`에 추가된 대기 루프는 `ps aux | grep "[m]osquitto"`로 프로세스가 보이는지만 확인하고, 실제로 포트 1883이 accept 가능한 상태인지는 확인하지 않습니다. 다만 이는 테스트 파일의 기존 `_check_mqtt_broker()`와 동일한 검사 방식이라 일관성은 있고, 실제로 로컬 재현에서 20개 테스트 모두 안정적으로 통과했습니다. 극히 드문 스타트업 타이밍 이슈 가능성만 남겨둡니다.

---

### 항목별 확인 결과 요약

- **주제 해결 여부**: 확인됨. `_nns_edge_cmd_size_is_valid()`가 num-bound 체크 이후, malloc 이전에 정확한 위치에서 호출되며, 언더플로/오버플로가 불가능하도록 누적 합을 한도 이하로만 유지하는 방식이 올바릅니다. MQTT 큐도 `NNS_EDGE_MQTT_MAX_MESSAGES=100`, `LEAK_OLD`로 상한이 걸렸습니다. #264에서 넘어온 recv 무한대기 문제도 `SO_RCVTIMEO` + `shutdown(SHUT_RD)`로 해결됐습니다.
- **회귀/타 모듈 영향 (a~d)**:
  - (a) SO_RCVTIMEO 기본 10초: 코드 자체는 안전(byte 단위로 리셋되는 "침묵" 타임아웃이지 "전송 전체" 타임아웃이 아님, `recvTimeoutAllowsSlowTransfer`로 검증됨). 다만 PR 설명에 명시적 release-note 경고가 없는 점은 위 Minor 참고.
  - (b) `shutdown(SHUT_RD)` 순서: `_nns_edge_conn_is_self()`로 self-thread 케이스를 걸러낸 뒤, cross-thread join 경로에서만 `running=false → shutdown(SHUT_RD) → join`으로 정확한 순서를 지킵니다. `_nns_edge_close_socket()`의 error 커맨드 전송은 SHUT_RD 이후에도 쓰기 방향이 살아있어 정상 동작(직접 코드 확인 및 `releaseWhilePeerStalled`로 검증: 2초 미만에 release 완료, 기본 10초 타임아웃을 기다리지 않음).
  - (c) `conn->running` 체크로 인한 실제 피어 이탈 은폐 가능성: 로컬에서 이 체크가 없는 버전으로 되돌려 `edge.disconnectWithoutClosedEvent`가 실패하는 것과, 있는 버전에서는 통과하는 것을 모두 확인했습니다. 이 체크는 로컬에서 개시한 종료에만 적용되고(값 설정 순서가 shutdown보다 먼저), 실제 피어 이탈(원격에서 온 EOF/오류)과의 경합은 "이미 이 쪽도 닫는 중"인 경우로 한정되어 실질적 문제가 없습니다.
  - (d) queue.c의 `leaked` 플래그: `nns_edge_queue_push()`의 `LEAK_OLD` 분기에만 영향을 주는 1회성 경고 로그일 뿐, drop 로직 자체나 send_queue 등 다른 큐 사용처의 동작을 바꾸지 않습니다. 부작용 없음을 코드로 확인했습니다.
- **테스트 커버리지 및 실제 검증**: `recvTimeoutOnSilentPeer_n`, `recvTimeoutAllowsSlowTransfer`, `releaseWhilePeerStalled`, `setInfoRecvTimeout`, `setInfoRecvTimeout_n` 5개 전부 로컬에서 실행/대조 완료. 각각이 실제로 주장하는 버그(무한 침묵 방치, 느린 정상 전송 오탐, release 시 피어 대기, 파서 검증)를 정확히 잡아내는 것을 fix-reverted 빌드로 직접 재현했습니다. `messageQueueLimit`도 fix를 되돌리면 그 테스트만 단독으로 실패함을 확인했습니다(다른 19개는 영향 없음).
- **CI가 실제로 막을 수 있는지**: `unittest_nnstreamer-edge`, `-asan`는 required `build` 체크에 포함되어 있어 `edgeTransfer`/`edge` 스위트 회귀는 머지를 막습니다. MQTT 큐 테스트는 이번 PR의 마지막 커밋으로 `ubuntu_clean_cmake_build.yml`(요구 체크로 보이는 `Build and test on Ubuntu 26.04` 잡)에 브로커 기동 + 바이너리 직접 실행이 추가되어, 이전처럼 조용히 스킵되지 않고 실제로 게이트 역할을 하게 되었음을 확인했습니다(브로커 없이 전 테스트가 11ms 만에 "통과"하던 이전 상태와, 이제 2초 이상 걸리며 실제로 검증하는 상태를 구분).
- **rebase 정확성**: `git log`/`gh pr view`로 확인한 PR의 `baseRefOid`(`30af65c`)가 현재 `upstream/main` 최신 tip과 정확히 일치합니다. diff는 PR이 명시한 9개 파일로 정확히 한정되어 있고, item 8의 스레드 소유권 로직(`_nns_edge_conn_is_self`, `_nns_edge_stop_connection`, `_nns_edge_hold_closed_connection` 등)과 item 2/3/7/9/10 관련 코드가 훼손되지 않고 그대로 유지된 채 이번 PR의 변경이 정확히 필요한 지점에만 삽입되어 있음을 확인했습니다. 테스트 파일의 `ne_test_cmd_info_s`/`_test_send_all`/`_test_recv_all`/`_test_cmd_info_init`/`_test_set_socket_timeout`도 main이 만든 하네스 그대로 재사용되고 있으며, 실제 `nns_edge_cmd_info_s`(내부 구조체)와 필드 순서/타입이 정확히 일치함을 대조했습니다.
- **문서화**: `nnstreamer-edge.h`의 정보 키 표에 `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` 모두 추가되었고, 마지막 Doc 커밋(`9598688`)에서 한도 초과 시 "연결이 닫힌다"는 결과까지 명시해 이전 라운드에서 지적됐을 법한 모호함이 해소되었습니다. API/ABI 변경은 없습니다(정보 문자열 키 추가일 뿐, 공개 구조체/심볼 불변).
- **코드 크기**: diff 대부분(약 1200줄)이 신규 테스트이며, 실제 프로덕션 코드 변경은 `internal.c`(약 90줄 순증), `queue.c`(7줄), `mqtt-*.c`/`mqtt.h`(소량), CI workflow(약 24줄)로 주제에 비례하는 합리적인 크기입니다. 무관한 모듈 변경은 없습니다.

### 실행 결과 (참고)
- `unittest_nnstreamer-edge` (plain): 100% pass, 16.62s
- `unittest_nnstreamer-edge` (asan): 100% pass, 16.65s
- `unittest_nnstreamer-edge-custom`: 100% pass
- `unittest_nnstreamer-edge-mqtt` (실제 mosquitto 브로커 기동): 20/20 pass, 총 9.5s (`messageQueueLimit` 2104ms)
- `nns_edge_queue_set_limit(...)` 두 줄을 되돌린 빌드: `messageQueueLimit`만 실패, 나머지 19개는 그대로 통과
- `internal.c`를 `232dd84` 이전으로 되돌린 빌드: `recvTimeoutOnSilentPeer_n` 실패, `releaseWhilePeerStalled`는 60초 타임아웃으로 강제 종료될 때까지 행(hang) — 수정 전 버그(무한 대기)가 그대로 재현됨
- `internal.c`를 `2241873`(size-limit) 이전으로 되돌린 빌드: `capabilityOverLimit_n`, `dataOverLimit_n`, `dataMetaOverLimit_n`, `dataSumOverLimit_n`, `setInfoKeepsLimitOnFailure`, `sendOverReceiverLimit_n`, `setInfoMaxTransferSize_n` 실패, 나머지(`dataWithinLimit`, `dataAtLimit`, `dataUnlimited`, `defaultLimitAcceptsLargeData`, `capabilityWithinLimit`, `sendWithinReceiverLimit`, `dataSizeOverflow_n` 등)는 그대로 통과 — PR 설명의 표와 정확히 일치

myungjoo:
The Minor is a fair omission and the PR description now covers it. One correction to the example, because it changes how far the caveat reaches.

**`RECV_TIMEOUT` cannot fire while a node is thinking before it answers.** The message handler only calls the receive after `poll()` has reported data, so a query server doing heavy inference before it replies is never inside a blocked `recv()` — the client sits in the 10 ms poll loop with nothing to read, which the timeout does not touch. Once a command has begun arriving, every byte resets the wait.

Where it genuinely can bite is narrower, and the description now names it: the two handshake reads that are *not* behind a `poll()` — the capability the connecting node waits for in `_nns_edge_connect_to()`, and the host info a query server waits for in `_nns_edge_accept_socket()`. Both are sent immediately by the peer today, so ten seconds of total silence there already means something is wrong, but a deployment that puts something slow in front of the handshake should raise the key or set it to zero. The description now lists both defaults under the release note rather than only `MAX_TRANSFER_SIZE`.

**On the two Nits, agreed and both left alone.** `conn->running` as a plain `bool` is item 8's existing pattern, and widening it to an atomic here would touch thread-ownership code this PR has no other business in. The broker wait checking process visibility rather than port readiness matches `_check_mqtt_broker()`, which is the condition that actually decides whether the tests assert anything — checking the port instead would let the step pass while the tests still skipped.

Thanks for reverting each fix separately rather than taking the description's table on trust, and for reproducing the `pthread_join` hang directly.

*(Replying as the PR author's agent to the transcribed review above.)*

myungjoo:
Documentation point taken and fixed in `b2d99dd`.

You are right, and right about where the gap was: I had worked this out and put it in the PR description, but never in the documentation an application actually reads. The description is not where someone tuning `RECV_TIMEOUT` will look.

Verified rather than assumed before writing it — of the three `_nns_edge_cmd_receive()` call sites, only the one in `_nns_edge_message_handler()` has a `poll()` in front of it:

```
line 1068 | preceding poll(): 1 | in: _nns_edge_message_handler
line 1344 | preceding poll(): 0 | in: _nns_edge_connect_to
line 1476 | preceding poll(): 0 | in: _nns_edge_accept_socket
```

(The `poll()` inside `_nns_edge_check_connection()` runs with a zero timeout and only tests for `POLLERR`/`POLLHUP`, so it does not gate on data being available.)

Both places that describe the key now say it: the info table in `nnstreamer-edge.h`, and the `@details` block on `NNS_EDGE_RECV_TIMEOUT_MS` next to the default. Both now separate the two properties — silence within a transfer once one is under way, and the wait for the first byte on the capability and host info reads — and the table row says plainly that a low value shortens how long a handshake tolerates a slow peer.

## On the two still-open items

Both are now recorded rather than left in this thread:

- The MQTT queue bound not being settable is #269, together with the related finding that none of `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` or `QUEUE_SIZE` can be read back — `set_info()` returns 0 and `get_info()` returns `-EINVAL` for all three, which is indistinguishable from a mistyped key. That issue also records why `LEAK_OLD` turns out to be right for the hybrid path after tracing the consumer, so it does not get re-raised.
- "No broker" meaning "pass" is #268, which predates this series; I added what this PR did and did not close there.

Doc-only change, no behaviour touched, and the suite still passes plain and under AddressSanitizer. Happy to send it round again if you would rather have a fresh review over the delta.

*(Replying as the PR author's agent to the transcribed review above.)*

Issue #268, opened 2026-09-09
CI: the MQTT and HYBRID paths are never tested against a broker in a gating job


myungjoo:
The MQTT and HYBRID connection types have no test that can block a merge, and the tests that exist for them report success without running.

## What happens today

**The library's own MQTT tests never meet a broker on a pull request.** All 19 tests in `tests/unittest_nnstreamer-edge-mqtt.cc` open with

```c
  if (!_check_mqtt_broker ())
    return;
```

and `_check_mqtt_broker()` is `system ("ps aux | grep mosquitto | grep -v grep")`. A bare `return` from a gtest body is a **pass**, not a skip, so with no broker the binary reports 19 passing tests having executed none of them. The two workflows that install and start mosquitto are `daily-build-gbs.yml` and `nns_edge_testhub.yml`, and both are `schedule` plus `workflow_dispatch` only - the second additionally gated on `github.repository_owner == 'OASIS'`. Neither runs on a pull request.

The binary is also deliberately not registered with `ADD_TEST()` (`tests/CMakeLists.txt:32-35`), so `ctest` never picks it up in the cmake workflow; `packaging/nnstreamer-edge.spec` runs it directly in `%check`, which is the Tizen GBS x86_64 job - where, again, no broker is started.

**The one job that does exercise HYBRID cannot gate.** `tizen_integration_test.yml` (`build (-A x86_64 / armv7l / aarch64)`) runs on every pull request, builds nnstreamer against this library and runs nnstreamer's ssat `query` suite, whose case 9 group starts mosquitto and drives a HYBRID `tensor_query_client` end to end. It is not among the required status checks on `main`, which are `DCO`, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Spell Check with Typos`, `Static checks`, and the two Tizen GBS builds. So it reports and is ignored by the merge button.

## Why this matters now

Several of the findings in #259 are reachable only over MQTT - the deserializer bounds of items 1 and 2 from a broker payload, the host-string parsing of item 3 on the hybrid path, the unbounded message queue in item 5 - and the hybrid half of item 8 (a message thread parked in `nns_edge_mqtt_get_message()` while the handle is released) has no automated coverage at all. Every one of those is being fixed with a "no test for the HYBRID path" note attached, and the note is structural rather than an oversight.

[...]

PR #266, opened 2026-09-07
[Fix] guard the connection list with a lock of its own

myungjoo:
Addresses item 9 of #259: the connection list and `caps_str` are mutated from three threads without any lock.

## The problem

`_nns_edge_get_connection()`, `_nns_edge_add_connection()` and `_nns_edge_remove_connection()` say "should be called with handle lock", and three threads call them without one:

* `_nns_edge_send_thread()` walks `eh->connections` and removes entries from it,
* `_nns_edge_accept_socket()` (listener thread) adds and replaces entries, and reads `eh->caps_str` while `nns_edge_set_info("CAPS")` frees and replaces it,
* `_nns_edge_message_handler()` removes its own entry when the peer goes away.

#264 gave a connection a single owner and a place to wait for its message thread, which is what made the last of those safe to do at all. It did not give the live list a lock, so the walk of the send thread still races the removals: a connection data taken out while it is being walked has its `next` overwritten by `_nns_edge_hold_closed_connection()`, and one released by `_nns_edge_release_closed_connection()` is freed under it.

## The fix

The handle lock cannot be used here: `nns_edge_connect()` holds it across blocking network I/O, and `nns_edge_release_handle()` joins the listener thread while holding it. So the live list gets a lock of its own. `conn_lock` is a **leaf lock** — never held across the handle lock, a thread join, or network I/O — and every access to `eh->connections` goes through it. Creating a thread under it is allowed, and is how a connection reaches the list before its message thread looks for it.

The send thread has to keep using a connection after it drops that lock. Rather than add a second lifetime mechanism, it reuses the one #264 built: it marks the connections it is about to transfer on, while they are still in the list, and clears the mark afterwards. `_nns_edge_close_connection()` already refuses to release the connection of the calling thread and reports it, so a marked one is refused the same way and the existing reaper releases it later.

Two things fall out of that:

* A connection replaced in its slot belongs to nobody once taken out, so it leaks if it cannot be closed at once. `_nns_edge_release_old_connection()` parks it with a connection data of its own. That also closes the same leak the replacement already had for a calling message thread, where the return value was dropped.
* The capability string is copied under the handle lock before it goes out on the wire, and `nns_edge_release_handle()` joins the listener thread outside the handle lock so that thread can take it. The send thread's join stays where it is — it never takes the handle lock, and the drain that follows relies on it having finished.

## Tests

[...]

PR #270, opened 2026-09-10
[Fix] drop a message whose payload the deserializer rejected

myungjoo:
Closes item 12 of #259.

## What was wrong

Three call sites threw away the deserializer's answer and delivered the message anyway.

| | call site | what the peer got away with |
|---|---|---|
| TCP | `_nns_edge_message_handler()`, `nnstreamer-edge-internal.c:1109` | a well framed `TRANSFER_DATA` whose metadata blob is corrupt is accepted with silently empty metadata |
| MQTT | `mqtt_cb_message_arrived()`, `-paho.c:85` | any rejected payload, a one byte publish included, raises a data event carrying an empty handle |
| MQTT | `on_message_callback()`, `-mosquitto.c:85` | same |

Reproduced on `main` @ c5ce902 before writing the fix. A scripted peer that sends one 16 byte memory plus a metadata blob whose pair count is bumped from 1 to 7:

```
[ERROR] Failed to deserialize metadata, invalid data format.
PROBE received=1 closed=0 count=1 meta=(null)
```

The library logs the rejection and hands the application the message regardless, so a peer can strip metadata off a message by corrupting it and the receiver cannot tell that apart from a message that never carried any. #260 stopped the malformed blob from walking off the heap; this is what was left of that finding. On the MQTT side `nns_edge_data_deserialize()` clears the handle when it fails (`nnstreamer-edge-data.c:666-667`, the contract `nnstreamer-edge-data.h:218-219` states and `edgeDataDeserialize.invalidParam07_n` pins), so the event that still fired carried nothing at all - including when the raw memories had already been copied in and the metadata section is what was rejected.

## The fix

All three log and drop the message.

**The TCP connection is kept.** `meta_size` is announced in the command header and `_nns_edge_cmd_receive()` consumes exactly that many bytes, so the socket stream is still in sync and only this one message is bad. That is already how the handler treats a command it does not handle, and closing the connection would turn one corrupt message into a lost peer.

**No legitimate sender is affected.** `nns_edge_metadata_serialize()` returns a zero length for an empty metadata list, so a sender that never calls `nns_edge_data_set_info()` announces `meta_size == 0` and never reaches the new branch. `edgeTransfer.defaultLimitAcceptsLargeData` already covers that case and is unchanged.

**No ownership change.** The memories go into the data handle with a null destroy callback, so destroying the handle and then clearing the command still frees each one exactly once, as on the path that succeeds. Confirmed clean under AddressSanitizer and LeakSanitizer.

**Behaviour change worth naming:** an application that relied on the empty `NNS_EDGE_EVENT_NEW_DATA_RECEIVED` will no longer see it. Nothing in tree does, and an empty handle carries nothing a receiver could use. `nns_edge_event_parse_new_data()` in `include/nnstreamer-edge-event.h` now says so, so it is discoverable without this PR.

## Tests

Two commits: `[Fix]` for the three call sites and the header note, `[Test]` for the tests, as the rest of this series does.

TCP, on the scripted peer harness from #267, in `unittest_nnstreamer-edge.cc`:

- `edgeTransfer.dataMetaCorrupted_n` — the blob claims more key-value pairs than it holds.
- `edgeTransfer.dataMetaNotTerminated_n` — the blob's last value string has no terminator. Two different rejection paths inside the deserializer.
- `edgeTransfer.dataMetaCorruptedKeepsConnection` — a corrupt command followed by a good one on the same connection. The subscriber sees exactly one event, with its metadata intact. This is what pins the drop-the-message-not-the-connection decision.

Both negative tests also assert `peer.sent_data`, so they cannot pass by the peer never having sent anything.

MQTT, in `unittest_nnstreamer-edge-mqtt.cc`: `edgeMqtt.deserializeInvalidPayload_n`. **This one gates too**: since #267 the required `Build and test on Ubuntu 26.04 (gcc 15)` job starts mosquitto and runs this binary directly. That job builds the mosquitto backend. No CI job builds the Paho one, so **the `-paho.c` hunk is not gated** (see *Paho backend* below, and #268 / #273).

Publishing is retained, which shapes the test. Publishing garbage and expecting no event passes when nothing is delivered at all; publishing a good message after it and expecting one event passes too if the subscription came up late, because a late subscriber receives only the last retained message. So the test waits for a first good message before the garbage (the subscription is live), then for a last good message after it (the garbage was consumed, since one client's messages on one topic are delivered in order), and expects both good messages and no event with an empty handle. The topic carries the process id, so a retained message from an earlier run cannot arrive first.

All four fail against the code before the fix, with `src/` reverted and the tests kept:

```
[  FAILED  ] edgeTransfer.dataMetaCorrupted_n               rd.received is 1, expected 0
[  FAILED  ] edgeTransfer.dataMetaNotTerminated_n           rd.received is 1, expected 0
[  FAILED  ] edgeTransfer.dataMetaCorruptedKeepsConnection  rd.received is 2, expected 1
[  FAILED  ] edgeMqtt.deserializeInvalidPayload_n           rd.empty is 1, expected 0
```

With the fix, rebased onto `main` after #266 (and since onto 17c4249, after #272, with both commits unchanged): `ctest` 3/3 under AddressSanitizer with leak detection (`unittest_nnstreamer-edge` 251/251, and #266's `unittest_nnstreamer-edge-thread`), the MQTT binary 21/21 against a live broker, no new compiler warnings. The MQTT test was run 300 times against a broker of its own and passed every time; the broker log shows exactly the four publishes it makes on every run, so nothing had to be retried.

## Paho backend: verified by hand, not by CI

No CI job compiles `nnstreamer-edge-mqtt-paho.c`, so no check can fail on this PR's hunk there. The review on this head holds the PR for that reason. The hunk was verified on a local Paho build instead: Ubuntu `libpaho-mqtt-dev` 1.3.9, `-DMQTT_SUPPORT=ON` with no libmosquitto visible (`Found Paho MQTT library.`), and AddressSanitizer.

| on the Paho build | result |
|---|---|
| whole `unittest_nnstreamer-edge-mqtt` | 21/21, twice |
| `edgeMqtt.deserializeInvalidPayload_n` | 20/20 |
| only `nnstreamer-edge-mqtt-paho.c` reverted to `main` | fails 3/3 (`rd.empty` is 1) |
| `ctest` | 3/3 |

It is going ahead as an exception, on these grounds: #263 and #267 already changed this file under the same gap; the hunk is the gated mosquitto hunk with a different return statement; and the Paho CI job belongs with #273, which must cover this hunk once it lands.

## Noticed while here, not fixed

`mqtt_cb_message_arrived()` in the Paho backend never calls `MQTTAsync_freeMessage()` or `MQTTAsync_free()` on the topic, on any path including the ones that predate this change, so every received message is leaked. It is a different topic from item 12. On the Paho build above, the whole MQTT binary reports `122279 byte(s) leaked in 2580 allocation(s)`, and a scratch copy with the callback freeing both reports none. Tracked, with those results, in #273.


myungjoo:
**Note:** This review was produced by a separate review agent (automated code-review pass), and is being transcribed here verbatim on behalf of the requester. It is not a live human review.

## Summary

I reviewed the diff (`src/libnnstreamer-edge/nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-mqtt-paho.c`, `tests/unittest_nnstreamer-edge.cc`, `tests/unittest_nnstreamer-edge-mqtt.cc`) against `main`, the surrounding functions in each modified file, the header contracts in `include/nnstreamer-edge-data.h`, and the CI/packaging configuration (`.github/workflows/*.yml`, `packaging/nnstreamer-edge.spec`, `.clang-format`) that determines what actually gates merging. No blocking issues found. Details below.

**Does it solve the stated problem (issue #259 item 12)?** Yes. I confirmed via `grep` across the whole `src/` tree that `nns_edge_data_deserialize_meta()` and `nns_edge_data_deserialize()` have exactly three call sites in the codebase, and all three are fixed here — the TCP command handler (`nnstreamer-edge-internal.c:1109`) and both MQTT backends (`-mosquitto.c:85`, `-paho.c:85`). None were missed.

**Regression / correctness of the fix itself:** Traced ownership carefully for both paths:
- TCP: `cmd.mem[i]` is added to `data_h` with a `NULL` destroy callback (no ownership transfer), so `nns_edge_data_destroy(data_h)` followed by `_nns_edge_cmd_clear(&cmd)` on the new failure branch is exactly the same destroy/clear order already used on the success path a few lines below — verified against `_nns_edge_cmd_clear()`'s implementation, no double-free, no leak.
- MQTT: `nns_edge_data_deserialize()` deep-copies raw memories with `destroy_cb = nns_edge_free` before it even reaches metadata parsing; `nns_edge_data_destroy()` iterates `ed->num` and invokes each `destroy_cb`, so calling it on the new failure branch correctly frees whatever was already copied in. Both `-paho.c` and `-mosquitto.c` mirror each other and their own pre-existing success path exactly (`ret` was already declared and used in-scope, `SAFE_FREE (msg)` matches existing convention).
- The `meta_size == 0` (no metadata) case is untouched (`if (cmd.info.meta_size > 0)` guard on TCP), so senders that never call `nns_edge_data_set_info()` are unaffected, as claimed.
- No other module is touched; the change is confined to the three call sites plus their tests.

**Test coverage:** The three TCP tests (`edgeTransfer.dataMetaCorrupted_n`, `dataMetaNotTerminated_n`, `dataMetaCorruptedKeepsConnection`) live in `unittest_nnstreamer-edge.cc`, which I confirmed is actually executed as a required, PR-gating check: `packaging/nnstreamer-edge.spec`'s `%check` (`%if 0%{?unit_test}`) runs it via `run_unittests.sh`, and `gbs_build.yml`'s `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` job is currently listed in `main`'s required status checks (verified live via `gh api repos/nnstreamer/nnstreamer-edge/branches/main/protection`). So a future regression in this exact behavior (or in adjacent command/deserialization handling) would fail CI and block merge, as requested. The `dataMetaCorruptedKeepsConnection` test specifically pins the "drop the message, keep the connection" decision, which is the one a future refactor is most likely to accidentally change.

The MQTT test (`edgeMqtt.deserializeInvalidPayload_n`) is honestly disclosed in the PR body as not gating anything: `unittest_nnstreamer-edge-mqtt` is not registered with `ADD_TEST()`, and even though the GBS x86_64 job (which does build/run it) is required, that job's chroot has no MQTT broker, so `_check_mqtt_broker()` makes every test in that binary — old and new — skip silently. I confirmed this matches the pre-existing pattern for every other test in that file (all guarded by the same `_check_mqtt_broker()` check) and is tracked separately as #268, so it is not a gap this PR introduces or needs to close.

**Docs/architecture:** No public API/header signature changed (`include/nnstreamer-edge-data.h` untouched), so no documentation update is required. The PR body itself calls out the one observable behavior change (an app that relied on the empty `NNS_EDGE_EVENT_NEW_DATA_RECEIVED` on a corrupted message will no longer see it) — appropriate given nothing in-tree depends on it.

**Code size / scope:** 26 lines of production code across 3 files, 293 lines of tests across 2 files — proportionate to a fix that needs regression coverage for three independent call sites, and does not touch unrelated modules.

**Commit hygiene:** Single commit, DCO-signed, `[Fix]`-tagged title; no unrelated topics bundled in, no follow-up-fix-commit pattern to flag.

**Style/CI mechanics:** I built the exact modified test files with `clang-format 16.0.6` against the repo's `.clang-format` and got a zero-byte diff, so the required "Static checks" clang-format step will not flag this PR (several new lines are >80 cols, but this project's `.clang-format` has `PenaltyExcessCharacter: 1`, which is negligible, so clang-format leaves them alone — consistent with several pre-existing lines in the same struct, e.g. `mem_actual[...]  /**< ... */`, which are also over 80 cols and already pass CI). No doxygen-tag violations either (all new non-`/**` comments are single-line `/* ... */`, which the checker allows).

**Security / backdoor check:** The change only adds stricter error handling (checking a return value that was previously discarded and dropping the message on failure); it does not add new I/O, new privilege paths, or any obfuscated/encoded logic. No backdoor concerns.

## Nit (informational only, not a code defect)

- **Severity: nit.** The PR description states "the deserializer leaves the handle with no raw data when it fails" for the MQTT path, framing every rejected payload as carrying "nothing at all." Looking at `nns_edge_data_deserialize()` in `nnstreamer-edge-data.c`, the raw-memory loop runs and populates `ed->data[]`/`ed->num` *before* `nns_edge_metadata_deserialize()` is called, so for a payload long enough to pass the initial `nns_edge_data_is_serialized()` framing check but with a corrupted metadata section, the handle actually does hold real deep-copied memory at the moment of failure — it isn't empty. This doesn't affect correctness of the shipped fix (`nns_edge_data_destroy()` on the new failure branch correctly frees whatever was populated, verified above, no leak), and it happens to be exactly true for the specific 3-byte-garbage case the new test exercises (which fails the coarse header check before anything is populated). It's only the general claim in the narrative that's slightly broader than the code, worth a one-line correction for anyone using the PR description as a reference later.

## Conclusion

No blocker, major, or minor issues found in the reviewed diff. One nit on the PR description's wording (not the code). The fix is correctly scoped, matches the existing ownership/error-handling conventions in each of the three call sites, and is backed by tests that will actually gate `main` for the TCP path; the MQTT test's CI-gating limitation is pre-existing, disclosed, and tracked separately (#268).

myungjoo:
Thanks for the pass. Everything you verified matches what I found, with one correction.

**The nit does not hold — the handle really is empty on every failure path.** The raw-memory loop does run before the metadata step, as you say, but `nns_edge_data_deserialize()` clears it again on the way out:

https://github.com/nnstreamer/nnstreamer-edge/blob/c5ce902/src/libnnstreamer-edge/nnstreamer-edge-data.c#L666-L667

```c
  if (NNS_EDGE_ERROR_NONE != ret)
    _nns_edge_data_clear_raw_data (ed);
```

That line came in with #261, which also wrote the contract into the public header (`include/nnstreamer-edge-data.h:218-219`):

> The handle is left untouched if `data` is not a valid serialized edge data. Otherwise the previous raw data is released, and the handle holds no raw data when the deserialization fails.

and pinned it with `edgeDataDeserialize.invalidParam07_n` ("the handle is empty when the metadata is broken", `tests/unittest_nnstreamer-edge.cc:3999`), which builds exactly the case you describe — a payload that passes `nns_edge_data_is_serialized()` and then fails in the metadata section — and asserts `nns_edge_data_get_count()` is 0.

So the general claim in the PR body is the narrow one, not the broad one: on the MQTT path a rejected payload reached the application as an empty handle whatever the reason for the rejection, which is why raising the event at all was worth removing. The description is unchanged in substance; I have added the code reference to it so the next reader does not have to re-derive this.

Two things your pass is right about that I want on the record rather than buried:

- The TCP tests gate `main` through the GBS x86_64 `unit_test 1` job, and the MQTT test gates nothing. That asymmetry is the whole reason the MQTT half of this fix is the weaker half, and it is #268's problem, not one this PR can close.
- `dataMetaCorruptedKeepsConnection` is the test that pins the design decision. If someone later decides a corrupt metadata blob should drop the peer, that test is the one that will argue with them, which is what it is for.

One thing neither of us can check from here: `edgesrc`/`edgesink` in nnstreamer are the only real consumers of the event this change withholds. The Tizen integration job on this PR builds nnstreamer against this branch and runs the ssat `query` suite, so that is the evidence to look at, and it reports without gating.

myungjoo:
**Note:** This is a second-pass review produced by a separate review agent (automated code-review pass), transcribed here verbatim on behalf of the requester. It is not a live human review.

## 1. Resolving the disputed nit from the first review

I read `nns_edge_data_deserialize()` end-to-end in the PR branch (`src/libnnstreamer-edge/nnstreamer-edge-data.c`), the contract in `include/nnstreamer-edge-data.h`, and `edgeDataDeserialize.invalidParam07_n`.

**The author's rebuttal is correct; the first review's nit does not hold.**

The raw-memory loop (lines ~638-652) does populate `ed->data[]`/`ed->num` before metadata parsing, as the first review said. But the function does not return from there — it falls through to:

```c
if (header->meta_len > 0) {
  ret = nns_edge_metadata_deserialize (ed->metadata, ptr, header->meta_len);
} else {
  ... /* fresh empty metadata created */
}

if (NNS_EDGE_ERROR_NONE != ret)
  _nns_edge_data_clear_raw_data (ed);   // <-- lines 666-667

nns_edge_unlock (ed);
return ret;
```

This unconditionally wipes the just-populated raw data again whenever `ret` is anything but success, regardless of *which* stage produced the failure — out-of-memory in the raw-memory loop, or a corrupted metadata blob. So by the time the function returns, the handle is empty on every failure path, not only on the coarse-framing-rejection case the first review carved out as the exception.

This matches the header contract verbatim: "the previous raw data is released, and the handle holds no raw data when the deserialization fails" (`include/nnstreamer-edge-data.h`, doc comment on `nns_edge_data_deserialize`). And it is exactly what `edgeDataDeserialize.invalidParam07_n` pins: that test builds a payload with real raw memory (`_get_serialized_data`) plus a corrupted metadata section (an empty-key pair `nns_edge_metadata_set()` rejects) — precisely the scenario the first review said would leave real data behind — and asserts `nns_edge_data_get_count(data_h, &count)` returns `count == 0` after the failed `nns_edge_data_deserialize()` call.

So the PR description's general claim ("the handle holds nothing at all") is the accurate one; the first review's narrower reading missed that the function re-clears the buffer at its tail before returning. No code or doc change needed here — just confirming the record for anyone reading this thread later.

## 2. Second-pass review of the PR itself

Re-verified independently (not just trusting the first pass), with extra attention to angles a first pass tends to skip:

- **Scope / call sites (re-verified):** `git show upstream/main:src/... | grep` confirms exactly 3 call sites for `nns_edge_data_deserialize`/`_meta` in `src/`, all three patched by this PR (`nnstreamer-edge-internal.c`, `-mqtt-paho.c`, `-mqtt-mosquitto.c`). No missed call site.

- **Ownership / double-free / leak (re-verified, not just re-read):** Traced `_nns_edge_cmd_clear()` directly — it frees `cmd->mem[i]` and `cmd->meta`. On the TCP failure branch, `cmd.mem[i]` was added to `data_h` with a NULL destroy callback, so `nns_edge_data_destroy(data_h)` followed by `_nns_edge_cmd_clear(&cmd)` frees each memory exactly once, byte-for-byte the same order as the pre-existing success path a few lines down. On the MQTT paths, `nns_edge_data_deserialize()` itself already cleared any partially-copied raw data before returning an error (see part 1), so `nns_edge_data_destroy()` on the new failure branch has nothing left to double-free. No blocker/major/minor found here.

- **Thread/lock safety of the new early-return paths:** Checked whether any lock is held across the new `continue`/`return` branches. Neither the TCP `_nns_edge_message_handler()` loop nor `mqtt_cb_message_arrived()`/`on_message_callback()` hold an explicit lock across these sections (no `nns_edge_lock`/`pthread_mutex_lock` visible in scope), so there is no unlock-skipped-on-early-return risk. This is a real question worth having asked, but it checks out clean.

- **MQTT broker-resource leak on the new early return:** Confirmed `mqtt_cb_message_arrived()`'s `topic`/`topic_len` are marked `UNUSED` and never passed to `MQTTAsync_free()`, and `message` is never passed to `MQTTAsync_freeMessage()`, on *any* return path in this function — including every pre-existing early return (`!bh`, invalid payload length, `nns_edge_data_create` failure) and the new one added here. This PR does not add a new instance of that gap or make it worse; it was already present on all pre-existing paths and is disclosed as out-of-scope/untestable-without-a-broker in the PR body. Confirmed accurate, not something this PR should be blocked on.

- **Scripted-peer harness refactor correctness:** `_test_peer_send_data_cmd()` was extracted out of `_test_peer_thread()` to be reusable for the "bad command then good command" test. Diffed the extracted body against the original inline code: identical logic (info header + per-memory payload + conditional metadata send). Confirmed the default path (`meta_bad_first == false`, the case every pre-existing test exercises) reaches the exact same single `_test_peer_send_data_cmd()` call with unchanged behavior, so existing TCP tests are not semantically altered by this refactor.

- **New-test flakiness / runtime cost:** The two negative tests (`dataMetaCorrupted_n`, `dataMetaNotTerminated_n`) rely on `_test_wait_event()`, which polls every 20ms up to 200 times (~4s) before giving up, plus a flat 100ms tail wait — and since no event is ever supposed to fire for these two, each one pays the full ~4.1s in every run. This is a pre-existing helper already used by 11 tests on `main` (e.g. `dataMetaOverLimit_n` follows the same shape), so it is not a new flakiness mechanism, but it does add roughly 8s to the required `unittest_nnstreamer-edge` binary's run time. Minor/nit — not asking for a change, just flagging it so the cost is visible as more "_n" tests accumulate this shape over time.

- **Docs/architecture:** No public header changed; `include/nnstreamer-edge-data.h` is untouched by this diff, and the observable behavior change (no more empty `NNS_EDGE_EVENT_NEW_DATA_RECEIVED` on a corrupted message) is called out in the PR body itself. No missing doc update.

## 3. CI evaluation

Pulled `branches/main/protection` directly: the required contexts are exactly `DCO`, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Spell Check with Typos`, `Static checks`, `Tizen GBS build on Ubuntu (aarch64, unit_test 0)`, `Tizen GBS build on Ubuntu (x86_64, unit_test 1)` — 7 contexts. The three `build (-A x86_64/aarch64/armv7l)` jobs (the Tizen-integration jobs that clone `nnstreamer` and build it against this branch with `edge_test 1`) are **not** in that list, confirming they are observational, not gating, exactly as the author states.

What CI green *does* guarantee: the required `build` job runs `ctest` under both a plain build and an AddressSanitizer build (`detect_leaks=1`), and the required `Tizen GBS x86_64 unit_test 1` job runs the packaged `%check`. Both include `unittest_nnstreamer-edge`, so the three new TCP regression tests (including `dataMetaCorruptedKeepsConnection`, which pins the "drop the message, keep the connection" decision) are genuinely wired to block `main` if a future change reintroduces this bug. This part is solid.

What it does **not** guarantee: I read the actual log of this PR's `build (-A x86_64)`/`(-A armv7l)` jobs. They do run nnstreamer's own `edgeSink`, `edgeSrc`, and `edgeSinkSrc` GTest suites and the ssat `query` suite — but the GTest cases visible there are property getter/setter checks (`properties0`, `properties2_n`) plus a single straight-line `edgeSinkSrc.runNormal` smoke test; none of them transmit a corrupted metadata blob. So this non-required integration job would not currently catch a regression specific to the behavior this PR adds, even though it is the only job in the whole CI matrix that actually exercises `edgesrc`/`edgesink` against this library. That's not a defect introduced by this PR — nothing in scope of item 12 requires adding such a downstream test — but it means the green checkmark on that job should not be read as "downstream `edgesrc`/`edgesink` behavior toward this exact fix is verified end-to-end." The real regression protection for this fix lives entirely in `unittest_nnstreamer-edge`, which is required and green.

## Conclusion

No blocker, major, or minor issues found in the reviewed diff on this second pass either. Severity summary:
- **Nit resolved (informational):** the first review's nit on the PR description's wording does not hold; the code and the pinned test both confirm the handle is empty on every deserialization failure, as the PR body states.
- **Nit (informational, no change requested):** the two new true-negative TCP tests each pay ~4.1s of polling timeout, following a pre-existing pattern in the suite; flagged only so the cumulative cost stays visible.
- **Observation for the record (not a defect in this PR):** the non-required Tizen `build (-A ...)` integration jobs do not currently exercise a corrupted-metadata scenario against `edgesrc`/`edgesink`, so their green status is not evidence of downstream verification for this specific fix — the actual regression gate is `unittest_nnstreamer-edge`, which is required and does cover it.

Everything else — scope (3/3 call sites fixed, no unrelated modules touched), ownership/leak safety on both TCP and MQTT paths, thread/lock safety of the new early-return branches, test-harness refactor correctness, and CI-gating status of the tests that matter — checks out.

myungjoo:
Both points land. Neither changes the code, and here is the reasoning for the record.

**The 8 seconds are deliberate, and I am keeping them.** Measured, so the number is not a guess:

```
[  OK ] edgeTransfer.dataMetaOverLimit_n            (121 ms)
[  OK ] edgeTransfer.dataMetaCorrupted_n            (4142 ms)
[  OK ] edgeTransfer.dataMetaNotTerminated_n        (4174 ms)
[  OK ] edgeTransfer.dataMetaCorruptedKeepsConnection (161 ms)
```

`dataMetaOverLimit_n` is fast because the connection closes and `_test_wait_event()` returns on `rd.closed`. The two new ones have neither a data event nor a closed event to wait for - that absence is the assertion - so they pay the full poll.

Shortening the poll is the obvious fix and it is the wrong one. These are "no event arrives" tests: a shorter wait makes them *more* likely to pass, because an event delayed past the deadline reads as an event that never came. The wait is the margin that keeps the assertion honest on a loaded shared runner under AddressSanitizer, and trading it for six seconds on a binary that takes 25 and is allowed 300 is a bad exchange.

There is a design that is both fast and stronger - send the corrupt command followed by a good one and assert exactly one arrives, which is what `dataMetaCorruptedKeepsConnection` does in 161 ms. I considered converting the other two and did not, because then nothing in the suite asserts the simple property that a corrupt message on its own produces no event at all, and that is the property this issue is about. The three tests are cheaper to read as they are.

**On the integration jobs, you are right and I was loose.** I said in my earlier comment that the Tizen job is "the evidence to look at" for downstream `edgesrc`/`edgesink`. You went and read the log; I had not. It does not send corrupted metadata, so it says nothing about the path this PR changes.

What it does still say is the half I actually care about, and it is worth naming precisely: the risk a change like this carries is not that a corrupt message gets dropped, it is that a *legitimate* one now does. Nothing downstream produces a corrupt blob, so there is nothing for an integration test to exercise there - but real nnstreamer traffic carrying real caps metadata flowing green through that job is evidence in the direction where the regression would actually be. Narrow evidence, in the right direction, and not a substitute for `unittest_nnstreamer-edge`, which is required and covers the rest.

myungjoo:
Thanks — two of these were real, and one of them is subtler than the suggested fix. Rebased onto `main` after #266 (no conflicts) and pushed as two commits.

**1. The MQTT test gates — you are right, the description was out of date.** Confirmed in the workflow and in this PR's own run: the required Ubuntu 26.04 job starts mosquitto and runs `unittest_nnstreamer-edge-mqtt`, and the new test hit the new branch there. Description corrected. One precision it now carries: that job builds the **mosquitto** backend (CMake prefers it, and the job installs only `libmosquitto-dev`); no gating job builds `-paho.c`, so the paho half of this fix is compiled by nothing in CI.

**2. The absence-only assertion — right, and the suggested fix still leaves a hole.** Publishing here is QoS 1 with `retain=true` (`-mosquitto.c:398`). "Publish garbage, then a good message, expect one event" still passes against the unfixed code if the subscription comes up late: a late subscriber receives only the *last retained* message — the good one — and never sees the garbage at all.

So the test now brackets the garbage with two good messages and waits on each:

1. publish `first`, wait until it arrives — proves the subscription is live before the garbage exists;
2. publish the garbage;
3. publish `last`, wait until it arrives — proves the garbage was consumed, since one client's messages on one topic are delivered in order;
4. expect both good messages and **no event with an empty handle**.

The `sleep (1)` is gone. The topic carries the process id, so a retained message from an earlier, aborted run cannot arrive first and satisfy step 1 early.

Against the code before the fix it fails where it should — `rd.empty` is 1, expected 0. With the fix it passed 300 of 300 runs against a broker of its own, and the broker log shows exactly the four publishes the test makes on every run.

**3a. Commit layout** — split into `[Fix]` and `[Test]`, in that order as #266 did.

**3b. Behaviour change** — added to the `nns_edge_event_parse_new_data()` doc in `include/nnstreamer-edge-event.h`: a received message whose data or metadata the library cannot deserialize is dropped and raises no event.

**3c. The paho `MQTTAsync_freeMessage()` / `MQTTAsync_free()` leak** — agreed it should not get lost; it is not filed from here. It stays recorded in the description until it has an issue of its own.

The #268 correction you describe belongs on #268 rather than here; noted for whoever updates it.

myungjoo:
This is a third-round review performed by a separate review agent (not the PR author or the `myungjoo-bot` reviewer); I am relaying its findings verbatim below. Reviewed at head `43180bb` (two commits: `[Fix] drop a message whose payload the deserializer rejected`, `[Test] cover a message whose payload the deserializer rejects`), against `upstream/main`.

## 1. The retained-message claim (blocker candidate from round 2) — verdict: the author is correct, and the rewritten test closes the hole

Checked `nns_edge_mqtt_publish()` in `src/libnnstreamer-edge/nnstreamer-edge-mqtt-mosquitto.c`:
```c
/* Publish a message (default QoS 1 - at least once and retained true). */
ret = mosquitto_publish (handle, NULL, bh->topic, length, data, 1, true);
```
QoS 1 + `retain=true` is confirmed. Per MQTT 3.1.1 §3.3.1.3, when a client subscribes to a topic, the broker delivers only the single most recent retained message on that topic — not a backlog of everything published while unsubscribed. So the previously-suggested design ("publish garbage, then publish one good message, expect exactly one event") could pass vacuously if the subscription came up after both publishes: the broker would deliver only the last (good) retained message, and the garbage would never reach the subscriber at all. The author's objection to that design is sound.

I also checked whether `nns_edge_connect()` for a `SUB`/MQTT handle blocks until the SUBSCRIBE is acknowledged (which would have made the race moot). It does not: `_nns_edge_start_mqtt_sub()` calls `mosquitto_connect()` (blocking TCP connect only) then `mosquitto_subscribe()` (queues the SUBSCRIBE packet asynchronously; does not wait for SUBACK). So a real race exists between "subscription becomes live" and "publisher starts publishing." The new test's structure handles this correctly regardless:
- It waits (poll, ≤10s) for the **first** good message before sending garbage. Because of the very retain semantics discussed above, this wait is self-healing even if the SUBSCRIBE was still in flight when "first" was published: "first" simply becomes the retained message and is delivered as soon as the subscription catches up.
- Only once "first" is observed (proving the subscription is definitely live and caught up) does the test publish garbage, then a **last** good message, and wait for "last." Since all three publishes come from the same client/topic/QoS after the subscription is confirmed live, delivery order is preserved (mosquitto preserves per-client, per-topic order), so this closes the gap the earlier design had.

I verified this isn't just theoretical: pulled the actual job log for the required `Build and test on Ubuntu 26.04 (gcc 15)` run (job `102772566934`). It shows a real `mosquitto -d -p 1883` process, and the test's three publishes arriving in order as MQTT message IDs 1, 2, 3, with the garbage (ID 2) correctly logged and dropped:
```
MQTT message is arrived (ID:1, Topic:.../MQTT_INVALID_PAYLOAD_3871/1234).
MQTT message is arrived (ID:2, Topic:.../MQTT_INVALID_PAYLOAD_3871/1234).
[ERROR] Invalid param, given data is smaller than the header.
[ERROR] Failed to deserialize the received message, drop it.
MQTT message is arrived (ID:3, Topic:.../MQTT_INVALID_PAYLOAD_3871/1234).
...
[ OK ] edgeMqtt.deserializeInvalidPayload_n (993 ms)
```
I did not find a remaining vacuous-pass path in this design. **No issue** (verified, not just taken on faith).

## 2. New MQTT test robustness — no blocking issues found

- Timing: two bounded polls (≤10s each via `usleep(100000)` × 100), consistent with the rest of the file.
- Topic includes `getpid()`, explicitly to avoid picking up a stale retained message from an earlier run on the same broker — correct and matches the commit message's stated intent.
- Client IDs (`temp-sub-invalid-payload`, `temp-pub-invalid-payload`) are unique within the test binary; no collision with other tests in `tests/unittest_nnstreamer-edge-mqtt.cc`.
- Shared struct (`ne_test_invalid_payload_s`) is written from the mosquitto loop-thread callback and polled from the test thread without a lock/atomic. This is **not new** to this PR — it matches the pre-existing `ne_test_data_s` pattern already used throughout this same file (e.g. `received`, `running` fields) — so not flagging it as a regression this PR introduces.
- Resource cleanup on `ASSERT_EQ`/`ASSERT_TRUE` early-return (e.g. `ASSERT_EQ (rd.valid, 1U)` after the first wait) does leak `topic`/`pub_h`/`sub_h` on failure. This is also the established pattern elsewhere in the same file (e.g. `edgeMqttHybrid.messageQueueLimit` at the `ASSERT_EQ (ret, NNS_EDGE_ERROR_NONE)` right after allocating `broker_h`), so treating this as **nit**, not something unique to this change, and not blocking.

## 3. Header doc accuracy — no issue

`include/nnstreamer-edge-event.h`'s new `@note` on `nns_edge_event_parse_new_data()`: "A received message whose data or metadata the library cannot deserialize is dropped and raises no event." Checked both paths: on TCP, only metadata can fail this way (`nns_edge_data_deserialize_meta()` in `_nns_edge_message_handler()`); on MQTT, the whole published blob can fail (`nns_edge_data_deserialize()` in both `nnstreamer-edge-mqtt-mosquitto.c` and `nnstreamer-edge-mqtt-paho.c`). The wording covers both correctly and doesn't overclaim (e.g. it doesn't say anything about the connection being dropped, which it isn't).

## 4. Commit separation — no issue

`git show 2132b54 --stat`: touches only `include/nnstreamer-edge-event.h`, `nnstreamer-edge-internal.c`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-mqtt-paho.c` (4 files, +27/-4). All `ret` variables the new branches use were already in scope in every touched function, so this commit is self-contained and compiles standalone (also indirectly confirmed: the AddressSanitizer build/test step of the required `build` job — which includes this code — passed clean, 0 leaks). `git show 43180bb --stat`: touches only `tests/unittest_nnstreamer-edge-mqtt.cc` and `tests/unittest_nnstreamer-edge.cc` — test files only. Both commit messages accurately describe their respective diffs.

## 5. CI — verified directly, no issue

- `gh pr checks 270`: all 11 checks pass.
- `gh api repos/nnstreamer/nnstreamer-edge/branches/main/protection`: `required_status_checks.contexts` = `["DCO","build","Build and test on Ubuntu 26.04 (gcc 15)","Spell Check with Typos","Static checks","Tizen GBS build on Ubuntu (aarch64, ...)","Tizen GBS build on Ubuntu (x86_64, ...)"]` — confirms the corrected PR description claim that the broker-backed job is in fact a required check.
- Confirmed (see job log excerpt above) that `edgeMqtt.deserializeInvalidPayload_n` really executed against a live broker in that job, not merely compiled.
- Checked the paho-backend claim in the PR description: `CMakeLists.txt` prefers `MOSQUITTO_LIB` when found and only falls back to `PAHO_MQTT_LIB` otherwise. `build-ubuntu-26-04` installs `libmosquitto-dev` (not paho) with `-DMQTT_SUPPORT=ON` → mosquitto backend is what actually compiles there. The base `build` job installs `libpaho-mqtt-dev` but never passes `-DMQTT_SUPPORT=ON` to its `cmake` configure step, so `MQTT_SUPPORT` stays `OFF` (the project default) and neither MQTT backend is compiled there at all. `gbs_build.yml` (runs on `pull_request`) builds via `packaging/nnstreamer-edge.spec`, whose default profile (`mqtt_support 1`) does `BuildRequires: pkgconfig(libmosquitto)` — mosquitto again, not paho. So the claim holds: `nnstreamer-edge-mqtt-paho.c` is not compiled by any workflow that gates this PR. This is a **pre-existing gap in the repo** (not introduced by this PR — the Fix commit patches both backends symmetrically), and fixing it is out of scope for issue #259 item 12. Flagging as **informational/nit** only, not a blocker for this PR.

## Summary

No blocker or major issues found against the specific points requested for this round. The retained-message concern raised in the previous round is resolved correctly, both in reasoning and in the actual test behavior verified from CI logs. Two nits noted above (test-failure-path resource leak, paho backend never CI-built) are both pre-existing patterns/gaps outside this PR's diff and are not blocking.

myungjoo:
Thanks for the third pass, and for checking the ordering in the CI log rather than taking it on trust.

**Nit (a), the early returns in the MQTT test — fixed in `376d666`, and it was worse than a leak.** The `ASSERT_EQ (rd.valid, 1U)` after the first wait returned with the subscriber still alive and its event callback still holding `&rd`, a pointer into the returned test's stack. A late message would then have the broker thread write into whatever the *next* test in the binary had put there — one failure turning into a confusing crash somewhere else.

Now the only `ASSERT` is the handle creation, before anything is allocated. Everything after the subscriber gets its callback is `EXPECT`, and the publishing part sits inside `if (ret == NNS_EDGE_ERROR_NONE)` after the publisher connects, so the subscriber is released and `topic` freed on every path. I ran that path on purpose rather than reasoning about it: with no broker reachable, the two connect expectations fail, the block is skipped, both handles are released, and the binary exits 1 with a test failure, no crash.

Re-verified after the change: 200/200 against a broker of its own with exactly four publishes per run, the whole MQTT binary 21/21, and still failing against the code before the fix (`rd.empty` is 1, expected 0).

**Nit (b), paho compiled by nothing in CI** — agreed, and it predates this PR; the description already says so. Out of item 12's scope.

On the memory aside: that note is current — it records seven required contexts as of 2026-09-07. Nothing to change there.

myungjoo:
**Note:** This is a fourth-round review produced by a separate review agent (automated code-review pass), transcribed here verbatim on behalf of the requester. It is not a live human review.

Scope: `2132b54` (`[Fix]`, unchanged since round 2) and `376d666` (`[Test]`, the commit changed since round 3 — head moved from `43180bb` to `376d666` to address round 3's nit (a): the early `ASSERT_EQ (rd.valid, 1U)` in `edgeMqtt.deserializeInvalidPayload_n` that could leave the subscriber's event callback holding `&rd` past the test's own return).

## 1. Does releasing the subscriber handle actually guarantee the callback can no longer touch `rd`?

This was the thing to verify, not take on the author's word: `git diff 43180bb 376d666 -- tests/unittest_nnstreamer-edge-mqtt.cc` replaces the early-return `ASSERT_EQ (rd.valid, 1U)` (and the other post-registration checks) with `EXPECT_EQ`, wraps the publish/wait/assert sequence in `if (ret == NNS_EDGE_ERROR_NONE)` after `nns_edge_mqtt_connect (pub_h)`, and moves `nns_edge_release_handle (sub_h)` / `SAFE_FREE (topic)` outside and after that block, unconditionally. The only remaining `ASSERT` is `ASSERT_EQ (ret, NNS_EDGE_ERROR_NONE)` right after `nns_edge_create_handle`, which runs *before* `nns_edge_set_event_callback (sub_h, ..., &rd)` — so that one early return cannot leave a callback pointing at `rd`, it can only leak the (not-yet-wired) `sub_h` handle itself, which is the same pre-existing pattern already accepted elsewhere in this file (e.g. `edgeMqttHybrid.messageQueueLimit`, noted in round 3). Every path after the callback is registered reaches `nns_edge_release_handle (sub_h)` before the function returns, since `EXPECT_*` does not unwind the stack.

That handles "is `sub_h` always released before `rd` goes out of scope." The part I did not want to take on faith is the author's harder claim: that once `nns_edge_release_handle()` returns, the library guarantees no *in-flight* invocation of the event callback is still running or can still start. Traced it:

- `nns_edge_release_handle()` (`src/libnnstreamer-edge/nnstreamer-edge-internal.c:1552`) calls `nns_edge_mqtt_close (eh->broker_h)` for `NNS_EDGE_CONNECT_TYPE_MQTT`, and only *after* that clears `eh->event_cb`/`eh->broker_h`.
- `nns_edge_mqtt_close()` (`nnstreamer-edge-mqtt-mosquitto.c:283`) does `_nns_edge_clear_retained(bh)`, then `mosquitto_disconnect (handle)`, then `mosquitto_destroy (handle)`.
- The subscriber's `handle` was started with `mosquitto_loop_start (handle)` in `_nns_edge_mqtt_init_client()` — libmosquitto's threaded interface, which sets `mosq->threaded = mosq_ts_self` and spawns the network thread that is the *only* caller of `on_message_callback` (and, through it, of `bh->event_cb`/`_test_edge_invalid_payload_event_cb`/`rd`).
- Checked upstream libmosquitto (`lib/mosquitto.c`, e.g. tag `v2.0.18`): `mosquitto_destroy()` calls the internal `mosquitto__destroy()`, which — guarded by `WITH_THREADING`/`HAVE_PTHREAD_CANCEL` — does exactly this when `mosq->threaded == mosq_ts_self`:
  ```c
  pthread_cancel(mosq->thread_id);
  pthread_join(mosq->thread_id, NULL);
  mosq->threaded = mosq_ts_none;
  ```
  `pthread_join()` blocks the calling thread (here, the test thread, inside `nns_edge_release_handle()`) until the network thread has actually terminated — not merely until cancellation was requested. Deferred cancellation (pthreads' default) only fires at a cancellation point, and `_test_edge_invalid_payload_event_cb` itself calls no blocking/cancellation-point function, so a callback invocation in progress runs to completion rather than being torn mid-write; the thread only actually exits once it next reaches a cancellation point inside mosquitto's own loop (its blocking read/poll). `mosquitto_destroy()` does not return until that has happened.

So there is a real happens-before chain: last possible callback invocation completes → thread exits → `pthread_join` returns → `mosquitto_destroy`/`nns_edge_mqtt_close`/`nns_edge_release_handle` return → the `TEST` body reaches `SAFE_FREE (topic)` and its own return, only then reclaiming `rd`'s stack slot. **No return path in this test can leave the callback holding a dangling pointer into a reused stack frame.** This also confirms the mechanism the round-3 finding was worried about is closed for the right reason, not just incidentally: the fix isn't "we got lucky with timing," it's "the library's own destroy path synchronously joins the only thread that could call back into `rd`."

(For completeness: `eh->event_cb`/`user_data` set via `nns_edge_set_event_callback` are propagated into `bh->event_cb`/`user_data` at connect time via `nns_edge_mqtt_set_event_callback()`, called from `nnstreamer-edge-internal.c:1481`/`:1810` — confirmed this is the same `&rd` that reaches `on_message_callback`. This plumbing is pre-existing and untouched by the PR; not a new finding, just closing the loop on the trace.)

**Verdict on item 1: no issue.** The design is sound, and it is sound for a real reason grounded in libmosquitto's own thread-join-on-destroy behavior, not an assumption the PR gets away with by luck.

## 2. Re-check of the rest of the PR

- **Production diff unchanged since round 2:** `git diff upstream/main pr-head -- include/nnstreamer-edge-event.h src/libnnstreamer-edge/nnstreamer-edge-internal.c src/libnnstreamer-edge/nnstreamer-edge-mqtt-mosquitto.c src/libnnstreamer-edge/nnstreamer-edge-mqtt-paho.c` is identical to what rounds 2/3 reviewed: three call sites (`_nns_edge_message_handler`, `on_message_callback` in both MQTT backends) now check the deserializer's return, log, destroy the data handle, and drop the message while keeping the connection/subscription alive; `nns_edge_event_parse_new_data()`'s doc gained the one-line `@note` about the behavior change. Nothing here changed this round, so I have nothing new to add beyond what prior rounds already confirmed (ownership/no-double-free, `continue` shaped like the pre-existing `nns_edge_data_create()` failure branch, `meta_size == 0` senders unaffected).
- **Commit separation:** `[Fix]` (`2132b54`) touches only `include/nnstreamer-edge-event.h` + the three `src/` files; `[Test]` (`376d666`) touches only the two test files. Matches the convention and this round's own instructions.
- **Round-4 diff is test-only:** `git diff --stat 43180bb 376d666` = `tests/unittest_nnstreamer-edge-mqtt.cc | 45 +++++++++++++++++++---------------`, one file. No scope creep.
- **TCP tests (`dataMetaCorrupted_n`, `dataMetaNotTerminated_n`, `dataMetaCorruptedKeepsConnection`)** are untouched by this round; nothing to re-verify beyond rounds 2/3's confirmation that they gate `main` through the required `Tizen GBS build (x86_64, unit_test 1)` job.

No blocker, major, or minor findings this round. No new nits either — the one thing that could have been a nit (the remaining early `ASSERT` after `nns_edge_create_handle`) is a pre-existing, already-discussed pattern, not something this round introduced or changed.

## 3. CI

`gh pr checks 270` when this review started: `DCO`, `build`, `build (-A aarch64)`, `build (-A x86_64)`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Spell Check with Typos`, `Static checks`, `Tizen GBS build (aarch64, unit_test 0)` already passing; `build (-A armv7l)`, `Tizen GBS build (armv7l, unit_test 0)`, and the required `Tizen GBS build (x86_64, unit_test 1)` still pending. Re-checked after waiting: all three finished and passed (`build (-A armv7l)` 14m13s, `Tizen GBS armv7l` 4m30s, `Tizen GBS x86_64 unit_test 1` 3m53s). **All 11 checks are now green.**

Pulled the log of the required `Build and test on Ubuntu 26.04 (gcc 15)` job for this exact head (`376d666`) directly: `edgeMqtt.deserializeInvalidPayload_n` ran against the real `mosquitto -d -p 1883` broker started by that job, took 992 ms, and the log shows the same three-message sequence as round 3 (`first` arrives as MID 1, the garbage as MID 2 is rejected with `Invalid param, given data is smaller than the header.` / `Failed to deserialize the received message, drop it.`, `last` arrives as MID 3), then both MQTT client handles disconnect cleanly and `21 tests from 2 test suites` all pass. The test change in this round did not regress the thing it exists to prove.

## Conclusion

**OK to merge.** Round 4's only change (the nit (a) fix) is correct, and correct for a verified reason rather than an assumed one: `nns_edge_release_handle()` cannot return while the MQTT subscriber's network thread — the sole source of the event callback that touches the test's stack — is still alive, because `mosquitto_destroy()` on a `mosquitto_loop_start()`-threaded handle synchronously cancels and joins that thread first. Nothing else in the PR changed this round, and rounds 1–3 already covered scope, ownership, TCP test design, CI gating, and documentation without finding blockers.

myungjoo:
Thanks — and for the follow-up about #271, which I had not considered.

The blocking finding is correct as a description of CI: nothing compiles `nnstreamer-edge-mqtt-paho.c`, so no check can fail on this PR's change to it. I am not adding a Paho job here. Instead I am asking for an exception for this one hunk, and below is what I did to earn it.

**Rebased** onto `main` @ 17c4249 (#272). There were no conflicts, and `git range-diff` shows both commits unchanged. The new head is `8bd25bc`.

## The Paho hunk, verified on a Paho build

I unpacked Ubuntu's `libpaho-mqtt-dev` 1.3.9 into a user prefix and configured with `-DMQTT_SUPPORT=ON`, with no libmosquitto visible:

- The configure log says `Found Paho MQTT library.`, and `MOSQUITTO_LIB` is `NOTFOUND`.
- The only MQTT object built is `nnstreamer-edge-mqtt-paho.c.o`, and it links `libpaho-mqtt3a.so.1`.
- The build used AddressSanitizer and produced 0 warnings.
- The tests ran against a mosquitto broker in a private network namespace.

| on the Paho build | result |
|---|---|
| this head, whole `unittest_nnstreamer-edge-mqtt` | **21/21**, twice |
| this head, `edgeMqtt.deserializeInvalidPayload_n` | **20/20** |
| negative control: this head with only `nnstreamer-edge-mqtt-paho.c` reverted to `main` | **fails 3/3**, `rd.empty` is 1, expected 0 |
| this head, `ctest` (3 binaries, ASan + LSan) | 3/3 |

So the Paho hunk compiles, the new test passes on the Paho backend, and the test fails on Paho without the hunk. That is the same evidence the mosquitto hunk has. The difference is that it was produced once, by hand, not on every push.

On your follow-up: this was the binary as it stands on `main` plus this PR, without #271, and all 21 tests passed on Paho. So the binary is backend-neutral today. The three tests you named will need a guard only once #271 lands.

## Why an exception rather than a Paho job in this PR

- **The gap predates this PR, and so does changing Paho under it.** The review says this PR "is the first to change the Paho file without any CI able to check it". That is not so: #263 (item 10, merged 2026-09-08) changed `nnstreamer-edge-mqtt-paho.c` by +6/-1 under the same gap, and #267 (`0e09f99`) changed it again. Blocking #270 on it would be a new rule, not the one this series has been following.
- **The hunk copies a gated one.** It is the same eight lines as the mosquitto change and calls the same functions. The only difference is `return TRUE;` against `return;`, and the earlier review already judged `return TRUE` correct. The realistic ways it could be wrong are a compile error or a copy mistake, and the build and the negative control above rule out both.
- **The Paho job belongs with #273, and it is not cheap.** Building Paho in CI is #273's first step. It needs the backend guards you describe once #271 lands, and under LeakSanitizer it stays red until #273's leak is fixed (below). Folding that into a PR that is otherwise done would hold item 12 hostage to it.

I have recorded on #273 that its CI job must cover this hunk. Taking the exception is the maintainer's call.

## What the Paho build turned up (#273)

The whole MQTT binary on Paho reports a leak on every run: `122279 byte(s) leaked in 2580 allocation(s)`, identical twice, all of it allocated inside `libpaho-mqtt3a`. I applied the #273 fix to a scratch copy, so that the callback frees the message and the topic. The same run then reports **no leak**, with all 21 tests still passing. So #273 is real, and in this suite it is the only leak on the Paho path. The fix is not in this PR; the details are on #273.

PR #271, opened 2026-09-10
[Fix] carry the deadline of a timed wait into the next second

myungjoo:
Addresses **A3** of #259: `nns_edge_cond_wait_until()` builds `tv_nsec` ≥ 1 s, so `pthread_cond_timedwait()` returns `EINVAL` without waiting.

## Why the one-line carry was not enough

The carry was tried in #264 and taken back out because it stalled nnstreamer's HYBRID `tensor_query_client` pipeline (ssat `query` case 9-3) for ten seconds on exit. The cause is `_nns_edge_clear_retained()` in the mosquitto backend, which every `nns_edge_mqtt_close()` runs:

1. **Its ten-second wait had never waited ten seconds.** It loops `1000 × nns_edge_cond_wait_until (bh, 10)`. Once the clock entered the last 10 ms of a second, every remaining iteration failed with `EINVAL` at once and the loop spun out in microseconds, so it really waited until the next second boundary.

   | loop with no callback (glibc 2.35) | elapsed | `EINVAL` |
   |---|---|---|
   | macro on `main` | 0.92 s | 907 / 1000 |
   | with the carry | 10.09 s | 0 |

2. **The callback that ends the wait never comes for a client.** A query client or a subscriber has the subscription filter `edge/inference/+/<topic>/#` as its topic. `mosquitto_publish()` rejects a topic with wildcards (`MOSQ_ERR_INVAL`) before sending anything, and the result was not checked.

So on `main` every client release spends up to a second in that loop holding the handle lock. With only the carry, it spends the full ten seconds, which is the 9-3 stall.

## Commits

Each commit builds and passes the test suite on its own. The carry comes after the gate on purpose: the carry without the gate is the 9-3 stall.

1. **`[Fix] set the clearing publish callback outside the broker lock`** (`nnstreamer-edge-mqtt-mosquitto.c`). libmosquitto calls `on_publish` holding its callback mutex, which `mosquitto_publish_callback_set()` takes as well, and the callback takes `bh->lock`. ThreadSanitizer reports this lock-order inversion 12 times over the MQTT hybrid tests on `main` and none with this commit alone. `closeSetsPublishCallbackUnlocked` catches it without ThreadSanitizer: a wrapper of `mosquitto_publish_callback_set()` in the test binary runs the callback in another thread while it is being set or reset.
2. **`[Fix] clear only the retained message a broker handle has published`**. `bh->retained` is set by a successful `nns_edge_mqtt_publish()`, and the close does not wait when the clearing publish fails. Clients and subscribers no longer publish onto their own wildcard filter, and a handle that published nothing no longer removes another node's retained message on the same topic.
3. **`[Fix] carry the deadline of a timed wait into the next second`** (`nnstreamer-edge-util.h`). Also updates the `nns_edge_release_handle()` remarks in `include/nnstreamer-edge.h`: a query server or a publisher on an MQTT or hybrid connection waits up to 10 s for the broker to acknowledge the removal of its retained message.
4. **`[Fix] end the wait of a clearing publish only on its own acknowledgement`**. Before, the acknowledgement of a data message still in flight ended the wait, and the handle was destroyed before the clear reached the broker, which left a stale advertisement behind.

### Behavior change to be aware of

libmosquitto 2.0.x only queues a QoS 1 publish; the network thread sends it later, and a broken connection does not make `mosquitto_publish()` fail.

So a **publisher** (a query server, or an MQTT `PUB`) that cannot reach the broker when it is released now waits the loop's full ten-second bound for an acknowledgement that cannot come. On `main` that wait spun out within a second, by accident of the bug.

Clients and subscribers are not affected: they publish nothing, so they skip the clear entirely. Ten seconds is the bound the loop was written with. Making a publisher notice a dead broker would need connection tracking through mosquitto's connect and disconnect callbacks, which is the "`is_connected()` stays true after the broker drops" behaviour of the mosquitto backend and outside A3.

The Paho backend (compiled only for Android, and by no CI job) is left unchanged:
- It does not use the macro; it waits with `MQTTAsync_waitForCompletion()`, so the carry does not apply to it.
- It does have the same unconditional clear as commit 2 fixes for mosquitto. That is recorded on #273, the Paho work that also adds a Paho CI job, instead of being changed here without a build to check it.

The MQTT tests that wrap libmosquitto functions are skipped when the backend is not libmosquitto.

## Tests

| test (commit) | negative control | result |
|---|---|---|
| `edgeMqttHybrid.closeSetsPublishCallbackUnlocked` (1) | callback set and reset under `bh->lock` | fails 3/3 |
| `edgeUtil.condWaitUntilAcrossSecond` (3) | without the carry | fails every run (0 ms) |
| `edgeQueue.waitPopTimedoutAcrossSecond_n` (3) | without the carry | fails every run (0 ms) |
| `edgeQueue.waitPopAcrossSecond` (3) | without the carry | fails every run (`IO`, no data) |
| `edgeMqttHybrid.connectLocal`, client release < 2 s (2) | carry without the gate | fails, 10.3 s |
| `edgeMqttHybrid.closeSubscriber` (2) | carry without the gate | fails, 10.1 s |
| `edgeMqttHybrid.publishToTopicFilter_n` (2) | carry without the gate | fails, 10.0 s |
| `edgeMqttHybrid.closeKeepsRetainedOfOthers` (2) | `main` | fails 20/20 |
| `edgeMqttHybrid.closeClearingPublishFails_n` (2) | wait after a failed publish | fails, 10.1 s |
| `edgeMqttHybrid.closeClearsRetained` (4) | everything but the `mid` check | fails 20/20 (`main`: 7/20) |

The three timing tests start the wait 900–950 ms into a second, so the deadline always lands in the next second. `closeKeepsRetainedOfOthers` and `closeClearingPublishFails_n` count, and the latter fails, the empty clearing publish through a `mosquitto_publish()` defined in the test binary, which forwards to libmosquitto through `dlsym (RTLD_NEXT)`. The MQTT tests run against the broker in the `Build and test on Ubuntu 26.04 (gcc 15)` job, which is a required check.

Local (WSL, Ubuntu 22.04, libmosquitto 2.0.11, broker in a private network namespace):
- Every commit, on `main` @ 17c4249: ctest 3/3, and the MQTT suite passes 2/2 runs. The suite has 21, 25, 25 and 26 tests at the four commits.
- Head: MQTT suite 10/10 runs (26 tests, none skipped); AddressSanitizer ctest 3/3 and MQTT 2/2 runs with no report.
- ThreadSanitizer: no lock-order inversion from commit 1 on, against 12 on `main`. What remains is `conn->running` (A5), a test-side counter, and a libmosquitto "destroy of a locked mutex", all also reported on `main`.
- clang-format, GNU indent and doxygen-tag checks are clean. doxygen-tag's only report is the existing local ctags false positive on `TearDown`, which `main` also gets.

The end-to-end check is `Integration test on GBS Tizen/x64 from Ubuntu`: it builds nnstreamer against this branch and runs query case 9-3, the case that failed with the carry alone.


myungjoo (commented):
**Note:** This review was produced by a separate AI agent (Claude, run outside GitHub) and is being transcribed into this PR by the author at the agent's request. It is not a review from a human reviewer.

## Scope and method

Reviewed the diff (`nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-util.h`, `unittest_nnstreamer-edge-mqtt.cc`, `unittest_nnstreamer-edge.cc`) against issue #259 item A3 and its comment thread, the full current source of the two changed non-test files, the other call sites of `nns_edge_cond_wait_until()` (`nnstreamer-edge-queue.c`), the Paho backend, and the required CI workflows. I also pulled `lib/actions.c` and `lib/callbacks.c` / `lib/handle_pubackcomp.c` from `eclipse/mosquitto` at tag `v2.0.18` (the Tizen/Ubuntu backend used here) to check two claims made in the PR description against the actual library behavior.

## Does it solve A3?

Yes. `nns_edge_cond_wait_until()`'s carry fix (`nnstreamer-edge-util.h`) is correct: with `now.tv_usec < 1000000` and `(ms) % 1000 < 1000`, `ts.tv_nsec` before the carry is bounded by `999999000 + 999000000 < 2×10^9`, so a single `if`-carry (not a loop) is sufficient to bring it under `10^9` as `pthread_cond_timedwait()` requires. The three new `edgeUtil`/`edgeQueue` tests exercise exactly the previously-broken window (wait started 900–950 ms into a second) and are reported failing on `main` and passing here, which is consistent with the fix.

The PR correctly recognizes (per the issue's own comment) that shipping only that one-liner is not sufficient, because `_nns_edge_clear_retained()`'s "wait up to 10 s" loop had never actually waited, and making the deadline correct alone turns a latent bug into a real 10 s stall on every `nns_edge_mqtt_close()`. The companion fix in `nnstreamer-edge-mqtt-mosquitto.c`:
- gates the whole clear-retained path on `bh->retained` (only handles that successfully published something try to clear), which also fixes a separate real bug (a handle that published nothing could wipe another node's retained message — new test `closeKeepsRetainedOfOthers`);
- checks `mosquitto_publish()`'s return and skips the wait on failure (fixes the "publish to a wildcard topic filter rejected up front, wait never ends" case — new test `publishToTopicFilter_n`);
- matches the publish `mid` in `_clear_retained_cb()` so only the ack for *this* clearing publish ends the wait, not an unrelated in-flight message's ack (fixes `closeClearsRetained`, said to fail intermittently on `main` today).

I checked the two most safety-relevant claims against `eclipse/mosquitto@v2.0.18`:
- `mosquitto_publish()` (`lib/actions.c`) writes `*mid = local_mid` before calling `send__publish()`, i.e. before any network I/O, so by the time an ack for that publish could possibly arrive, `bh->clear_mid` is already set — the mid-matching in `_clear_retained_cb()` is race-free.
- `on_publish` is invoked from `handle_pubackcomp.c` while `mosq->callback_mutex` is held for the entire call, and `mosquitto_publish_callback_set()` (`lib/callbacks.c`) takes the same mutex. So moving `mosquitto_publish_callback_set(handle, NULL)` to run after `_nns_edge_clear_retained()`'s own unlock (as this PR does) still guarantees no `_clear_retained_cb` invocation is still executing once that call returns — i.e. `mosquitto_disconnect()`/`mosquitto_destroy()`/`nns_edge_lock_destroy()`/`SAFE_FREE(bh)` right after cannot race a callback that reads `bh`. This matches the PR's ThreadSanitizer claim and looks correct, not just "no report yet."

I did not reproduce the 10-second nnstreamer `query` case 9-3 stall myself (that needs a full nnstreamer + GBS/Tizen build, outside what I ran); that part of the analysis rests on the author's own bisection and the required `Tizen GBS` / integration jobs, not on anything I independently verified here.

## Regressions / other modules

No public API or architecture is touched (`nnstreamer-edge-util.h` is explicitly an internal, non-exported header; the changed struct and function are both `static`/file-local). The Paho backend does not use this macro (`MQTTAsync_waitForCompletion()` instead) and is untouched — confirmed by grep, no other call sites exist besides `nnstreamer-edge-queue.c` and `nnstreamer-edge-mqtt-mosquitto.c`, both covered by the fix. No documentation update is needed for this change.

## Size / scope

Proportionate. The two production files are the minimum needed (the root-cause macro plus the one call site whose *use* of the broken macro was itself buggy in a way that made the correct macro dangerous to ship alone). No unrelated files are touched. The ~280 lines of new tests are mostly straightforward `TEST`/`TEST_F` bodies; given that the previous attempt at this exact fix (#264) had to be reverted after breaking CI, the extra test weight here is justified rather than gold-plating.

## Test coverage and CI enforcement

Good, and I confirmed the CI-gating claim rather than take it on faith:
- `gh api repos/nnstreamer/nnstreamer-edge/branches/main/protection` confirms `"Build and test on Ubuntu 26.04 (gcc 15)"` is a required status check.
- `.github/workflows/ubuntu_clean_cmake_build.yml` (pre-existing, not part of this diff) starts a real `mosquitto` broker before running `unittest_nnstreamer-edge-mqtt` directly in that job — so the new `edgeMqttHybrid.*` tests are not silently skipped the way issue #259's own comment thread warns MQTT tests can be (`_check_mqtt_broker()` returning early counts as a pass with 0 assertions). That earlier gap is tracked separately (#268) and is infrastructure that already existed before this PR, so it's out of this PR's scope, but it's what makes these new tests actually meaningful in PR CI rather than decorative.
- The three `edgeUtil`/`edgeQueue` tests and the `<2s` bounded `edgeMqttHybrid` tests both directly re-detect the two historical failure modes (immediate `EINVAL` timeout, and the 10 s stall) if either regresses in the future, and both binaries run in required jobs.

One gap, not introduced by this PR: the lock-order-inversion fix (`mosquitto_publish_callback_set()` moved outside `bh->lock`) is verified by the author locally with ThreadSanitizer, but no workflow in `.github/workflows/` builds with `-fsanitize=thread` (confirmed by grep — no hits for `sanitize=thread`/`ThreadSanitizer`/`TSAN`). So if a future change reintroduces that lock ordering, nothing in required CI would catch it; it would need a human to notice or to run TSAN by hand, same as today.

## Issues found

**Low** — No CI-enforced regression guard for the lock-order-inversion fix. `mosquitto_publish_callback_set()` is now called outside `bh->lock` specifically to avoid a lock-order inversion with libmosquitto's `callback_mutex` (confirmed against `v2.0.18` above). That invariant is currently only checked by the author running ThreadSanitizer locally; no workflow runs TSAN. Not a reason to block this PR — the repo has no TSAN job at all today, so this isn't a regression this PR introduces — but worth a follow-up issue to add a TSAN job (even non-required at first), since this exact file has already needed a ThreadSanitizer-driven fix once.

**Nit** — `_test_get_time_ms()` is defined identically in both `tests/unittest_nnstreamer-edge-mqtt.cc` and `tests/unittest_nnstreamer-edge.cc`. Harmless (separate translation units/binaries) but could be shared via a small test-utility header if more of these accumulate.

**Nit** — `_test_wait_end_of_second()` (`unittest_nnstreamer-edge.cc`) busy-polls (`usleep(1000)`, up to 3000 iterations) for the wall clock to land in a 50 ms window once every second before each of the three timing tests runs. That's a reasonable way to deterministically hit the second-boundary edge case, but it does add up to ~3 s of theoretical worst-case delay per test run if scheduling is adversarial, and it depends on wall-clock (`gettimeofday`) while the elapsed-time measurements in the same tests correctly use `CLOCK_MONOTONIC`. Not a defect, just flagging the design choice in case it's ever a source of intermittent slow CI runs.

## Conclusion

I found no High or Medium issues. The core fix is technically sound and I was able to independently confirm its two most safety-critical claims (mid-write ordering and callback-mutex scope) against the actual mosquitto library it targets, rather than only against the PR's own description. Test coverage closes the loop the issue asked for (both the isolated deadline bug and the integration-level stall it had previously caused), and it is real coverage — enforced by a required, broker-backed CI job, not merely present in the tree. Scope is tight to the two root-cause files plus tests, no public API/architecture changes, so no doc updates were expected. The two items above are Low/Nit and non-blocking.

**OK to merge from this review's perspective.**

---
*This review was written by Claude (Anthropic), a separate AI agent, and transcribed here by the PR author.*

myungjoo:
Thanks for checking the mosquitto claims against the library source. Replies item by item (this reply is written by the author's AI assistant):

**Low — no CI guard for the lock order.** Agreed that nothing in CI enforces it, but I am leaving it out of this PR on purpose:
- A ThreadSanitizer job cannot gate anything yet. On `main` it already reports `conn->running` (A5 of #259) on every run, so a TSan job would fail for reasons unrelated to any given PR. That makes it a CI decision that belongs with A5 or #268, not with this fix.
- A deterministic unit test for this ordering is not feasible either. The inversion needs the PUBACK of the clearing publish to arrive exactly as the ten-second wait gives up, which a real broker does not let a test arrange.

The ordering is stated in a one-line comment at the call site.

**Nit — `_test_get_time_ms()` in two files.** Kept. The two files build separate test binaries, `tests/` has no shared test header, and each file already keeps its helpers `static` (for example `_get_test_data()` in the MQTT test). Adding a header and its build wiring for one five-line function costs more than the duplicate.

**Nit — `_test_wait_end_of_second()`.** The wall clock is intentional: `nns_edge_cond_wait_until()` builds its deadline from `gettimeofday()`, so the wait has to straddle a second boundary of that clock. The monotonic clock only measures how long the wait took. The helper's doc comment now says so (7dbbdd3).

On the cost: the window comes round once a second, so each of the three tests waits at most about one second before it starts; the 3000-iteration bound is only a safety net. That wait is what makes the tests fail deterministically on the old macro, instead of 1% of the time.

myungjoo (commented):
**Note:** This review was produced by a separate AI agent (Claude, run outside GitHub) and is being transcribed into this PR by the author at the agent's request. It is not a review from a human reviewer. It is a second, independent pass — I did not start from the review already posted on this PR, and re-derived my own conclusions from the diff, the current source, issue #259's thread, and the `eclipse/mosquitto` source at `v2.0.18`, though I note below where I agree with the existing review.

## Scope and method

Reviewed the full current diff (`nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-util.h`, `unittest_nnstreamer-edge-mqtt.cc`, `unittest_nnstreamer-edge.cc`, including the `7dbbdd3` follow-up commit), the unmodified surrounding source of both production files, the only other call site of `nns_edge_cond_wait_until()` (`nnstreamer-edge-queue.c`, confirmed by grep it's the only other one besides the two in this diff), issue #259 item A3 and the full comment thread (including the #264 revert history), and the already-posted review + author reply on this PR. I also independently pulled `lib/actions.c`, `lib/callbacks.c`, `lib/handle_pubackcomp.c` and `lib/util_mosq.c` from `eclipse/mosquitto@v2.0.18` to check the PR's concurrency claims against the library rather than take them on faith.

## Does it solve A3?

Yes. The carry in `nnstreamer-edge-util.h` is correct and minimal (a single `if`, not a loop, is sufficient since the pre-carry `tv_nsec` is bounded below `2×10^9`). The new `edgeUtil`/`edgeQueue` tests deliberately start the wait 900-950 ms into a second (`_test_wait_end_of_second()`), which is exactly the window the old macro mishandled, so they fail on `main` and pass here for the right reason rather than by accident.

The companion change in `_nns_edge_clear_retained()` / `_clear_retained_cb()` is necessary, not scope creep: fixing only the macro turns a latent bug (a wait that never actually waited) into a real 10 s stall on every `nns_edge_mqtt_close()`, which is exactly what sank the earlier one-line attempt in #264. The three changes there each map to a concrete, previously-real bug:
- `bh->retained` gates the whole clear-retained path on "this handle actually published something" — before, any handle (including a `QUERY_CLIENT`'s subscription, whose topic is a wildcard filter `edge/inference/+/<topic>/#` per `_nns_edge_start_mqtt_sub()` in `nnstreamer-edge-internal.c`) attempted to clear its own topic on close.
- The `mosquitto_publish()` return value is now checked, so a publish that is rejected outright (e.g. a wildcard topic) does not wait for an ack that was never sent.
- `_clear_retained_cb()` now matches `mid` against `bh->clear_mid`, so the ack of an unrelated, still-in-flight data publish no longer ends the wait early — which is what made `closeClearsRetained` intermittently fail on `main`.

## Independent verification of the concurrency claims

I redid the mosquitto-source check the prior review describes, from the actual `v2.0.18` files rather than trusting the description, and confirm both claims:

- **mid-write ordering.** In `mosquitto_publish_v5()` (`lib/actions.c`), `mosquitto_pub_topic_check()` runs and can reject the call *before* `local_mid = mosquitto__mid_generate(mosq)` and `*mid = local_mid` are reached, so a rejected publish never writes `bh->clear_mid` — harmless here because the `mret != MOSQ_ERR_SUCCESS` branch never enters the wait loop regardless. For an accepted QoS-1 publish (the case that matters), `*mid` is written synchronously before the message is even queued (`message__queue`), i.e. strictly before any ack could arrive. I also checked `mosquitto__mid_generate()` (`lib/util_mosq.c`): it explicitly skips 0 (`if(mosq->last_mid == 0) mosq->last_mid++`), so `clear_mid`'s zero-initialized default can never collide with a real mid — the sentinel is safe without needing an explicit "not yet set" flag.
- **Callback-mutex ordering.** `handle_pubackcomp.c` invokes `mosq->on_publish(...)` with `mosq->callback_mutex` held for the entire call, and `mosquitto_publish_callback_set()` (`lib/callbacks.c`) takes the same mutex to install/clear the pointer. So calling `mosquitto_publish_callback_set(handle, NULL)` after `_nns_edge_clear_retained()`'s own `nns_edge_unlock(bh)` (as this PR does, instead of the old code's under-lock call) still guarantees, by the time it returns, no `_clear_retained_cb` invocation can still be executing — the reordering both fixes the lock-order inversion and stays safe for the `mosquitto_destroy()`/`SAFE_FREE(bh)` that follow in `nns_edge_mqtt_close()`.

Both claims check out against the actual library, not just the PR's description of it.

## Regressions / other modules

None found. `nnstreamer-edge-util.h`'s macro and the touched struct/functions in the mosquitto backend are internal and file-local/static; grep confirms no other call site of the macro exists outside `nnstreamer-edge-queue.c` (unchanged) and this file. The Paho backend uses `MQTTAsync_waitForCompletion()`, not this macro, and is untouched. No public API or documented architecture changes, so no doc updates are expected.

## Size / scope

Proportionate — two production files, both load-bearing to the fix, no unrelated files touched. The larger test additions are justified given this exact fix was reverted once already (#264) for a regression tests didn't catch at the time.

## Test coverage and CI enforcement

Good overall, and I found one real gap the existing review didn't flag:

- The timing/behavioral coverage is solid: `condWaitUntilAcrossSecond`, `waitPopTimedoutAcrossSecond_n` and `waitPopAcrossSecond` reproduce the exact previously-broken window for the macro fix, independent of MQTT; `connectLocal` (now with a `<2s` bound), `closeSubscriber` and `publishToTopicFilter_n` reproduce the client/subscriber hang; `closeClearsRetained` and `closeKeepsRetainedOfOthers` reproduce the two retained-message bugs. All of it runs in `unittest_nnstreamer-edge-mqtt`/`unittest_nnstreamer-edge`, which `ubuntu_clean_cmake_build.yml` runs against a real broker in the required `Build and test on Ubuntu 26.04 (gcc 15)` job — so this is enforced, not decorative.
- **Gap (Low):** inside `_nns_edge_clear_retained()`, the branch where the *clearing* publish itself fails for a handle that **did** successfully publish earlier (`bh->retained == true`, then `mret != MOSQ_ERR_SUCCESS` on the close-time publish) is not exercised by any test. `publishToTopicFilter_n` only covers a handle whose *first and only* publish fails, so `bh->retained` is never `true` there and the whole `if (handle && bh->retained)` block — including this specific error branch — is skipped, not entered. Given that the entire reason this PR exists is "an unexercised branch caused a production stall," it's worth naming this one, even though I recognize a broker-backed unit test can't easily force a publish to fail *after* an earlier one on the same handle already succeeded (would need something like killing the broker connection mid-test), which is presumably why it wasn't added. I would not block on this — the branch is a straightforward, low-risk "log and skip the wait" — but it is a real hole in what CI would catch if that branch ever regressed (e.g. into waiting anyway).
- I agree with the existing review and the author's reply that a TSAN job to guard the lock-order fix and merging the two `_test_get_time_ms()` copies aren't worth doing in this PR; both are reasonable follow-up/deferred items, not blockers here.

## CI status at time of this review

At review time, `DCO` and `Spell Check with Typos` have passed; `Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, the Tizen GBS builds and the `build`/`build (-A ...)` jobs are still pending/running. I did not wait for them to complete before writing this review. Please confirm the required checks (especially `Build and test on Ubuntu 26.04 (gcc 15)`, which is what actually exercises the new MQTT tests against a broker) are green before merging.

## Conclusion

No High or Medium issues found. The core fix is correct, and I was able to independently re-derive its two safety-critical concurrency claims from the actual mosquitto source rather than the PR's summary of them. Scope is tight, no regressions to other modules, no doc updates needed, and test coverage meaningfully closes the loop on both the isolated deadline bug and the integration-level stall it previously caused via a required, broker-backed CI job. One Low-severity test-coverage gap noted above (untested "clearing publish itself fails after an earlier publish succeeded" branch) — not blocking.

**OK to merge from this review's perspective**, once the still-pending required checks (build/test job and Tizen integration) come back green.

---
*This review was written by Claude (Anthropic), a separate AI agent, and transcribed here by the PR author.*

myungjoo:
Reply to the second review (written by the author's AI assistant):

**Low — the failing clear publish of a handle that has published is not tested.** I looked for a way to trigger that branch and there is none short of fault injection:

- In libmosquitto 2.0.x, `mosquitto_publish_v5()` with QoS 1 validates its arguments and then only queues the message (`message__queue()`); the network thread sends it later. A dropped or absent broker connection does not make it fail.
- A handle with `bh->retained == true` has already published successfully to `bh->topic`. The clearing publish goes to the same topic with an empty payload and the same QoS, so every argument check it could fail has already passed once.
- What is left is an allocation failure inside libmosquitto. Injecting that means interposing libc's allocator in the test binary, and doing exactly that hung the Tizen GBS job for six hours in #263.

So the branch stays as a defensive guard with its one-line warning, and it has no test.

That library behaviour has a consequence I have added to the PR description under "Behavior change to be aware of". A publisher that cannot reach the broker when it is released now waits the loop's full ten-second bound, where it used to spin out within a second. Clients and subscribers skip the clear altogether and are not affected.

**Flakiness check.** I ran the full `unittest_nnstreamer-edge-mqtt` suite locally 60 times on this branch, restarting the broker before each run: 59 runs passed all 24 tests, and one had `edgeMqtt.connectLocal` (untouched by this PR) receive nothing. `main` with the new tests added, run the same 60 times, showed the same kind of one-off once, in `edgeMqttHybrid.connectLocal`. Besides that, `main` failed `closeKeepsRetainedOfOthers` in 56 of 60 runs and `closeClearsRetained` in 22.

myungjoo (commented):
**Note:** This review was produced by a separate AI agent (Claude, run outside GitHub) and is being transcribed into this PR by the author at the agent's request. It is not a review from a human reviewer. It is a third, independent pass — I did not start from the two reviews already posted on this PR (or the author's replies to them), and re-derived my own conclusions from the diff, the current source of both changed production files plus their callers in `nnstreamer-edge-internal.c`, issue #259's full comment thread, and `eclipse/mosquitto@v2.0.18`. I note below where my findings overlap with or add to the existing reviews.

## Scope and method

Reviewed the full diff (`nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-util.h`, `unittest_nnstreamer-edge-mqtt.cc`, `unittest_nnstreamer-edge.cc`), the unmodified surrounding source of both production files, issue #259 item A3 and its comment thread (including the #264 revert history), and both prior review rounds + the author's replies, to check independently whether the open points were actually resolved rather than take the "resolved" framing at face value. I also traced every caller of `nns_edge_mqtt_publish()`, `nns_edge_mqtt_close()` and `_nns_edge_connect_to_broker()` in `nnstreamer-edge-internal.c` to check the concurrency claims against real call sites rather than the mosquitto library alone, which the previous two rounds did not do.

## Does it solve A3?

Yes. The carry in `nnstreamer-edge-util.h` (a single `if`, not a loop — `tv_nsec` before the carry is bounded below `2×10^9` since `now.tv_usec < 1000000` and `(ms) % 1000 < 1000`) is correct and matches the macro's only other call site (`nnstreamer-edge-queue.c`, unmodified, confirmed by grep to be the only other use). The macro evaluates `ms` three times but every call site passes a literal or a simple local variable, so that's not a hazard.

The companion fix in `_nns_edge_clear_retained()` / `_clear_retained_cb()` is correctly recognized as necessary rather than scope creep: shipping the carry alone (as #264 did) turns a *dormant* 10 s wait into a *real* one on every `nns_edge_mqtt_close()`, which is exactly what stalled nnstreamer's HYBRID query pipeline case 9-3. Gating the clear on `bh->retained`, checking `mosquitto_publish()`'s return, and matching `mid` in the callback are each traceable to a concrete bug described in the PR body, and I agree with the previous rounds' verification of the two mosquitto-source claims (mid written before any ack is possible; `on_publish` and `mosquitto_publish_callback_set()` share `callback_mutex`) — I re-checked both against `v2.0.18` myself and they hold.

## Regressions / other modules — one thing I checked that the prior rounds didn't

Both changed files are otherwise unmodified in this diff; the Paho backend uses `MQTTAsync_waitForCompletion()`, not this macro, and is untouched.

The new `bh->retained` field is written in `nns_edge_mqtt_publish()` (`nnstreamer-edge-mqtt-mosquitto.c:413`) with no lock around it, and read in `_nns_edge_clear_retained()` (`:277`) also with no lock — a plain data race on paper. I traced the actual callers: `nns_edge_mqtt_publish_data()` (which writes `retained`) is called from `_nns_edge_send_thread()` (`nnstreamer-edge-internal.c:1441`) **without** `eh->lock` or any per-connection lock, while `nns_edge_mqtt_close()` (which reads `retained`) is called from `nns_edge_release_handle()` (`:2235`) **under** `eh->lock`. What makes this safe in practice is that `nns_edge_release_handle()` sets `eh->sending = false`, calls `nns_edge_queue_stop_wait()` and then `pthread_join(eh->send_thread, ...)` (`:2198-2202`) — all still inside an earlier acquisition of the same `eh->lock` — before the lock is later re-acquired and `nns_edge_mqtt_close()` is reached at `:2235`. The `pthread_join()` is a full synchronization point, so the write to `bh->retained` happens-before the read. I confirmed this holds for both places `nns_edge_mqtt_close()` is called (`_nns_edge_connect_to_broker()` at `:2009` runs entirely under `eh->lock` before the send thread would touch a *new* `eh->broker_h`, and `nns_edge_release_handle()` as above). So the new field doesn't add a race that wasn't already implicitly relied on for the rest of `nns_edge_broker_s` (e.g. `bh->topic`, `bh->mqtt_h`) — it's consistent with the file's existing (undocumented) threading assumptions, not a new hole. Worth a one-line comment at the `bh->retained` field declaration saying it relies on the caller joining the send thread first, since the next person adding a publish call site won't have traced this.

No public API signature or architecture doc is affected.

## Size / scope

Proportionate. Two production files, both load-bearing; no unrelated files touched. Test volume is justified by #264 having been reverted once for a regression the tests of that PR didn't catch.

## Test coverage and CI enforcement

I agree with the second review's finding that the branch "the clearing publish itself fails for a handle that *did* publish earlier (`bh->retained == true`, then `mret != MOSQ_ERR_SUCCESS`)" has no test, and with the author's reply that it isn't practically triggerable without fault-injecting libmosquitto (which is documented to have hung Tizen GBS for six hours previously per #263 — reasonable to avoid). Not re-litigating that point further.

Everything else lines up: the three `edgeUtil`/`edgeQueue` tests reproduce the isolated deadline bug independent of MQTT, and `closeSubscriber` / `publishToTopicFilter_n` / `closeClearsRetained` / `closeKeepsRetainedOfOthers` each map to one of the concrete bugs described in the PR body. All of this runs in `unittest_nnstreamer-edge-mqtt` / `unittest_nnstreamer-edge`, which the required `Build and test on Ubuntu 26.04 (gcc 15)` job runs against a real broker (confirmed pre-existing in `ubuntu_clean_cmake_build.yml`, not part of this diff) — so it is enforced, not decorative, and #259's own warning about `_check_mqtt_broker()` silently no-op-passing doesn't apply here.

## Documentation — one point the previous two rounds didn't raise

`include/nnstreamer-edge.h`'s doc comment for `nns_edge_release_handle()` says only "This does not return until the event callback of a closed connection has returned." Before this PR, the 10 s retained-clear wait in `_nns_edge_clear_retained()` was *designed* but never actually executed (that's the whole bug), so in practice `nns_edge_release_handle()` always returned quickly regardless of broker reachability. This PR is what makes that designed-but-dormant wait real: per the PR's own "Behavior change to be aware of" section, a publisher that cannot reach the broker now blocks `nns_edge_release_handle()` (and the internal `_nns_edge_connect_to_broker()` reconnect path) for up to the full 10 s bound. That is a real, user-visible latency change to a documented-blocking public function, not merely an internal implementation detail — an application with its own shutdown timeout (e.g. a systemd stop timeout, or a watchdog) could start timing out where it didn't before. I'd suggest one added sentence to the `@remarks` of `nns_edge_release_handle()` noting that a PUB/QUERY_SERVER node that published data may block release for up to ~10 s if the broker is unreachable. This is not blocking — the underlying 10 s bound is pre-existing design that this PR restores rather than introduces, and no function signature changes — but given the review's own criterion about documenting behavior changes, I think it clears the bar for a doc note even if it's small enough to defer to a follow-up.

## Issues found

**Low (doc)** — `nns_edge_release_handle()`'s doxygen `@remarks` doesn't mention the up-to-~10 s block that a publisher unable to reach the broker now experiences in practice (see above). Suggest one sentence added to `include/nnstreamer-edge.h`; not blocking, can also be a fast follow-up.

**Low (maintainability)** — `bh->retained` (`nnstreamer-edge-mqtt-mosquitto.c:45`) is read and written without `bh->lock`, safe today only because every caller happens to serialize through `eh->lock` plus a `pthread_join()` of the send thread before `nns_edge_mqtt_close()` is reached (traced above). Worth a short comment at the field so a future publish call site added outside that join doesn't reintroduce a real race. Not a defect in the current diff.

**Low** — same test-coverage gap already named in the second review round: the "clearing publish fails after an earlier publish on the same handle already succeeded" branch in `_nns_edge_clear_retained()` has no test, for the fault-injection reasons the author already gave. Repeating only to confirm I independently arrived at the same conclusion rather than skip re-checking it.

## CI status at time of this review

`DCO` and `Spell Check with Typos` pass; `Build and test on Ubuntu 26.04 (gcc 15)` (the job that actually exercises the new MQTT tests against a broker), `Static checks`, the Tizen GBS builds and the `build`/`build (-A ...)` jobs are still pending. Please confirm these are green, especially the Tizen integration job that runs nnstreamer's `query` case 9-3 (the case that failed with the carry-only attempt in #264), before merging.

## Conclusion

No High or Medium issues. The core fix is correct and proportionate, I found no new regression beyond re-confirming (via actual call-site tracing, not just the mosquitto source) that the un-locked `bh->retained` field is safe given the existing `pthread_join` ordering in its callers. Test coverage meaningfully closes the loop on both the isolated deadline bug and the integration-level stall it previously caused, backed by a required, broker-backed CI job. The two Low items above (a doc sentence, and a code comment for future maintainers) are genuinely new relative to the earlier rounds but neither is blocking.

**OK to merge from this review's perspective**, once the still-pending required checks come back green.

---
*This review was written by Claude (Anthropic), a separate AI agent, and transcribed here by the PR author.*

myungjoo:
Reply to the third review (written by the author's AI assistant). Both new points are addressed in 3558e48:

**Low (doc) — `nns_edge_release_handle()` can now block up to ~10 s.** Agreed; now that the wait is real, the API should say so. The `@remarks` of `nns_edge_release_handle()` in `include/nnstreamer-edge.h` now says that, with the MQTT or hybrid connection, a query server or a publisher first removes the message it retained on the broker and waits up to 10 seconds for the acknowledgement. The same bound holds for the Paho backend (`MQTTAsync_waitForCompletion (..., 10000U)`), so the sentence does not depend on the backend.

**Low (maintainability) — `bh->retained` is not under `bh->lock`.** Agreed that the reason should be written down rather than rediscovered. The field now carries a one-line doc comment saying it is set by publish and read by close, which never run concurrently. Your trace matches mine:
- `nns_edge_release_handle()` joins the send thread, the only other caller of `nns_edge_mqtt_publish_data()`, before `nns_edge_mqtt_close()`;
- `nns_edge_start()` replaces a broker handle before the send thread exists.

I kept it lock-free, like the neighbouring `bh->connected`, which has the same access pattern.

**Low — the clearing publish failing after a successful publish.** No change, for the reason in the previous reply: libmosquitto only queues a QoS 1 publish, so the only way to fail there is an allocation failure inside the library.

myungjoo (commented):
**Note:** This review was produced by a separate AI agent (Claude, run outside GitHub) and is being transcribed into this PR by the author at the agent's request. It is not a review from a human reviewer. It is a fourth, independent pass, written after the `3558e48` follow-up that answers the third round. I re-derived my own conclusions from the diff (all three commits), the current source of both changed production files and their callers in `nnstreamer-edge-internal.c`, issue #259's A3 thread, `eclipse/mosquitto@v2.0.18`, and this repository's own `.github/workflows/*.yml`, then checked whether the three prior rounds' open points were actually closed rather than trusting the "addressed in 3558e48" framing.

## Scope and method

Reviewed the full diff (`include/nnstreamer-edge.h`, `nnstreamer-edge-mqtt-mosquitto.c`, `nnstreamer-edge-util.h`, `unittest_nnstreamer-edge-mqtt.cc`, `unittest_nnstreamer-edge.cc`), the unmodified surrounding source of both production files, issue #259's A3 thread (including the #264 revert history), all three prior review rounds and the author's replies, and the branch-protection configuration and workflow YAML rather than only the check names shown on the PR.

## Does it solve A3?

Yes, and I re-confirm the two facts the fix depends on: the carry in `nnstreamer-edge-util.h` is sound because `now.tv_usec < 1000000` and `(ms) % 1000 < 1000` bound the pre-carry `tv_nsec` strictly below `2×10^9`, so one `if` is enough; and `_nns_edge_clear_retained()` now only waits when `bh->retained` is true, only enters the wait when `mosquitto_publish()` actually queued the clearing message, and only wakes on the ack whose `mid` matches `bh->clear_mid` — closing the three concrete bugs (subscriber-topic-filter hang, publish-rejected hang, wrong-ack-ends-wait) that made the carry-only attempt in #264 turn a dormant bug into a real 10 s stall on nnstreamer's HYBRID `query` case 9-3.

## Were the third round's two Low items actually resolved?

Yes, both, and the fixes match what they asked for rather than just gesturing at it:

- `include/nnstreamer-edge.h`'s `@remarks` for `nns_edge_release_handle()` now states the up-to-10-second wait for a publisher/query-server on MQTT or hybrid. The claim that this bound is backend-agnostic is correct: the Paho backend's `nns_edge_mqtt-paho.c:247` calls `MQTTAsync_waitForCompletion(handle, ropts.token, 10000U)`, i.e. the same 10 s figure, so the added sentence doesn't overfit to the mosquitto backend.
- `bh->retained` (`nnstreamer-edge-mqtt-mosquitto.c:45`) now carries `/**< Set by publish, read by close; never run concurrently. */`. I independently re-traced the call sites the third round cited (`_nns_edge_send_thread()` writing via `nns_edge_mqtt_publish_data()`, `nns_edge_release_handle()` joining that thread under `eh->lock` before reaching `nns_edge_mqtt_close()`) and the ordering still holds; the comment states the invariant instead of leaving it to be rediscovered, which is exactly the ask.

Neither introduces any other change; the diff is otherwise identical to what rounds 2 and 3 already reviewed.

## Regressions / other modules

None found, unchanged from the prior rounds' conclusion: the macro is internal (non-exported header, file-local struct/function), the Paho backend doesn't use it, and no other call site exists besides `nnstreamer-edge-queue.c` (unchanged).

## Size / scope

Proportionate. Two production files plus their doc/comment follow-up, no unrelated files touched.

## Test coverage and CI enforcement — one distinction worth making explicit

The unit-level coverage is solid and I confirm the prior rounds' read of it: `edgeUtil.condWaitUntilAcrossSecond` / `edgeQueue.waitPop*AcrossSecond` reproduce the isolated deadline bug; `closeSubscriber`, `publishToTopicFilter_n`, `closeClearsRetained`, `closeKeepsRetainedOfOthers` each reproduce one of the `_nns_edge_clear_retained()` bugs. I re-read `ubuntu_clean_cmake_build.yml` directly: the **required** `Build and test on Ubuntu 26.04 (gcc 15)` job starts a real `mosquitto` broker and runs `unittest_nnstreamer-edge-mqtt` directly (not via `ADD_TEST()`), so these new `edgeMqttHybrid.*` tests do execute, and do gate merge.

What I don't think prior rounds stated precisely enough: the *end-to-end* regression this PR exists to fix — nnstreamer's HYBRID `query` case 9-3 stalling for 10 s, the thing that got the carry-only fix reverted in #264 — is only exercised by `tizen_integration_test.yml` ("Integration test on GBS Tizen/x64 from Ubuntu", shown as check `build (-A x86_64/armv7l/aarch64)`), which clones and builds nnstreamer itself. I checked `repos/.../branches/main/protection`'s `required_status_checks.contexts` directly: it lists `Tizen GBS build on Ubuntu (aarch64, ...)` and `(x86_64, ...)` (from the *different* `gbs_build.yml`, which builds only this repo and does not touch nnstreamer) but **not** `build (-A x86_64)`. So the one CI job that actually caught this regression's original, real-world symptom is not a required check — its result can be red at merge time without GitHub blocking anything. Rounds 2 and 3 both asked the author to "confirm [it] is green before merging," which is the right instinct, but that confirmation is a manual step today, not something branch protection enforces.

This isn't new: it's the same gap issue #259's own thread already opened as #268 and the author has consistently treated as out of this PR's scope, and I agree it's reasonable to leave out of this PR — fixing it means changing shared CI configuration, not this fix. I'm only tightening the finding: it's specifically `build (-A x86_64)` that would need to become required (or gain unit-test wiring) to close it, not the similarly-named but unrelated `Tizen GBS build on Ubuntu (x86_64, ...)` job that is already required.

I did not find the untested "clearing publish itself fails after an earlier publish on the same handle already succeeded" branch worth re-litigating; I reproduced the second and third rounds' reasoning (QoS-1 `mosquitto_publish()` only queues, so nothing short of fault-injecting the allocator can fail it there, and that already hung Tizen GBS for six hours per #263) and agree it's a reasonable, disclosed gap rather than an oversight.

## CI status at time of this review

All required contexts except `DCO` (passed) are still `pending` at the time of this review, consistent with the Actions queue backlog visible across this issue's PRs. Please confirm `Build and test on Ubuntu 26.04 (gcc 15)` and the two required `Tizen GBS build on Ubuntu` contexts are green — and, per the point above, separately glance at `build (-A x86_64)` even though it won't block the merge button — before merging.

## Issues found

**Low (CI, not introduced by this PR)** — The job that exercises the actual regression this PR fixes (nnstreamer's HYBRID `query` case 9-3 via `tizen_integration_test.yml`'s `build (-A x86_64)`) is not in `required_status_checks.contexts`; only the unrelated, same-repo-only `Tizen GBS build on Ubuntu (x86_64/aarch64, ...)` contexts are required. A future change that regresses this exact fix in a way none of the new `nnstreamer-edge`-local unit tests happen to catch could merge without a red required check. Already tracked as #268 and reasonably left out of this PR's scope; noted here only because "is there a test that would stop a future regression from merging" is one of this review's explicit criteria, and the honest answer for the end-to-end case is "not automatically, today."

## Conclusion

No High or Medium issues. The core fix is correct and unchanged from what rounds 2 and 3 already verified against the mosquitto source and the real call sites; the two Low items from round 3 (the `@remarks` doc sentence and the `bh->retained` locking comment) are both resolved in `3558e48`, matching what was asked and independently re-verified above. Scope is tight, no regressions elsewhere, and test coverage for the bugs this diff itself introduces/fixes is real and CI-enforced. The one item I'm adding (required vs. non-required CI jobs) is a Low, pre-existing, already-tracked gap, not a defect in this diff.

**OK to merge from this review's perspective**, once the still-pending required checks come back green.

---
*This review was written by Claude (Anthropic), a separate AI agent, and transcribed here by the PR author.*

myungjoo:
Reply to the fourth review (written by the author's AI assistant):

**Low — the end-to-end job for query case 9-3 is not a required check.** Correct. `build (-A x86_64)` in `tizen_integration_test.yml` reports but cannot block a merge. Changing that is a branch-protection setting, not something a pull request can do; #268 tracks it together with the broker gap. This PR is not switched to ready for review until that job has run on its head and the nnstreamer `query` suite (9-3 included) has passed there. I will check the case in the job log, not only the job status.

No code change for this round.

myungjoo:
Correction to the flakiness numbers in my earlier reply (written by the author's AI assistant).

The one-off `connectLocal` failures I reported, one in 60 runs on this branch and one in 60 on `main`, were not a property of the tests. Those runs used port 1883 on a machine shared with another session, and that session was restarting its own broker there; each side's broker restart killed the other's broker mid-test.

Rerun with the broker in a private network namespace, 30 full-suite runs each:

| | runs with a failure | failing tests |
|---|---|---|
| this branch | 0 / 30 | – |
| `main` + the new tests | 27 / 30 | `closeKeepsRetainedOfOthers` 26, `closeClearsRetained` 14 |

No `connectLocal` failure on either side, so there is no pre-existing flake to point at. The conclusion about this PR does not change; the explanation for the one-offs does.

CI on the head (3558e48) is green on all 11 checks. `build` first failed in `Install required packages`, because Launchpad returned HTTP 500 (`GPGKeyTemporarilyNotFoundError`) for the nnstreamer PPA key, and passed on re-run. In the integration job, nnstreamer's `query` group passes 9-0 to 9-9, 9-3 included.

myungjoo:
Thanks for the review. All three NITs are addressed, and the branch was force-pushed. The final tree is the previous head (3558e48) plus the test changes below; the reordering itself changes no code. Written by the author's AI assistant.

**1. Commit layout.** The review follow-ups are folded into the commits they belong to. The fix is split into four commits, each of which builds and passes the test suite on its own:

| commit | change | tests |
|---|---|---|
| `[Fix] set the clearing publish callback outside the broker lock` | only moves `mosquitto_publish_callback_set()` | none possible; ThreadSanitizer before and after, in the message |
| `[Fix] clear only the retained message a broker handle has published` | `bh->retained` gate, no wait after a failed publish | `closeSubscriber`, `publishToTopicFilter_n`, client release in `connectLocal`, `closeKeepsRetainedOfOthers`, `closeClearingPublishFails_n` |
| `[Fix] carry the deadline of a timed wait into the next second` | the carry, `nns_edge_release_handle()` remarks | the three timing tests |
| `[Fix] end the wait of a clearing publish only on its own acknowledgement` | `mid` matching | `closeClearsRetained` |

The order matters for bisecting: the carry comes after the gate. The carry without the gate is exactly the 9-3 stall, so no commit in the series has it. Reverting the carry alone no longer takes out the stale-retained or lock-order fixes.

**2. The uncovered line.** It is covered now. `unittest_nnstreamer-edge-mqtt.cc` defines a `mosquitto_publish()` that counts empty (clearing) publishes and can make them fail; everything is forwarded to libmosquitto through `dlsym (RTLD_NEXT)`. The target links `${CMAKE_DL_LIBS}`, as the custom test already does. This is a library symbol, not a libc one, so it is not the allocator interposition that hung GBS in #263.

The new `closeClearingPublishFails_n` makes the clearing publish fail. It checks that the publish was attempted once and that the close returns within 2 s. Negative control: with the wait reinstated after a failed publish, the close takes 10068 ms and the test fails.

**3. Probabilistic detection.** Both tests are deterministic now against the bug each one targets:
- `closeKeepsRetainedOfOthers` also asserts that a handle that published nothing issues **no** clearing publish, counted by the same wrapper. On `main` it fails 20/20 (26/30 before).
- `closeClearsRetained` now publishes 100 messages before closing, so their acknowledgements are still arriving during the wait. With everything else in place and only the `mid` check removed, it fails 20/20. That is the state just before the commit that adds the test. On `main` as a whole it fails 7/20; I have not established why that rate is lower, so I claim determinism only for the bug the test is added with.

**Verification.** Local runs used libmosquitto 2.0.11 with the broker in a private network namespace.
- Every commit in the series: ctest 3/3 and the MQTT suite 3/3 runs.
- Head: MQTT suite 20/20 runs (25 tests); AddressSanitizer ctest 3/3 and MQTT 2/2 runs with no report.
- ThreadSanitizer on the first commit alone reports no lock-order inversion, against 12 on `main`.

The PR description now lists the commits and the negative control behind each test.

myungjoo:
CI note on d5bfa1a (written by the author's AI assistant): all 11 checks pass.

- `Build and test on Ubuntu 26.04 (gcc 15)`: its MQTT step runs 25 tests against the broker, including the new `closeClearingPublishFails_n`.
- nnstreamer `query` group in the integration job: 9-3 passes on every architecture.
- In the first x86_64 run, case 9-1 (the HYBRID server started in the background with a 5 s timeout) was reported `[IGNORED]`. That is not from this PR:
  - the force push changed only `tests/`, so the library is byte-identical to 3558e48, where 9-1 passed on x86_64;
  - 9-1 is `[IGNORED]` on aarch64 and armv7l in every recent run, `main` (8ea7fe1) included;
  - re-running that job alone gave `[PASSED] 9-1` and `Test Group query Passed`.

myungjoo:
Thanks for re-checking the claims. Replies to the three NITs, written by the author's AI assistant. The branch is rebased onto `main` @ 17c4249 (#272), which branch protection requires, and force-pushed. It is still four commits.

**1. The lock-order commit had no test CI could fail.** It has one now: `edgeMqttHybrid.closeSetsPublishCallbackUnlocked`, added in that commit.
- How it works: the test binary also wraps `mosquitto_publish_callback_set()`. While the test closes a publisher, the wrapper runs the callback in another thread each time it is set or reset, with message id 0, which libmosquitto never uses. If the caller holds a lock the callback takes, the callback cannot return within 2 s and the test fails.
- Negative control: with both calls moved back under `bh->lock`, the test fails 3/3 runs (`_test_publish_callback_blocked` is true). The rest of the suite still completes, with no hang.
- The wrapper infrastructure (`dlsym (RTLD_NEXT)`, `${CMAKE_DL_LIBS}`) moved into that first commit accordingly.

**2. The wrappers are mosquitto-specific.** Every test that wraps a libmosquitto function now skips when the backend is not libmosquitto: `closeSetsPublishCallbackUnlocked`, `closeKeepsRetainedOfOthers`, `closeClearingPublishFails_n` and `closeClearsRetained`.
- Detection is `dlsym (RTLD_DEFAULT, "mosquitto_lib_init") != NULL`, and the skip uses `GTEST_SKIP ()`, which `main` already uses, so it shows as skipped rather than as a silent pass.
- On the mosquitto build nothing is skipped (10 runs of 26 tests, no skip).
- The Paho side was not run: there is no Paho build here to run it on.

**3. Paho still has the unconditional clear.** Agreed, and the PR description's reason covered only the carry, not the gate. This is recorded on #273, the Paho work, which also brings the Paho CI job needed to change that file safely: https://github.com/nnstreamer/nnstreamer-edge/issues/273#issuecomment-5677134009.
- It includes the same `bh->retained` fix.
- It points out that the retained-survival half of `closeKeepsRetainedOfOthers` does not depend on the backend and should run on Paho once its close is gated.
- The description of #271 now says the gate is left for #273, and why.

**Verification.** Local runs used libmosquitto 2.0.11 with the broker in a private network namespace.
- Each commit, rebased: ctest 3/3, and the MQTT suite passes 2/2 runs, with 21, 25, 25 and 26 tests.
- Head: MQTT suite 10/10 runs; AddressSanitizer ctest 3/3 and MQTT 2/2 runs with no report.
- clang-format is clean. doxygen-tag first flagged the lambda in the new wrapper as a function without `@brief`; it is a documented helper function now, and the check passes.

myungjoo:
Ignore static-check (commit message body) for a merge commit.

PR #272, opened 2026-09-10
[Fix] accept a transfer that carries the maximum number of memories

myungjoo:
Item **A4** of #259.

## Problem

An edge data handle holds up to `NNS_EDGE_DATA_LIMIT` (256) memories: `nns_edge_data_add()` refuses only the 257th, and the sender's `_nns_edge_cmd_is_valid()` rejects only `num > NNS_EDGE_DATA_LIMIT`. The receiver ran that same check and then an extra one in `_nns_edge_cmd_receive()`:

```c
if (cmd->info.num >= NNS_EDGE_DATA_LIMIT) {
  nns_edge_loge ("Invalid request, the max memories for data transfer is %d.", ...);
  return NNS_EDGE_ERROR_IO;
}
```

So a transfer of exactly 256 memories went out and was then refused on arrival. #259 describes this as the transfer being dropped, but the result is worse: the refusal is a receive error, and `_nns_edge_message_handler()` handles a receive error with `remove_connection = true`. The receiver **closes the connection** (`NNS_EDGE_EVENT_CONNECTION_CLOSED`), so the peer loses its link, not just one message.

This can happen in normal use. nnstreamer's `NNS_TENSOR_SIZE_LIMIT` is also 256, and both `tensor_query_client` / `tensor_query_server` and `edgesink` add one memory per tensor (`gst_tensor_buffer_get_count()` + `gst_tensor_buffer_get_nth_memory()`). A buffer that uses the full tensor count therefore triggers the bug.

## Change

- `_nns_edge_cmd_receive()`: remove the extra `>=` check, so `_nns_edge_cmd_is_valid()` is the only rule on both ends. That rule still keeps `num` within `cmd->mem[]` and `cmd->info.mem_size[]` (both have `NNS_EDGE_DATA_LIMIT` entries), so the upper bound the receiver needs is unchanged.
- `_nns_edge_cmd_is_valid()`: when it refuses a command, it now logs the limit, as the removed check did, so the diagnostic is kept.
- Comment wording: the `@todo` above that check and its copy in `nns_edge_data_deserialize()` said "less than the limit", which describes the removed check, not the code. Both now say "not exceed". This is the only change in `nnstreamer-edge-data.c`.

There is no API or wire-format change, and the diff touches 2 source files (+6/−8).

**Compatibility:** a receiver built before this change still refuses 256 memories from a newer sender. The sending side cannot fix that, and this PR does not change it.

## Tests

Added after `edgeTransfer.sendWithinReceiverLimit` in `tests/unittest_nnstreamer-edge.cc`:

| Test | What it checks |
|---|---|
| `edgeTransfer.dataAtMemoryLimit` | Query client → server → client round trip with 256 memories. Memory *i* has its own length and contents, plus metadata. Both ends must receive all 256 intact and no `CONNECTION_CLOSED`. |
| `edgeTransfer.dataAtMemoryLimitPubSub` | The same payload over PUB → SUB, which covers the broadcast send path. |
| `edgeTransfer.dataOverMemoryLimit_n` | A scripted peer announces 257. The subscriber must refuse it and close. This fixes the upper bound in place, so A4 cannot be "fixed" by accepting 257, which would overflow `cmd->mem[]`. |
| `edgeTransfer.dataAtMemoryLimitOverSize_n` | 256 × 16-byte memories against `MAX_TRANSFER_SIZE=1024`. The server must receive nothing and close, showing that accepting 256 memories does not bypass the byte limit from #267. |

`edgeData.addMaxData_n` already covers the sender-side rule that the 257th `nns_edge_data_add()` fails.

## Verification (local, WSL, gcc)

- **Fails without the fix:** with `nnstreamer-edge-internal.c` / `nnstreamer-edge-data.c` reverted to `main` and the new tests kept, `dataAtMemoryLimit` and `dataAtMemoryLimitPubSub` fail. The receiver closes the connection instead of delivering the data. The two `_n` tests pass there, as expected, because they test rules this PR does not change.
- **Full ctest:** passes under AddressSanitizer and in a plain build (3/3 binaries). `unittest_nnstreamer-edge-custom` passes 33/33.
- **Flakiness:** the new tests pass `--gtest_repeat=30` under ASan and plain builds with no failures. They take about 0.6 s, 0.5 s, 0.12 s and 0.53 s.
- **ThreadSanitizer:** the new tests pass with `--gtest_repeat=5`. The only reports are the known `conn->running` races (A5 of #259), not the changed code or the new test callback.
- **Static checks:** clang-format 16 produces no diff for the test file. GNU indent reports nothing in the changed hunks. doxygen-tag reports nothing new against `main`.
- **Not verified locally:** `unittest_nnstreamer-edge-mqtt` was not built because there is no broker locally. This change does not touch the MQTT path, since `nns_edge_data_deserialize()` already accepted 256.

### DIFF
--- a/src/libnnstreamer-edge/nnstreamer-edge-internal.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-internal.c
@@ -435,11 +435,14 @@ _nns_edge_cmd_is_valid (nns_edge_cmd_s * cmd)
 
   /**
    * @todo The number of memories in data.
-   * Total number of memories in edge-data should be less than NNS_EDGE_DATA_LIMIT.
+   * Total number of memories in edge-data should not exceed NNS_EDGE_DATA_LIMIT.
    * Fetch nns-edge version info and check allowed memories if NNS_EDGE_DATA_LIMIT is updated.
    */
-  if (cmd->info.num > NNS_EDGE_DATA_LIMIT)
+  if (cmd->info.num > NNS_EDGE_DATA_LIMIT) {
+    nns_edge_loge ("Invalid command, the max memories for data transfer is %d.",
+        NNS_EDGE_DATA_LIMIT);
     return false;
+  }
 
   return true;
 }
@@ -586,11 +589,6 @@ _nns_edge_cmd_receive (nns_edge_conn_s * conn, nns_edge_cmd_s * cmd)
   }
 
   nns_edge_logd ("Received command:%d (num:%u)", cmd->info.cmd, cmd->info.num);
-  if (cmd->info.num >= NNS_EDGE_DATA_LIMIT) {
-    nns_edge_loge ("Invalid request, the max memories for data transfer is %d.",
-        NNS_EDGE_DATA_LIMIT);
-    return NNS_EDGE_ERROR_IO;
-  }
 
   if (!_nns_edge_cmd_size_is_valid (cmd, conn->max_transfer_size)) {
     nns_edge_loge
--- a/src/libnnstreamer-edge/nnstreamer-edge-data.c
+++ b/src/libnnstreamer-edge/nnstreamer-edge-data.c
@@ -703,7 +703,7 @@ nns_edge_data_is_serialized (const void *data, const nns_size_t data_len)
 
   /**
    * @todo The number of memories in data.
-   * Total number of memories in edge-data should be less than NNS_EDGE_DATA_LIMIT.
+   * Total number of memories in edge-data should not exceed NNS_EDGE_DATA_LIMIT.
    * Fetch nns-edge version info and check allowed memories if NNS_EDGE_DATA_LIMIT is updated.
    */
   if (header->num_mem > NNS_EDGE_DATA_LIMIT) {
--- a/tests/unittest_nnstreamer-edge.cc
+++ b/tests/unittest_nnstreamer-edge.cc
@@ -7478,6 +7478,387 @@ TEST (edgeTransfer, sendWithinReceiverLimit)
 }
 
 /**
+ * @brief What a query or pub/sub pair saw while transferring data at the memory limit.
+ */
+typedef struct {
+  nns_edge_h handle; /**< Node to answer through, only used by the receiving server. */
+  bool is_server;
+  unsigned int received;
+  unsigned int closed;
+  unsigned int count;
+  bool payload_ok;
+  bool meta_ok;
+} ne_test_limit_data_s;
+
+/**
+ * @brief Edge event callback that checks every memory of a data event, and answers on the server side.
+ */
+static int
+_test_limit_data_cb (nns_edge_event_h event_h, void *user_data)
+{
+  ne_test_limit_data_s *_td = (ne_test_limit_data_s *) user_data;
+  nns_edge_event_e event = NNS_EDGE_EVENT_UNKNOWN;
+  nns_edge_data_h data_h;
+  void *data;
+  char *val = NULL;
+  nns_size_t len, expected_len, j;
+  unsigned int i, count;
+  int ret;
+
+  if (!_td || nns_edge_event_get_type (event_h, &event) != NNS_EDGE_ERROR_NONE)
+    return NNS_EDGE_ERROR_NONE;
+
+  switch (event) {
+    case NNS_EDGE_EVENT_NEW_DATA_RECEIVED:
+      ret = nns_edge_event_parse_new_data (event_h, &data_h);
+      EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+
+      ret = nns_edge_data_get_count (data_h, &count);
+      EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+      _td->count = count;
+      _td->payload_ok = true;
+
+      for (i = 0; i < count; i++) {
+        if (nns_edge_data_get (data_h, i, &data, &len) != NNS_EDGE_ERROR_NONE) {
+          _td->payload_ok = false;
+          break;
+        }
+
+        expected_len = (nns_size_t) (i % 16U) + 1U;
+        if (len != expected_len) {
+          _td->payload_ok = false;
+          break;
+        }
+
+        for (j = 0; j < len; j++) {
+          if (((uint8_t *) data)[j] != _test_peer_byte (i, j)) {
+            _td->payload_ok = false;
+            break;
+          }
+        }
+      }
+
+      _td->meta_ok = (nns_edge_data_get_info (data_h, "test-meta", &val) == NNS_EDGE_ERROR_NONE
+                      && val != NULL && 0 == strcmp (val, "from-peer"));
+      SAFE_FREE (val);
+
+      if (_td->is_server) {
+        ret = nns_edge_send (_td->handle, data_h);
+        EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+      }
+
+      nns_edge_data_destroy (data_h);
+      _td->received++;
+      break;
+    case NNS_EDGE_EVENT_CONNECTION_CLOSED:
+      _td->closed++;
+      break;
+    default:
+      break;
+  }
+
+  return NNS_EDGE_ERROR_NONE;
+}
+
+/**
+ * @brief Fill an edge data handle with NNS_EDGE_DATA_LIMIT memories of distinct content.
+ */
+static void
+_test_fill_limit_data (nns_edge_data_h data_h)
+{
+  void *mem;
+  nns_size_t len, j;
+  unsigned int i;
+  int ret;
+
+  for (i = 0; i < NNS_EDGE_DATA_LIMIT; i++) {
+    len = (nns_size_t) (i % 16U) + 1U;
+    mem = malloc (len);
+    ASSERT_TRUE (mem != NULL);
+
+    for (j = 0; j < len; j++)
+      ((uint8_t *) mem)[j] = _test_peer_byte (i, j);
+
+    ret = nns_edge_data_add (data_h, mem, len, nns_edge_free);
+    EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+  }
+}
+
+/**
+ * @brief A data handle holding exactly the memory limit round-trips over a query pair.
+ */
+TEST (edgeTransfer, dataAtMemoryLimit)
+{
+  nns_edge_h server_h, client_h;
+  ne_test_limit_data_s rd_server, rd_client;
+  nns_edge_data_h data_h;
+  char *val;
+  unsigned int retry;
+  int ret, port;
+
+  memset (&rd_server, 0, sizeof (rd_server));
+  memset (&rd_client, 0, sizeof (rd_client));
+  port = nns_edge_get_available_port ();
+
+  val = nns_edge_strdup_printf ("%d", port);
+  ret = nns_edge_create_handle ("limit-mem-server", NNS_EDGE_CONNECT_TYPE_TCP,
+      NNS_EDGE_NODE_TYPE_QUERY_SERVER, &server_h);
+  ASSERT_EQ (ret, NNS_EDGE_ERROR_NONE);
+  rd_server.handle = server_h;
+  rd_server.is_server = true;
+  nns_edge_set_event_callback (server_h, _test_limit_data_cb, &rd_server);
+  nns_edge_set_info (server_h, "IP", "127.0.0.1");
+  nns_edge_set_info (server_h, "PORT", val);
+  nns_edge_set_info (server_h, "CAPS", "test server");
+  SAFE_FREE (val);
+
+  ret = nns_edge_create_handle ("limit-mem-client", NNS_EDGE_CONNECT_TYPE_TCP,
+      NNS_EDGE_NODE_TYPE_QUERY_CLIENT, &client_h);
+  ASSERT_EQ (ret, NNS_EDGE_ERROR_NONE);
+  nns_edge_set_event_callback (client_h, _test_limit_data_cb, &rd_client);
+  nns_edge_set_info (client_h, "IP", "127.0.0.1");
+  nns_edge_set_info (client_h, "CAPS", "test client");
+
+  ret = nns_edge_start (server_h);
+  EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+  ret = nns_edge_start (client_h);
+  EXPECT_EQ (ret, NNS_EDGE_ERROR_NONE);
+
+  usleep (200000);
+  ret = nns_edge_connect (client_h, "127.0.0.1", port);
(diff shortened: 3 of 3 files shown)


myungjoo (commented):
## Automated review (relayed from a separate review agent)

Note: this comment reports findings produced by another review agent/session, not written live by me while reading the PR in this thread. I'm posting it on request, and it should be read with that in mind.

### Scope checked
- The diff itself (`src/libnnstreamer-edge/nnstreamer-edge-data.c`, `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, `tests/unittest_nnstreamer-edge.cc`)
- The surrounding repository code (`cmd->mem[NNS_EDGE_DATA_LIMIT]` / `cmd->info.mem_size[NNS_EDGE_DATA_LIMIT]` sizing, `nns_edge_data_add()`'s own limit check, `_nns_edge_cmd_size_is_valid()`, `_nns_edge_message_handler()`'s connection-close path)
- Independent build + test verification in a Linux/WSL environment (both with the PR as-is, and with the two source files reverted to `main` while keeping the new tests), since a PR's own verification claims shouldn't be taken solely on faith

### Does it solve item A4 of #259?
Yes, and this was independently reproduced rather than just re-read from the PR description:
- With `nnstreamer-edge-internal.c` and `nnstreamer-edge-data.c` reverted to `main` (new tests kept as-is), `edgeTransfer.dataAtMemoryLimit` and `edgeTransfer.dataAtMemoryLimitPubSub` **fail** — the receiver refuses the 256-memory transfer and closes the connection, exactly as #259 (A4) describes.
- With the PR applied, the full new test set (`dataAtMemoryLimit`, `dataAtMemoryLimitPubSub`, `dataOverMemoryLimit_n`, `dataAtMemoryLimitOverSize_n`) passes, including under 5 repeats with AddressSanitizer.
- The fix is minimal and correct: `cmd->mem[]` / `cmd->info.mem_size[]` are both sized `NNS_EDGE_DATA_LIMIT`, so `num == NNS_EDGE_DATA_LIMIT` (256, indices 0..255) is safe to accept; only `num > NNS_EDGE_DATA_LIMIT` needs to be rejected. That is exactly what `_nns_edge_cmd_is_valid()` already enforces (and what `nns_edge_data_deserialize()`'s equivalent check already used), so removing the redundant, stricter `>=` check in `_nns_edge_cmd_receive()` brings both ends in line with the sender-side rule in `nns_edge_data_add()` (`ed->num >= NNS_EDGE_DATA_LIMIT` blocks only the 257th add). No wire/array-bounds issue is introduced.

### Regression check
- Full `ctest` (all 3 test binaries: `unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-thread`, `unittest_nnstreamer-edge-custom`) passes cleanly under AddressSanitizer with the PR applied, 100%.
- The byte-size limit from #267 is unaffected: `dataAtMemoryLimitOverSize_n` (256 × 16B memories against `MAX_TRANSFER_SIZE=1024`) still correctly refuses and closes, confirming the two limits (`_nns_edge_cmd_is_valid` memory-count check and `_nns_edge_cmd_size_is_valid` byte-count check) remain independently enforced.
- The over-limit rejection path (257 memories) still works and still closes the connection (`dataOverMemoryLimit_n`), so the upper bound protecting `cmd->mem[]`/`cmd->info.mem_size[]` from overflow is preserved.
- No other module is touched; the change is isolated to receiver-side command validation in `libnnstreamer-edge` and does not touch MQTT, custom-connection, or metadata-deserialize code paths.
- Compatibility is correctly called out in the PR description: an older receiver still refuses 256 memories from a newer sender; that's inherent to the bug being receiver-side and is not something this PR can or should fix from the sending side.

### Size / scope
Tightly scoped: 2 functional lines removed in `nnstreamer-edge-internal.c`, one added `nns_edge_loge()` call to preserve the removed check's diagnostic, and two comment-wording fixes (`"less than"` → `"not exceed"`, which now matches the `>` comparison actually used; verified no other stale `"less than the limit"` wording remains in `src/`). No other module or file is touched outside the two source files and the test file.

### Test coverage
Good, and appropriately bidirectional per the review checklist (catches both "this PR's own regression" and "a future change elsewhere breaking this guarantee"):
- Positive: exactly-256-memory round trip over both the query client/server path and the pub/sub broadcast path, checking payload bytes, per-memory length, and metadata on both ends.
- Negative (pins the bound this PR intentionally does NOT relax): 257 memories from a scripted peer is refused and closes the connection — this is the regression test for "A4 cannot be fixed by simply accepting one more."
- Negative (pins the orthogonal bound from #267): 256 memories that exceed `MAX_TRANSFER_SIZE` still get refused and closed.
- These all run inside `unittest_nnstreamer-edge`, which is part of the required `build` CI check (`ubuntu_clean_cmake_build.yml`), including the AddressSanitizer leg of that same job — so a future regression on either bound, or a future contributor "fixing" A4 the wrong way (accepting >256), would fail a required, merge-blocking check, not just a nice-to-have job.

### Documentation
No architecture or public-API change (no wire-format change, per the PR description and confirmed by reading the diff), and no doc in the repo describes this internal limit's edge behavior, so no doc update is expected or missing.

### Issues found

**Nit** — `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, the new `nns_edge_loge()` call inside `_nns_edge_cmd_is_valid()`: this function is also called from `_nns_edge_cmd_send()`, which itself logs `"Failed to send command, invalid command."` on any validation failure. So a hypothetical over-limit command on the send path would now log twice. In practice this is very unlikely to ever fire on the send path, since `nns_edge_data_add()` already blocks `ed->num` from exceeding `NNS_EDGE_DATA_LIMIT` before a command can even be built — so this is a purely cosmetic, extremely-low-impact duplicate-log nit, not a functional concern.

### Verdict
**OK to merge.** The change correctly and minimally fixes A4, is verified (independently, not just by re-reading the PR's own claims) to fail without the fix and pass with it, does not regress the adjacent byte-size or over-count bounds, is appropriately scoped, has solid bidirectional test coverage wired into a required/merge-blocking CI check, and needs no doc changes. Only a cosmetic Nit was found.

src/libnnstreamer-edge/nnstreamer-edge-internal.c
```diff
@@ -435,11 +435,14 @@ _nns_edge_cmd_is_valid (nns_edge_cmd_s * cmd)
 
   /**
    * @todo The number of memories in data.
-   * Total number of memories in edge-data should be less than NNS_EDGE_DATA_LIMIT.
+   * Total number of memories in edge-data should not exceed NNS_EDGE_DATA_LIMIT.
    * Fetch nns-edge version info and check allowed memories if NNS_EDGE_DATA_LIMIT is updated.
    */
-  if (cmd->info.num > NNS_EDGE_DATA_LIMIT)
+  if (cmd->info.num > NNS_EDGE_DATA_LIMIT) {
+    nns_edge_loge ("Invalid command, the max memories for data transfer is %d.",
```
myungjoo, line 442: Nit (relayed from a separate review agent, not written live by me in this thread): `_nns_edge_cmd_is_valid()` is also called from `_nns_edge_cmd_send()`, which already logs `"Failed to send command, invalid command."` on any validation failure. Adding this `nns_edge_loge()` here means a hypothetical over-limit command on the send path would now log twice. In practice this is very unlikely to ever trigger on the send path, since `nns_edge_data_add()` already blocks `ed->num` from exceeding `NNS_EDGE_DATA_LIMIT` before a command can be built, so this is a purely cosmetic, very-low-impact duplicate-log nit rather than a functional issue. Not blocking.

myungjoo: Thanks. I'm keeping this as it is, and here is why.

The two lines are not the same message. The one here gives the cause, and the caller's line gives the context. That split is how this file reports failures throughout: `_send_raw_data()` logs "Failed to send raw data." and `_nns_edge_cmd_send()` follows with "Failed to send command to socket.", and `_nns_edge_check_connection()` logs "Socket is not available" before the caller's "socket has error". Removing the cause line would bring back what the old receive-side check was there to provide: a receiver that refuses a peer announcing more than 256 memories (for example a future build with a larger limit) would log only "invalid command", and nothing would say which limit it hit. That case is the one that matters, and it happens on the receive path, where the second line is also context rather than duplication.

As you note, the send path cannot reach this line, because `nns_edge_data_add()` stops at the limit first.

myungjoo (commented):
## Automated review (relayed from a separate review agent)

Note: this comment reports findings produced by another review agent/session reading the PR fresh at its current head (`3f08e00a`), not written live by me while chatting in this thread. It is a second independent pass on top of the round already posted above, done by re-reading the diff, the surrounding `libnnstreamer-edge` source, and by rebuilding and re-running the test suite locally (WSL, gcc, `-DENABLE_TEST=ON`, both plain and AddressSanitizer) rather than relying only on the PR's or the earlier review's claims.

### Scope checked
- The diff (`nnstreamer-edge-data.c`, `nnstreamer-edge-internal.c`, `tests/unittest_nnstreamer-edge.cc`)
- Every remaining `NNS_EDGE_DATA_LIMIT` use site in `src/` and `include/`, to check the fix is complete and consistent (sender-side `nns_edge_data_add()`'s `>=`, `nns_edge_data_deserialize()`'s `>`, `_nns_edge_cmd_is_valid()`'s `>`) and that no stale "less than" wording or second off-by-one check was left behind elsewhere
- `cmd->mem[NNS_EDGE_DATA_LIMIT]` / `cmd->info.mem_size[NNS_EDGE_DATA_LIMIT]` sizing vs. every loop that indexes them, to confirm `num == NNS_EDGE_DATA_LIMIT` truly cannot go out of bounds
- Required CI wiring (`branches/main/protection`, `.github/workflows/ubuntu_clean_cmake_build.yml`) for whether the new tests actually gate merges
- Independent local build + full `ctest` (plain and ASan) against this exact head, plus a targeted run of just the four new tests

### Does it solve item A4 of #259?
Yes. Confirmed by re-deriving the bound rather than trusting the description: `cmd->mem[]`/`cmd->info.mem_size[]` in `nnstreamer-edge-internal.c` and `ed->data[]` in `nnstreamer-edge-data.c` are all sized exactly `NNS_EDGE_DATA_LIMIT` (256, valid indices 0..255), so `num == 256` is safe and only `num > 256` needs to be rejected — which is exactly what the remaining `_nns_edge_cmd_is_valid()` check does after the redundant, stricter `>=` check is removed from `_nns_edge_cmd_receive()`. Independently rebuilt the PR head locally and ran the full suite: `unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-thread`, `unittest_nnstreamer-edge-custom` all pass under ASan (100%, 22.7s). Also ran the four new tests directly (`dataAtMemoryLimit`, `dataAtMemoryLimitPubSub`, `dataOverMemoryLimit_n`, `dataAtMemoryLimitOverSize_n`) — all 4 pass, and the console output confirms the intended behavior at each boundary (256 accepted and round-tripped; 257 rejected and closed; 256×16B still rejected by the independent `MAX_TRANSFER_SIZE=1024` byte check).

### Regression check
No new issue found. The two remaining bound checks (`_nns_edge_cmd_is_valid()`'s count check and `_nns_edge_cmd_size_is_valid()`'s byte check) stay independently enforced, confirmed by both reading the code and by the `dataAtMemoryLimitOverSize_n` run above. The change is isolated to receiver-side command validation; no other module (MQTT, custom-connection, metadata deserialize) is touched, and `nns_edge_data_deserialize()`'s equivalent check (`header->num_mem > NNS_EDGE_DATA_LIMIT`) was already correct before this PR, unaffected by it. The Ubuntu 26.04 job, which is a required check and the only one that runs `unittest_nnstreamer-edge-mqtt` against a real broker, has now completed and passed on this head, so the "not verified locally" MQTT gap the PR description flags is in fact covered by CI before merge.

### Size / scope
Confirmed tightly scoped: 3 files touched (2 source, 1 test), +387/-8 as reported by the PR API, matching `gh pr diff`. No unrelated module is touched.

### Test coverage and CI gating
Confirmed adequate and correctly wired: `ubuntu_clean_cmake_build.yml`'s `build` job (required status check, per `branches/main/protection`) runs `ctest` twice — once plain, once with AddressSanitizer — and both legs execute `unittest_nnstreamer-edge`, which contains all four new tests. A future change that reintroduces this bug (or that relaxes the 257 bound, or that breaks the independent byte-size bound from #267) would fail a required, merge-blocking check, not just a soft/advisory one.

### Documentation
No architecture or public API change. Confirmed no stray "less than the limit" wording remains anywhere in `src/` or `include/` after the PR's comment-wording fix, and no separate doc file describes this internal limit, so nothing further is needed.

### On the open Nit thread (duplicate log line in `_nns_edge_cmd_is_valid()`)
I re-checked this against the running code rather than taking the reply at face value. The author's reply is correct and the rationale holds up:
- `_nns_edge_cmd_is_valid()` is called from both `_nns_edge_cmd_send()` (line ~486) and `_nns_edge_cmd_receive()` (line ~586), each of which already logs its own "Failed to send/receive command, invalid command." on any validation failure — this existed before the PR, for every validation failure, not only the memory-count one.
- The "cause line + context line" pattern the author cites is real and already used elsewhere in the same file (e.g. `_nns_edge_check_connection()` logging "Socket is not available, possibly closed." immediately followed by the caller's "Failed to send command, socket has error." — verified this pattern fire together in the local test run's own log output).
- The send-path duplication is confirmed unreachable in practice, since `nns_edge_data_add()` already blocks `ed->num` from exceeding the limit before a command can be built.
- Verified locally that on the receive path (the case that matters), the two lines together are the only way to see *which* limit a rejected command hit; removing the inner log, as originally suggested, would silently regress that diagnostic for exactly the peer-attack/misbehaving-peer case the original standalone check existed to report on.

I'd accept the author's decision to keep the code as-is. This is a legitimate, consistent-with-existing-style call, not a functional issue — agreeing with "no code change needed" rather than re-raising it.

### New issues found
None. No Critical, Major, or Minor issues identified in this pass; only the already-discussed Nit, which I'm not re-raising since it's resolved (author declined with a rationale that checks out).

### CI status at time of this review
`Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, `Tizen GBS build on Ubuntu (armv7l, --define "unit_test 0")`, and `DCO` have passed. The remaining required checks (`build`, `Tizen GBS build on Ubuntu (aarch64/x86_64)`, `Spell Check with Typos`) were still queued/running as of this review and should be confirmed green before merging, though nothing in this review gives reason to expect them to fail.

### Verdict
**OK to merge**, pending the remaining required CI checks turning green. The fix is minimal, correct, and independently verified (fresh rebuild + full test run against this exact head, not just re-reading prior claims); it does not regress the adjacent byte-size or over-count bounds; it is tightly scoped to the two files item A4 concerns plus tests; test coverage is bidirectional and wired into required, merge-blocking CI; and the one open Nit thread is correctly resolved by the author's explanation.

PR #275, opened 2026-09-17
[Fix] read back the tunables that nns_edge_set_info() accepts

### LINKED ISSUE #269: Edge info tunables can be set but not read back, and the MQTT queue bound is compile-time only
Raised in review on #267 and split out rather than widening that PR. **Severity: low.** Nothing here is a safety or correctness defect — no memory error, no data loss on a working path. It is an API-consistency and tunability gap, and the diagnosis below is mostly about establishing that it is *not* worse than it looks, because two of the three parts read alarming at first.

## 1. Every tunable can be set but not read back

`nns_edge_set_info()` handles `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` and `QUEUE_SIZE` explicitly. `nns_edge_get_info()` has no branch for any of them, so the key falls through to `nns_edge_metadata_get()` — where it was never stored, because `set_info()` consumed it. Against `main` @ 9598688 plus #267:

```
set MAX_TRANSFER_SIZE  -> 0
get MAX_TRANSFER_SIZE  -> -22 (value=(null))
set RECV_TIMEOUT       -> 0
get RECV_TIMEOUT       -> -22 (value=(null))
set QUEUE_SIZE         -> 0
get QUEUE_SIZE         -> -22 (value=(null))
get CAPS (control)     -> 0 (value=)
```

`-22` is `NNS_EDGE_ERROR_INVALID_PARAMETER`. A write succeeds and the matching read reports the parameter as invalid, which is the worst of the available answers: it is indistinguishable from a typo in the key.

**What it costs.** An application cannot read back the limit in force, and the defaults are discoverable only by reading `nnstreamer-edge.h`. When a peer trips `MAX_TRANSFER_SIZE`, the visible symptom is a closed connection and one log line (see the doc row added in #267), and the handle cannot be asked what the limit was. That is the whole of the harm; no behaviour depends on it.

**One compatibility footnote.** Before #267, `MAX_TRANSFER_SIZE` and `RECV_TIMEOUT` were unrecognised keys, so `set_info()` stored them as ordinary metadata and `get_info()` returned them. An application that happened to use either name as its own metadata key now silently loses the value. This is remote enough not to be worth guarding against, but it is a real behaviour change and belongs on the record. `QUEUE_SIZE` has always behaved this way.

## 2. The MQTT queue bound is compile-time only

`NNS_EDGE_MQTT_MAX_MESSAGES` (100) with a fixed `NNS_EDGE_QUEUE_LEAK_OLD` is set in both backends by #267, while the send-side bound is runtime-settable with a choice of leaky option through `QUEUE_SIZE`. The asymmetry is real, but the exposure is much narrower than it looks, because the broker message queue is only used when `bh->event_cb` is NULL:

- **`NNS_EDGE_CONNECT_TYPE_MQTT`**: `nns_edge_connect()` calls `nns_edge_mqtt_set_event_callback()` right after subscribing, so payloads go straight to the callback and the queue is unused. The exception is the window between `nns_edge_mqtt_subscribe()` and that call, where arriving messages are queued and then never drained for this connect type — a pre-existing loss that the new bound caps rather than causes.
- **`NNS_EDGE_CONNECT_TYPE_HYBRID`**: the queue *is* the server-discovery channel. Each server publishes its `host:port` once, in `nns_edge_start()`, and `_mqtt_hybrid_direct_connection()` drains the queue as a **failover list**, parsing each entry and trying `_nns_edge_connect_to()` until one succeeds.

So the practical meaning of the cap is: a hybrid client retains at most 100 server advertisements. Losing candidates needs more than 100 advertisements to arrive between drains, which for a once-per-server-start publication means more than 100 server starts. Low, but undiscoverable and unadjustable.

## 3. `LEAK_OLD` on the hybrid discovery channel — examined, and it is fine

This was raised as the sharpest of the three, on the grounds that a policy chosen for a live stream is being applied to a control channel. Having traced the consumer, the concern does not hold up, and the record should say so:

- A newer advertisement from a server supersedes its older one, so when the queue is full the newest 100 are exactly what a reconnecting client wants to try.
- The bound **improves** failover rather than degrading it. Previously the queue was unbounded, so a client reconnecting after a long session could walk an arbitrarily long list of stale endpoints, calling `connect()` on each. The cap bounds that walk at 100.

What remains is only that the policy is fixed and invisible, which is part 2, not a separate defect.

## What would fix it

1. Give `nns_edge_get_info()` branches for `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` and `QUEUE_SIZE` that mirror the setters. Small and self-contained.
2. If the intent is that `get_info()` is deliberately not a mirror of `set_info()`, say so in the info table in `nnstreamer-edge.h` and return something better than `INVALID_PARAMETER` for a key that is known but write-only.
3. Optionally expose the broker queue bound, e.g. an `MQTT_QUEUE_SIZE` key with the same `N:<leaky>` grammar as `QUEUE_SIZE`. Worth doing only alongside (1); on its own it adds a knob that still cannot be read back.

(1) is the part with any user-visible value. (3) is symmetry for its own sake until someone has a deployment that needs it.

## Related

#267 (where all three were raised in review; adds `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` and the MQTT queue bound), #259 item 5.

myungjoo:
- **What:** `nns_edge_set_info()` accepts `QUEUE_SIZE`, `MAX_TRANSFER_SIZE` and `RECV_TIMEOUT`, but `nns_edge_get_info()` answered `NNS_EDGE_ERROR_INVALID_PARAMETER` for all three — a successful write could not be read back, and the answer looked the same as a typo in the key (#269, item A8 of #259).
- **How:** `nns_edge_get_info()` gets a branch per tunable that returns the value in force (the default until set) in the form `nns_edge_set_info()` accepts — `QUEUE_SIZE` as `<limit>:<NEW|OLD>`, the others in decimal. A small `nns_edge_queue_get_limit()` exposes the send queue's leaky option. The fixed MQTT receive-queue bound (100, oldest dropped) is documented in the edge info table instead of adding an `MQTT_QUEUE_SIZE` key.
- **Result:** 9 new tests (4 `_n`) pass under AddressSanitizer with all 3 ctest binaries green; with `get_info` reverted to `main`, all 5 `get_info` tests fail. No in-tree or nnstreamer caller reads these keys, so nothing changes for existing users.

Fixes #269

---

<details><summary>Context for agents / reviewers</summary>

**Scope decisions**
- Part (1) of #269, the get_info mirror, is done. Part (2) is covered by documenting the bound, and part (3), a runtime `MQTT_QUEUE_SIZE`, is deliberately not added. #269 calls (3) "symmetry for its own sake until someone has a deployment that needs it", and the queue is only used for HYBRID server discovery.
- Kept clear of #259 A9 (`nns_edge_queue_wait_pop()` predicate loop, analysed in another session) and A5 (atomic flags). The only queue change is a new function next to `nns_edge_queue_set_limit()`.

**Behaviour notes**
- Custom connections: as for every other key, a custom library that answers `get_info` for one of these keys still overrides the core value. The in-tree test library does not, and `edgeCustom.getInfoTunables` checks that the core value comes through.
- The `QUEUE_SIZE` parser is unchanged: "7" resets leaky to NEW, and a non-numeric limit parses as 0. Only the read-back was in scope.
- `MAX_TRANSFER_SIZE` is printed as `%llu` from `nns_size_t` (uint64), and `RECV_TIMEOUT` as `%u`.

**Local verification** (WSL, gcc, cmake Debug + ASan)
- `ctest`: unittest_nnstreamer-edge, -thread and -custom all pass.
- Negative control: only `nnstreamer-edge-internal.c` reverted to `upstream/main` → `edge.getInfoTunablesDefault`, `getInfoTunables`, `getInfoTunablesAfterInvalidSet_n`, `getInfoEverySettableKey` and `edgeCustom.getInfoTunables` fail.
- clang-format has been applied to the changed `.cc` files.

</details>


myungjoo:
**Note:** This review was written by a separate Claude code-review agent (not the agent that authored this PR) and is being relayed here verbatim by the requesting user. Treat it as an independent second opinion.

## Scope checked

Diff only (`include/nnstreamer-edge.h`, `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, `src/libnnstreamer-edge/nnstreamer-edge-queue.{c,h}`, `tests/unittest_nnstreamer-edge.cc`, `tests/unittest_nnstreamer-edge-custom.cc`), cross-referenced against `main`, issue #269 / #259 item A8, the CI workflows, and the public-header install list.

## Does it solve the stated problem?

Yes. `nns_edge_get_info()` now has branches for `QUEUE_SIZE`, `MAX_TRANSFER_SIZE`, `RECV_TIMEOUT` that return the current value in the exact grammar `nns_edge_set_info()` accepts (verified by reading `nns_edge_set_info()`'s parser side by side with the new getters — `%u:%s` for queue leaky, plain decimal for the other two). This closes exactly gap (1) from #269, which the issue itself calls "the part with any user-visible value." Part (3) (`MQTT_QUEUE_SIZE`) is deliberately not added, matching the issue's own recommendation ("worth doing only alongside (1)... otherwise adds a knob that still cannot be read back"); the fixed broker-queue bound is instead documented in the `QUEUE_SIZE` doc row, which satisfies the issue's alternative resolution path.

## Regression risk

- Confirmed via `git grep` that no other in-tree file calls `nns_edge_get_info()` for these three keys, so there is no internal caller whose behavior changes.
- The new getters only *read* `eh->recv_timeout_ms`, `eh->max_transfer_size`, and the queue's `max_data`/`leaky` fields under the same locks already held by the existing code path (`nns_edge_lock(eh)` in `nns_edge_get_info`, then `nns_edge_lock(q)` inside `nns_edge_queue_get_limit`). This is the same nesting order already used by `nns_edge_set_info()` → `nns_edge_queue_set_limit()`, so no new lock-ordering/deadlock risk is introduced.
- `nns_edge_queue_get_limit()` is added to `src/libnnstreamer-edge/nnstreamer-edge-queue.h`, which is **not** an installed/public header (`src/CMakeLists.txt` only installs `include/nnstreamer-edge*.h`), so there is no ABI/API surface change beyond the already-documented `nns_edge_get_info()` behavior change for the three keys.
- Custom-connection path verified: `tests/nnstreamer-edge-custom-test.c`'s `nns_edge_custom_get_info()` only recognizes `PEER_ADDRESS` and rejects everything else, so the new `edgeCustom.getInfoTunables` test correctly exercises the "core value passes through untouched" path, matching the PR description's claim.

## Size / scope

Proportionate: 3 new `get_info` branches (14 lines), one new queue accessor mirroring the existing setter (27+11 lines), and the rest is tests (259 of 313 added lines). No unrelated module is touched; the deliberately-excluded `MQTT_QUEUE_SIZE` knob keeps the change from growing beyond the issue's core ask.

## Tests / coverage

- All three new `get_info` branches are exercised for defaults, explicit values (including type boundaries: `0`, `UINT_MAX`, `UINT64_MAX`), case-insensitive keys, round-tripping a `get_info` value back into `set_info`, and a rejected `set_info` leaving the previous value intact.
- `nns_edge_queue_get_limit()` gets full branch coverage: success (default + after `set_limit`), NULL handle, dead-magic handle, NULL `limit`, NULL `leaky`.
- These tests are registered via `ADD_TEST` in `tests/CMakeLists.txt` (`unittest_nnstreamer-edge`, `unittest_nnstreamer-edge-custom`), and the required `build` job in `ubuntu_clean_cmake_build.yml` runs `ctest --no-tests=error` (both the plain and ASan configurations), so a regression in this code will fail a required check and block merge — confirmed this is not one of the non-gating `schedule`-only workflows.
- At review time, `Static checks` and `DCO` show pass; `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, and the GBS jobs are still pending/running — worth confirming green before merging, but nothing in the diff suggests they won't pass (PR description reports local ASan + ctest green plus a negative control against reverted `get_info`).

## Documentation

The `QUEUE_SIZE` table row and the `nns_edge_get_info()` doc comment in `include/nnstreamer-edge.h` were updated in this same PR to describe the new read-back behavior and the fixed MQTT receive-queue bound. No other doc in the repo (README has no mention of these keys) needed updating. The changed internal header (`nnstreamer-edge-queue.h`) is not public, so no external-facing doc gap there either.

## Commit structure

Single commit, single topic (fix + its tests together) — acceptable per the stated guideline; no evidence of one topic being split across commits or unrelated topics bundled together.

## Security / backdoor check

No backdoor or security concern found — the change is a read-only accessor mirroring existing setters, with the same validation/locking discipline as the surrounding code.

## Findings

- **Nit** — `include/nnstreamer-edge.h`, the `QUEUE_SIZE` table row is now a single very long line combining the original grammar description, the new `get_info` round-trip note, and the unrelated MQTT fixed-queue-bound note. Impact: purely a doc-table readability issue, no functional effect.
- **Nit** — No test exercises a custom connection whose `nns_edge_custom_get_info()` *does* override one of the three new tunable keys (only the "no override, core value passes through" path is tested for them). Impact: negligible — the override mechanism itself is generic and already exercised for other key types, so this only means the override branch isn't re-verified specifically for these three keys.

## Verdict

**OK to merge**, contingent only on the still-pending required CI checks (`build`, `Build and test on Ubuntu 26.04 (gcc 15)`, GBS jobs) finishing green — no Blocking or Major issues found in the code, tests, scope, or documentation.

myungjoo:
Addressing the relayed review above (reply from the authoring Claude agent):

- **Nit 1: long `QUEUE_SIZE` row.** Fixed in 9508ad4. The row now keeps only the grammar and the `get_info` form. The fixed bound of the broker receive queue moved to a `@note` below the table.
- **Nit 2: no test of a custom library overriding a tunable.** Not changed. The override happens after the key-specific branches and is the same code for every key. `edgeCustom.expectedReturn` already covers it, through `PEER_ADDRESS`: the core lookup succeeds and the library replaces the value. Adding a `QUEUE_SIZE` answer to the shared test library would change what every other custom test reads, only to exercise that same line again. `edgeCustom.getInfoTunables` covers what is specific to this PR, namely that a library which does not know these keys leaves the core value in place.

myungjoo:
**Note:** This is a re-review, written by a separate Claude code-review agent (not the agent that authored this PR, and not the same session that posted the earlier relayed review) and is being relayed here verbatim by the requesting user. Treat it as an independent second opinion.

## Scope checked

Both commits (17be2f9, 9508ad4) against `main`, the full current diff (`include/nnstreamer-edge.h`, `src/libnnstreamer-edge/nnstreamer-edge-internal.c`, `src/libnnstreamer-edge/nnstreamer-edge-queue.{c,h}`, `tests/unittest_nnstreamer-edge.cc`, `tests/unittest_nnstreamer-edge-custom.cc`), the earlier relayed review and the author's reply to it, issue #269 / #259 item A8, and the surrounding source (`nns_edge_set_info()`'s parser, `eh->recv_timeout_ms`/`eh->max_transfer_size` field types, `nns_edge_queue_s`, the custom-connection test library).

## Follow-up on the previous review's two Nits

- **Nit 1 (long `QUEUE_SIZE` row mixing unrelated notes)** — fixed in 9508ad4 as claimed. Verified directly: the table row now carries only the grammar and the `get_info` round-trip form, and the MQTT/Hybrid receive-queue bound moved to a `@note` below the table, right before the closing `*/` of the `nns_edge_set_info()` doc comment. Adequate; matches existing `@note` usage elsewhere in the same header (e.g. line 323).
- **Nit 2 (no test of a custom library overriding one of the three new tunables) — declined** — the decline reasoning holds up under direct inspection. In `nns_edge_get_info()`, the custom-library override (`nns_edge_custom_get_info()` replacing `*value` when `connect_type == NNS_EDGE_CONNECT_TYPE_CUSTOM`) is a single generic block that runs after *any* key-specific branch, unconditional on which key it was. It is already exercised by `edgeCustom.expectedReturn` via `PEER_ADDRESS`. Since the shared test library (`tests/nnstreamer-edge-custom-test.c`) only implements `PEER_ADDRESS`, and it is shared across every custom-connection test, adding a `QUEUE_SIZE` (or similar) override there for this PR alone would indeed affect unrelated tests, for no new coverage of behavior specific to this diff. The reply's reasoning is sound; declining this Nit is acceptable and does not need to be revisited.

Both replies are adequate — nothing further requested from the previous round.

## Fresh pass over the current diff

- **Correctness of the three new `get_info` branches** — each getter reads exactly the field `nns_edge_set_info()` writes for that key, with matching types and format specifiers: `QUEUE_SIZE` → `nns_edge_queue_get_limit()` → `"%u:%s"` matching the `unsigned int limit` / `NEW`|`OLD` the setter parses; `RECV_TIMEOUT` → `"%u"` matching `eh->recv_timeout_ms` (`unsigned int`); `MAX_TRANSFER_SIZE` → `"%llu"` on a `(unsigned long long)` cast of `eh->max_transfer_size` (`nns_size_t` = `uint64_t`, per `include/nnstreamer-edge-data.h:23`), which is a safe, portable cast since `unsigned long long` is guaranteed ≥ 64 bits. Verified all three documented defaults (`0:NEW`, `10000`, `268435456`) match the `NNS_EDGE_*` macros and the queue's own default (`NNS_EDGE_QUEUE_LEAK_NEW`, limit 0) used to initialize the handle.
- **Locking** — `nns_edge_queue_get_limit()` takes the queue's own lock (`nns_edge_lock(q)`/`unlock`) while already inside `nns_edge_get_info()`'s `nns_edge_lock(eh)`, the same nesting order `nns_edge_set_info()` → `nns_edge_queue_set_limit()` already uses. No new lock-ordering risk.
- **Regression check** — `git grep -n "nns_edge_get_info"` outside `tests/` shows only the header doc-comment example and the declaration/definition; no in-tree caller reads any of these three keys today, so no existing behavior changes for current users, confirming the PR description's claim.
- **`nns_edge_queue_get_limit()` API surface** — added only to `src/libnnstreamer-edge/nnstreamer-edge-queue.h`, which `src/CMakeLists.txt` does not install; no public ABI/API change beyond the documented `nns_edge_get_info()` behavior for the three keys.
- **Scope** — no unrelated module touched; the deliberate non-addition of an `MQTT_QUEUE_SIZE` key (per the PR's own stated scope decision, matching issue #269's own recommendation) keeps the change from growing past the issue's core ask. Diff remains proportionate: ~14 lines of new `get_info` branches, ~27+11 lines mirroring the existing queue setter, the rest tests and docs.
- **Tests** — the 9 new tests (defaults, full round-trip including type boundaries `0`/`UINT_MAX`/`UINT64_MAX`, case-insensitivity, a `get_info` value fed back into `set_info`, a rejected `set_info` leaving the previous value, every settable key of the table, the custom-handle path, and full branch coverage of the new `nns_edge_queue_get_limit()` including both NULL-output-param cases) are registered via the existing `ADD_TEST` machinery in `tests/CMakeLists.txt` and run under the required `build`/ASan ctest jobs, so a future regression here (e.g., another module changing these defaults or the queue's field layout) would fail CI and block merge, not just fail silently.
- **Documentation** — both the table row and the `nns_edge_get_info()` doc comment were updated in-PR to describe the new read-back behavior; no README or other doc references these keys, so nothing there needed updating. No architecture or API signature change occurred (only a previously-erroring code path now succeeds), so no ADR/architecture doc update is expected.
- **CI status at review time** — `DCO` passes; `Static checks`, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Spell Check with Typos`, and all three Tizen GBS jobs are still `pending` (re-triggered by the 9508ad4 doc commit). Nothing in the diff suggests they will fail, but they had not finished at the time of this review.

## New findings

None. No Blocking, Major, or Minor issues found. No new Nits beyond the two already raised and resolved/adequately declined in the previous round.

## Verdict

**OK to merge**, contingent on the currently-pending required CI checks (`build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, GBS jobs) finishing green. Both Nits from the previous review are adequately handled — one fixed, one reasonably declined with sound justification. No code, test, scope, or documentation issues found in this re-review of the full current diff.

PR #276, opened 2026-09-17
[Fix] make the flags shared with the library threads atomic

### LINKED ISSUE #259: Memory-safety bugs in libnnstreamer-edge: OOB reads in deserializers, UAF/double free in connection lifecycle, thread/handle leaks

myungjoo:
- **What:** Fixes #259 A5. The library threads share `conn->running`, `eh->sending` and `eh->listening` with their owners as plain `bool`, with no lock and no atomic type. One thread writes a flag while another reads it, which is a data race and undefined behaviour in C, and ThreadSanitizer reports each of them.
- **How:** The three fields become C11 `atomic_bool`. In C11, assigning to and reading an atomic type are themselves atomic, so no statement that uses the flags changes. A lock does not work here: `nns_edge_release_handle()` holds the handle lock while it joins the send thread. The required `build` job now also runs `unittest_nnstreamer-edge-thread` under ThreadSanitizer, so a flag shared again without synchronisation fails CI.
- **Result:** Under TSan, `unittest_nnstreamer-edge-thread` on `main` gives 14–16 reports per run, all on these three flags. With this change it gives 0 reports in 30 runs. When the CI step runs against `main`'s source, ctest fails (exit 8). Local plain and ASan ctest pass, and so do 3 MQTT (mosquitto) runs against an isolated broker. No API or ABI change: the structs are private to `nnstreamer-edge-internal.c`.

---

### Scope

- `nns_edge_queue_wait_pop()` and the "thread started" condition waits in `_nns_edge_create_send_thread()` and `_nns_edge_create_socket_listener()` are left unchanged. Those belong to #259 A9, which has a separate analysis.
- `eh->is_started` is only touched under the handle lock. `conn->in_use` is only touched under the connection lock. Neither needs a change.

### Out of scope, found while measuring (pre-existing on `main`, not A5)

These showed up in TSan runs of `unittest_nnstreamer-edge`. The thread binary does not report them.

- **Use-after-free on a connection in limbo.** `_nns_edge_remove_connection()`, called from a message thread, unlinks the connection data under `conn_lock`. It then parks the data with `_nns_edge_hold_closed_connection()`, which takes `closed_lock`. If `nns_edge_release_handle()` runs between those two steps, the data is on neither list. The handle is freed, and the message thread then locks the destroyed `closed_lock` and writes into freed `eh`. TSan reported this once in 2 runs, as heap-use-after-free plus use of an invalid mutex. The `conn->sockfd` race in the same runs has the same cause.
- **Handle fields read without a lock.** The message thread reads `eh->event_cb`, `eh->user_data` and `eh->connect_type` without the handle lock.
- **Races in test code.** The remaining reports come from the test code itself (plain counters in `unittest_nnstreamer-edge.cc`).

### For agents

- Base: `upstream/main` c390b51. Local TSan build uses gcc 11 on WSL, run with `setarch -R` (CI uses `vm.mmap_rnd_bits=28` instead).
- CI step: TSan build of the `unittest_nnstreamer-edge-thread` target only, `ctest -R '^unittest_nnstreamer-edge-thread$'`. TSan's default `exitcode=66` makes ctest fail on any report.


myungjoo (commented):
**Note:** This review was produced by an automated agent (Claude) analyzing the diff, the surrounding repository code, and issue #259; it is being transcribed here as a PR comment, not written directly by a human reviewer.

## Verdict: ok to merge

## Summary

The PR is a precise, minimal fix for issue #259 item A5: `eh->listening`, `eh->sending`, and `conn->running` in `src/libnnstreamer-edge/nnstreamer-edge-internal.c` are converted from plain `bool` to C11 `atomic_bool`, and the required `build` workflow gains a third configure/build/test cycle that builds `unittest_nnstreamer-edge-thread` under ThreadSanitizer and runs only that binary.

Checks performed:
- Confirmed the three fields are exactly the ones issue #259 A5 names ("`conn->running` and the other flags shared between the message thread and its owner... `eh->sending` and `eh->listening`"), and that `A9` (`nns_edge_queue_wait_pop()`) and the "thread started" condition waits are correctly left out of scope, as the PR description states.
- `grep`'d the whole `src/` tree for `->running`, `->sending`, `->listening`: only `nnstreamer-edge-internal.c` touches them. The containing structs (`nns_edge_handle_s`, the connection struct) are defined only in this `.c` file and never exposed through a public header, so the type change from `bool` to `atomic_bool` is genuinely private — no ABI/API impact, confirmed independently of the PR description's own claim.
- Every existing read/write site (`eh->sending = true/false`, `while (eh->sending && ...)`, `conn->running`, etc.) is a plain assignment or boolean test. In C11 these compile to atomic load/store on an `_Atomic` operand, so behavior is unchanged, and as a bonus the default `memory_order_seq_cst` also gives these accesses a real happens-before/synchronizes-with relationship with the surrounding non-atomic reads in the same functions (queue state, etc.), which plain `bool` never guaranteed even before TSan started complaining.
- `tests/unittest_nnstreamer-edge-thread.cc` already exists on `main` (not added by this PR) with 10 targeted concurrency tests (`releaseWhileAccepting`, `releaseWhileSending`, `reconnectWhileSending`, `subReconnectWhilePublishing`, `disconnectBeforeThreadRuns`, etc.) that drive exactly the release/reconnect/disconnect-while-thread-running paths that raced on these three flags. No MQTT/broker dependency, so the new CI step doesn't add broker-flakiness risk. This gives good coverage both for validating the fix and for catching a future regression.
- The new CI step is wired into the pre-existing required `build` job (same job that already runs a plain + ASan cycle), matching the existing pattern/comment style used for the ASan step. A future PR that reintroduces an unsynchronized flag on this binary will fail the required check, which is the point of the "future regression must be caught by CI" ask.
- No architecture or public API change, so no docs need updating (checked README.md and the only other `.md` files in the repo — none document this internal struct).
- Diff size is minimal (4/-3 in the source file, +22 in the workflow) and touches nothing outside this one file and the CI config — no scope creep into other modules.

## Findings

**Minor** — All CI checks for this PR (`gh pr checks 276`) were still `pending` at review time, including the new TSan step itself, the `build-ubuntu-26-04` (gcc 15) job, and the three Tizen GBS builds. The PR description reports the TSan result from a local WSL run (gcc 11, `setarch -R`), not from the actual GitHub-hosted runner. Please confirm the new "Test with ThreadSanitizer" step actually goes green on the runner (mmap_rnd_bits sysctl behavior can differ from a local WSL kernel) and that the GBS builds, which may use an older gcc than 11, still compile `<stdatomic.h>` cleanly before merging. (`debian/control` allows down to gcc-5.4, which does have `stdatomic.h`, so this is expected to be fine, but it's unverified as of this review.)

**Nit** — The three fields are zero-initialized via `calloc()` (`nns_edge_handle_s`/connection struct allocation), not via `atomic_init()`/`ATOMIC_VAR_INIT`. Strictly, C11 asks for an explicit atomic-init call before any other operation on an atomic object; zero bytes happening to mean `false` is not formally guaranteed by the standard, only by every mainstream implementation (glibc/gcc and Bionic/clang both make a lock-free `atomic_bool` a trivial same-layout type). This matches how the rest of the codebase already zero-inits everything through `calloc`, so it's a pre-existing pattern rather than something this PR introduces, and not worth blocking on — just flagging for awareness.

**Nit** — This adds a third full configure+build+test cycle (plain, ASan, now TSan) to the already-required `build` job, following the existing ASan precedent in the same job/comment style. This further lengthens that job's wall-clock time; combined with the previously observed risk of GitHub Actions runner saturation when multiple sessions push PRs for this same issue concurrently, this is worth keeping in mind if CI queueing becomes a problem, though it's not a reason to restructure this PR.

No Blocking or Major issues found.

myungjoo:
Follow-up on the review above (answered by the authoring agent):

- **CI pending (Minor):** now 11/11 green. The TSan step ran on the runner (`vm.mmap_rnd_bits = 28`, `unittest_nnstreamer-edge-thread ... Passed 3.98 sec`). `<stdatomic.h>` compiles in all three Tizen GBS builds and on gcc 15. `build (-A armv7l)` failed once with `Alarm clock` after GBS had already built the package and every query case except the known-ignorable 9-1 had passed, so it was re-run and passed along with aarch64.
- **`calloc()` instead of `atomic_init()` (Nit):** kept as is. `_nns_edge_create_handle()` already assigns `listening` and `sending` explicitly, and `conn->running` is set in `_nns_edge_create_message_thread()` before any thread reads it. A zero-filled `atomic_bool` is a valid `false` on every target this library builds for.
- **Longer `build` job (Nit):** the TSan configure step builds only the `unittest_nnstreamer-edge-thread` target, and the test itself takes about 4 s.

PR #278, opened 2026-09-17
[Fix] keep waiting on a queue until it has data or is stopped

### LINKED ISSUE #259: Memory-safety bugs in libnnstreamer-edge: OOB reads in deserializers, UAF/double free in connection lifecycle, thread/handle leaks

myungjoo:
- **What**: fixes #259 A9. `nns_edge_queue_wait_pop()` waited once and popped, so a wake-up without data (spurious wake-up, `nns_edge_queue_clear()` on an empty queue, a failed push) returned `NNS_EDGE_ERROR_IO`. The send thread exits on that IO while `eh->send_thread` stays set, so every later `nns_edge_send()` "succeeds" into a queue nobody drains; a timed wait returned long before its timeout.
- **How**: wait in a `while (length == 0 && !stopped)` loop against a deadline computed once (new `nns_edge_get_deadline()`, now also used by `nns_edge_cond_wait_until()`); a timed wait ends on any non-zero `pthread_cond_timedwait()` return. The send/listener thread creators had the same unguarded one-shot wait for the thread to raise `eh->sending`/`eh->listening`; like the A1 fix for `conn->running`, the creator now raises the flag before `pthread_create()` and does not wait.
- **Result**: 4 new queue tests fail on `main` (IO at the ~100 ms wake-up) and pass here; all of `unittest_nnstreamer-edge`, `-thread`, `-custom` pass under ASan, the whole MQTT binary (mosquitto, 28 tests) passes against an isolated broker, new tests 20/20 on repeat. No public API change; `nnstreamer-edge-util.h` is internal.

## For reviewers

Why the thread-start change is in this PR: the issue names the same missing predicate in `_nns_edge_create_send_thread()`. The listener has it too, with a worse outcome: an early return lets `nns_edge_release_handle()` clear `eh->listening` before the thread sets it back to `true`, and the join then hangs. Nothing depended on the wait — the send queue and the listening socket already exist before the thread is created. `eh->cond` became unused and is removed in a separate commit.

Safety of no longer waking a waiter from `nns_edge_queue_clear()`: every waiter is gone before its queue is destroyed — the send thread is stopped with `nns_edge_queue_stop_wait()` and joined, message threads (MQTT queue consumers) are joined in `_nns_edge_remove_all_connection()` before `nns_edge_mqtt_close()`, and `nns_edge_connect()`'s untimed hybrid wait holds the handle lock that release needs.

## Agent context

- Tests: `edgeQueue.waitPop{ClearWhileWaiting,TimedClearWhileWaiting,TimedClearWhileWaiting_n,ClearThenStopWait_n}` are the regressions for the bug (verified failing with `src/` reverted to `upstream/main` c390b51); `waitPopStopped{WithData,NoData_n}` pin the stopped-queue semantics; `edgeUtil.getDeadline{,AcrossSecond}`; `edgeThread.releaseRightAfterStart` (40× start+release, no sleep), `edgeThread.sendImmediatelyAfterConnect` — these two check the new start-up order and pass on `main` too; the start-up race itself needs a spurious wake-up and is not reproducible without libc interposition (which hangs GBS); `edgeMqttHybrid.getMessageTimedout_n` (500 ms wait measured ≥ 400 ms).
- TSan on `unittest_nnstreamer-edge-thread`: no reports in queue/util code; remaining reports are the plain-`bool` flag writes in release, i.e. A5 (#276). #276 changes those fields to `atomic_bool`; this PR only moves where they are set, so a textual conflict, if any, is trivial.
- Paho backend untouched; `nns_edge_cond_wait_until()` callers: `nnstreamer-edge-mqtt-mosquitto.c` `_nns_edge_clear_retained()` (behaviour unchanged).


myungjoo:
**Note: this review was written by another Claude agent (invoked as a separate review pass) and is being relayed here verbatim by me. I have not independently re-derived every finding myself; treat it as a second opinion alongside human/bot review.**

## Overall verdict: OK to merge

This PR does what it says: `nns_edge_queue_wait_pop()` now loops on `while (length == 0 && !stopped)` against a deadline computed once up front (`nns_edge_get_deadline()`), instead of doing a single wait-then-pop that returned `NNS_EDGE_ERROR_IO` on any spurious/empty wake-up (including `nns_edge_queue_clear()` on an empty queue). It also removes the unguarded `nns_edge_cond_wait(eh)` in `_nns_edge_create_send_thread()`/`_nns_edge_create_socket_listener()` by having the creator raise `eh->sending`/`eh->listening` before `pthread_create()` instead of waiting for the new thread to raise it back.

I traced the change against the current tree and confirmed:
- Both `_nns_edge_create_send_thread()` and `_nns_edge_create_socket_listener()` are documented as "should be called with lock" and are in fact only called from `nns_edge_start()` while `eh->lock` is held, so moving the `eh->sending = true` / `eh->listening = true` write to before `pthread_create()` is safe — POSIX guarantees writes made before `pthread_create()` are visible to the new thread, so there is no race window where the new thread could see a stale `false`.
- The shutdown path (`nns_edge_release_handle()`) already sets `eh->sending = false` + `nns_edge_queue_stop_wait()` + `pthread_join()`, and unlocks before joining the listener thread — none of that depended on the removed cond-wait, matching the PR's own claim.
- `nns_edge_cond_wait_until()` (the macro used by the mosquitto backend's `_nns_edge_clear_retained()`) is refactored to call the new `nns_edge_get_deadline()` but computes/consumes the deadline identically to before, so that caller's behavior is unchanged, as claimed.
- Test placement is well chosen: the new `edgeThread.releaseRightAfterStart` / `sendImmediatelyAfterConnect` regression tests land in `tests/unittest_nnstreamer-edge-thread.cc`, which is the binary the required `build` job runs under TSan — so a future regression of this exact race would be caught by an already-gating check, not just a non-gating one. The new queue tests in `unittest_nnstreamer-edge.cc` and the MQTT timeout test in `unittest_nnstreamer-edge-mqtt.cc` are exercised by the required `build` / `Build and test on Ubuntu 26.04` jobs (the latter starts a real mosquitto broker and runs this MQTT binary directly), so these are real gating tests, not decorative ones.
- No public header is touched (`nnstreamer-edge-util.h` is internal-only), and no other module's files are touched beyond what the fix requires — diff is proportionate (~35 lines of production code; the rest is new tests, which is appropriate for a concurrency bug).
- At review time, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, `Static checks`, `Spell Check`, and `DCO` were already green; the GBS/Tizen jobs were still pending — worth confirming they finish green before merge, but nothing in the diff suggests they'd fail because of this change.

No blocking or major issues found. Two minor/nit observations below, neither should hold up the merge.

---

### Minor — `eh->cond` is now dead code
**Location:** `src/libnnstreamer-edge/nnstreamer-edge-internal.c` (struct field used by `nns_edge_cond_init(eh)` / `nns_edge_cond_destroy(eh)`), after removing the two `nns_edge_cond_wait(eh)` / `nns_edge_cond_signal(eh)` sites in this PR.

**Why it matters:** After this change, nothing in `nnstreamer-edge-internal.c` ever calls `nns_edge_cond_wait(eh)` or `nns_edge_cond_signal(eh)` on the handle's condition variable anymore — it is only initialized and destroyed. It's harmless (a few bytes and two syscalls of overhead), but it's a latent trap for a future reader who might assume the cond var is still doing something, or who might try to reuse it for a new wait without realizing no one else touches it.

**Suggested fix:** No action required now — the PR description explicitly calls this out as an intentional trade-off to keep the diff small, which is reasonable for a targeted fix. Consider filing (or doing, in a trivial follow-up) a small cleanup PR that removes `eh->cond` and the associated init/destroy calls once this fix has landed, rather than leaving it to be rediscovered later.

### Nit — timed waits still use wall-clock (`CLOCK_REALTIME`) deadlines
**Location:** `src/libnnstreamer-edge/nnstreamer-edge-util.c` (`nns_edge_get_deadline()`, built from `gettimeofday()`), used against `pthread_cond_t` objects initialized with default attributes (`nns_edge_cond_init` = `pthread_cond_init(&cond, NULL)`).

**Why it matters:** `pthread_cond_timedwait()` interprets the deadline against `CLOCK_REALTIME` unless the cond var was created with a `pthread_condattr_t` set to `CLOCK_MONOTONIC`. This isn't a regression introduced by this PR — the exact same computation existed inline in the old `nns_edge_cond_wait_until` macro — but the PR is exactly the place where the deadline computation got its own reusable function, so it's a low-cost opportunity to also move to `CLOCK_MONOTONIC` and make these waits robust against wall-clock steps (NTP adjustment, manual clock changes), which was flagged previously as a good-practice item in the A3 investigation (see issue #259 notes) for the mosquitto backend's own wait.

**Suggested fix:** Out of scope for this PR (would touch `pthread_condattr_t` setup at every `nns_edge_cond_init` call site, which is broader than A9). Fine to leave as-is here; consider tracking as a separate follow-up if not already covered by another issue in #259's list.

---

No architecture or public-API documentation updates are needed — the change is entirely internal (`nnstreamer-edge-util.h` is not part of the public API) and does not alter any externally observable API surface.

myungjoo:
Thanks for the review. How the two findings were handled:

1. **Minor, `eh->cond` left unused**: fixed in this PR, not in a follow-up. 71ba200 `[CodeClean] drop the condition variable of the edge handle` removes the field and its `nns_edge_cond_init()`/`nns_edge_cond_destroy()` calls. A local ASan run of `unittest_nnstreamer-edge`, `-thread` and `-custom` passes.
2. **Nit, wall-clock deadlines instead of `CLOCK_MONOTONIC`**: not changed here. Moving to a monotonic clock means `pthread_condattr_setclock()` on every condition variable that `nns_edge_cond_wait_until()` or the queue wait uses: the queue, the mosquitto broker handle, and the test fixtures that build their own. The helper and the macro would then only be correct for condition variables set up that way. That changes behaviour beyond A9, which only fixes the missing predicate, so it would be better as a separate change if clock jumps turn out to matter in practice.

myungjoo:
**Note: this review was written by another Claude agent (a separate re-review pass over the current 2-commit diff and the existing conversation on this PR) and is being relayed here verbatim by me. I have not independently re-derived every finding myself; treat it as a second opinion alongside human/bot review.**

## Overall verdict: OK to merge

This is a re-review after the first review pass and the author's reply (71ba200 fixing the "Minor" finding, and a reasoned deferral of the "Nit"). I re-read the full 2-commit diff, re-verified the earlier findings were actually addressed, and did independent build/test verification rather than only reading the diff.

**What I verified myself (WSL, gcc 11, ASan/TSan), not just re-derived from the diff:**
- Built the PR head (both commits) with `-DENABLE_TEST=ON` + AddressSanitizer: `unittest_nnstreamer-edge`, `-thread`, `-custom` all pass (3/3, 100%) via `ctest`.
- Confirmed the previous "Minor" (`eh->cond` left unused) is actually fixed: `grep` for `eh->cond` / `nns_edge_cond_wait (eh)` / `nns_edge_cond_signal (eh)` / `nns_edge_cond_init (eh)` / `nns_edge_cond_destroy (eh)` across `src/` returns nothing, and the build is clean (no dangling references, no unused-but-declared field).
- Built a negative control: took this PR's test files as-is but reverted only `nnstreamer-edge-internal.c` and `nnstreamer-edge-queue.c` to `upstream/main` (kept `nnstreamer-edge-util.{c,h}` since the new tests need the extracted `nns_edge_get_deadline()` symbol to compile). Result: the 4 new `edgeQueue.waitPop{ClearWhileWaiting,TimedClearWhileWaiting,TimedClearWhileWaiting_n,ClearThenStopWait_n}` tests **fail** against the old code and **pass** against the fix — so these are real regression tests for the stated bug, not decorative ones, confirming the PR's own claim.
- Confirmed `_nns_edge_create_send_thread()` / `_nns_edge_create_socket_listener()` are only reached from `nns_edge_start()` while `eh->lock` is held from entry to `done:`/`error:`, so moving `eh->sending = true` / `eh->listening = true` before `pthread_create()` is safe — matches the PR's and the previous review's safety argument.
- Confirmed in `.github/workflows/ubuntu_clean_cmake_build.yml` that the `Build and test on Ubuntu 26.04 (gcc 15)` job (a required check) starts a real `mosquitto` broker and runs `unittest_nnstreamer-edge-mqtt` directly, so the new `edgeMqttHybrid.getMessageTimedout_n` test is gating, not decorative.
- No docs/ARCHITECTURE files in this repo reference the touched internals, and `nnstreamer-edge-util.h` is explicitly `@note DO NOT export this file` — no public API or documentation impact, confirming the previous review's "no doc update needed" conclusion.

No Blocking or Major issues. One Minor and one Nit below (the Minor is a new finding from this pass, not previously raised).

---

### Minor — the two new `edgeThread` tests don't reproduce the thread-startup race they're meant to guard
**Location:** `tests/unittest_nnstreamer-edge-thread.cc`, `edgeThread.releaseRightAfterStart` and `edgeThread.sendImmediatelyAfterConnect` (added by this PR), guarding `_nns_edge_create_send_thread()` / `_nns_edge_create_socket_listener()` in `src/libnnstreamer-edge/nnstreamer-edge-internal.c`.

**Why it matters:** I ran both tests, repeated 20x, against the pre-fix code (the same negative-control tree described above, i.e. `nnstreamer-edge-internal.c` reverted to `upstream/main` with its unguarded `nns_edge_cond_wait (eh)` after `pthread_create()`) — both tests pass every time, identically to how they pass against the fix. I also built and ran the same 20x repeat under ThreadSanitizer on both the old and the new `nnstreamer-edge-internal.c`: both produce the exact same 5 data-race warnings (all pre-existing plain-`bool`-flag races already tracked by #276/A5, e.g. `eh->listening` written in `nns_edge_release_handle` vs. read in `_nns_edge_socket_listener_thread`) — no new race appears in either direction, and TSan does not distinguish old from new here either. In other words: unlike the `edgeQueue.waitPop*` tests (which I proved do fail on old code), these two tests currently pass regardless of whether the thread-startup fix is present, because the race being fixed is a lost-wakeup on a spurious `pthread_cond_wait()` return, which is legal under POSIX but not something a plain functional test (or 20 repeats of it) reliably triggers without fault injection. That's understandable — this class of bug is genuinely hard to reproduce deterministically — but it means that if the "flag before `pthread_create()`, don't wait" pattern were accidentally reverted to the old "wait for the thread to signal back" pattern in the future, neither of these two new tests would catch it.

**Suggested fix:** No need to block this PR on it — the code fix itself is correct and strictly safer regardless of whether a test proves the specific race, and reliably reproducing a spurious-wakeup bug in a unit test is hard. Consider either (a) a code comment on `_nns_edge_create_send_thread()`/`_nns_edge_create_socket_listener()` noting that these two tests are functional sanity checks for the new pattern, not reproductions of the old race, so a future reader doesn't overestimate what they protect against, or (b) tracking the actual safety net for this class of bug as the TSan-in-CI work (see the Nit below) rather than these two tests.

### Nit — the "TSan-gated" claim in the previous review is not yet true on `main`
**Location:** N/A to this PR's diff directly; context for the previous review comment's claim that `unittest_nnstreamer-edge-thread` "is the binary the required build job runs under TSan."

**Why it matters:** I checked `.github/workflows/*.yml` on this branch and found no TSan step anywhere, and confirmed PR #276 ("[Fix] make the flags shared with the library threads atomic", which is where the TSan-in-CI step and the atomic-flag fix for the A5 races mentioned above live) is still open, not merged. So today, the new `edgeThread` tests added here gate merges only via plain `ctest` (in the ASan-enabled `build` job and the `build-ubuntu-26-04` job), which is still real gating for functional regressions — just not the TSan-level sensitivity the previous review credited them with.

**Suggested fix:** Nothing to change in this PR. Just flagging so the record is accurate: once #276 merges, the "future regression would be caught by TSan" claim becomes true; until then, it isn't, and the Minor above (these two tests not reproducing the actual race) is the more relevant caveat.

No other Nits.

myungjoo:
Handling of the re-review:

- **Minor, the start-up tests do not guard the start-up race**: agreed, and there is no deterministic way to guard it. On `main` the early return needs a spurious wake-up of `pthread_cond_wait()`, and a test can only force one by interposing libc, which has hung the GBS build here before. 5961377 notes in both tests that they cover the new start-up order, not the race, and the PR description now says the same. What guards the fix is how simple it is: the flag is raised before `pthread_create()` and nothing waits.
- **Nit, the TSan gate**: correct, that gate comes with #276. The description of this PR only reports a local TSan run.

myungjoo:
**Note: this review was written by another Claude agent (a third, independent pass over the full 3-commit diff and the entire prior conversation on this PR — the two earlier relayed reviews and the author's replies) and is being relayed here verbatim by me. I have not treated the earlier findings as settled; I re-derived the technical claims myself by reading the current source, and I explicitly evaluated whether the author's two declines are adequate rather than assuming they are.**

## Overall verdict: Approve — OK to merge

I independently re-read `nnstreamer-edge-queue.c`, `nnstreamer-edge-internal.c`, `nnstreamer-edge-util.{c,h}` and all three touched test files at the current head (5961377), not just the diff hunks:

- `nns_edge_queue_wait_pop()` now loops `while (q->length == 0U && !q->stopped)`, using `nns_edge_cond_wait(q)` for the untimed case and a single deadline (`nns_edge_get_deadline()`, computed once before the lock) for the timed case, breaking on any non-zero `pthread_cond_timedwait()` return. This correctly fixes both halves of A9: a spurious/empty wake-up (including `nns_edge_queue_clear()` signalling an empty queue) no longer short-circuits the wait with `NNS_EDGE_ERROR_IO`, and a timed wait can no longer return early for the same reason while still honoring its original deadline (not extending it on each wake-up). `_pop_data()` after the loop is unconditional, which is correct — it simply returns nothing when the loop exited via stop/timeout with an empty queue.
- `_nns_edge_create_send_thread()` / `_nns_edge_create_socket_listener()` now set `eh->sending` / `eh->listening = true` before `pthread_create()` and no longer wait for the new thread to signal back. I confirmed both are only reachable from `nns_edge_start()` while `eh->lock` is held for the whole call (lines ~2012–2044 of `nnstreamer-edge-internal.c`), so this is safe per POSIX (writes before `pthread_create()` are visible to the new thread) and removes the prior hang window where `nns_edge_release_handle()` could clear the flag before the new thread ever set it back to `true`. Both creators correctly reset the flag to `false` on `pthread_create()` failure, so there's no leaked "started" state on the error path.
- The `eh->cond` removal (071ba200) is complete: no remaining reference to `nns_edge_cond_wait(eh)`/`nns_edge_cond_signal(eh)`/`nns_edge_cond_init(eh)`/`nns_edge_cond_destroy(eh)` in `src/`, and the field is gone from the struct.
- Scope is proportionate and self-contained: 7 files, all within the queue/internal/util/tests boundary the fix requires; no public header (`nnstreamer-edge-util.h` is explicitly internal, "DO NOT export"), no other module touched, no ARCHITECTURE/API doc update needed since nothing externally observable changed.
- Regression coverage is real, not decorative: the four new `edgeQueue.waitPop*` tests exercise `nnstreamer-edge-queue.c` directly and match the described bug (a wake-up with no data must not end the wait); `edgeUtil.getDeadline{,AcrossSecond}` cover the extracted deadline helper directly, including the second-boundary carry case. These land in `unittest_nnstreamer-edge.cc` / `-mqtt.cc`, which required jobs (`build`, `Build and test on Ubuntu 26.04 (gcc 15)`, the MQTT test against a real broker) run via ctest — so a future regression here would fail a gating check, consistent with this repo's CI facts (GBS's silent MQTT skip is the one path that would *not* catch it, but the 26.04 job does).

### On the declined/handled items from the earlier rounds — I agree with both

1. **`eh->cond` dead code (Minor, round 1):** fixed outright in 71ba200 rather than deferred. Verified clean.
2. **`CLOCK_REALTIME` vs `CLOCK_MONOTONIC` deadlines (Nit, round 1):** correctly scoped out. Switching clocks would require `pthread_condattr_setclock()` at every `nns_edge_cond_init()` call site (the queue, the mosquitto broker handle, ad-hoc test fixtures), which is a behavior change well beyond "add the missing predicate." Reasonable to track separately if wall-clock steps ever prove to matter in practice.
3. **The `edgeThread.releaseRightAfterStart` / `sendImmediatelyAfterConnect` tests don't reproduce the actual start-up race (Minor, round 2), and the author's reply that this can't be fixed without libc interposition:** I agree this decline is adequate, for a reason I checked independently rather than took on faith — this project has direct, documented precedent that interposing a libc symbol (`calloc`) in this test binary hung the Tizen GBS job for 6 hours until the GitHub Actions time limit killed it (PR #263), while interposing a *library's own* symbol was safe. A `pthread_cond_wait()` spurious-wakeup is not something you can force from outside libc without the same kind of interposition (or a custom pthread shim), so the same hang risk applies. Given that, forcing this specific race into a deterministic unit test is a bad trade against a real, previously-paid cost in this exact repo, not a hypothetical one. The commit that documents this (5961377, `[Test] say what the start-up tests of the edge thread cover`) is the right-sized response: it doesn't invent a fragile test, it just stops the two existing tests from silently over-claiming what they guard, by adding `@note` doc-comments and matching prose in the PR description. That is honest and sufficient; I would not ask for more here.
4. **The "TSan-gated" claim was premature (Nit, round 2):** correctly acknowledged (PR #276, which adds the TSan CI step, is still open) with no code change needed — this was a documentation-accuracy point about a different PR, not about this one.

No Blocking or Major issues, and I found no new issue beyond what's already been raised and resolved above.

### Nit (new, mine) — confirm gating jobs finish green before merge
**Location:** N/A to the diff; CI state at review time.
**Why it matters:** As of this review, `build`, `Build and test on Ubuntu 26.04 (gcc 15)`, and all three GBS jobs are still `pending`; only `Static checks`, `Spell Check`, and `DCO` have completed. Nothing in the diff suggests they'd fail, and the PR description already reports a local ASan/TSan run plus a full MQTT pass, but per this repo's branch protection these are required checks and should be watched to green before merge rather than assumed.
**Suggested fix:** No code change; just confirm before merging.

Nothing else remains — no further Blocking, Major, Minor, or Nit findings, and no documentation gap (no public API or architecture change).

